PR #4, opened 2022-04-20
Faster CD solver when `n_samples >> n_features`

PABannier:
The goal of this PR is to write a CD (BCD) solver when `n_samples >> n_features`.
Such configurations are solved much faster by pre-computing a Gram matrix XtX and updating the gradient (rather than the residuals) at every CD cycle.

A quick experiment with 1e6 samples and 300 features:
```
########
Lasso
########

Celer: 5.43s
Gram: 1.89s

##########
Group Lasso
##########

Celer: 43.41s
Gram: 4.23s
```

### DIFF
--- a/skglm/gram_solver.py
+++ b/skglm/gram_solver.py
@@ -1,0 +1,170 @@
+from time import time
+import numpy as np
+from numpy.linalg import norm
+from numba import njit
+from celer import Lasso, GroupLasso
+from benchopt.datasets.simulated import make_correlated_data
+from skglm.utils import BST, ST
+
+
+def _grp_converter(groups, n_features):
+    if isinstance(groups, int):
+        grp_size = groups
+        if n_features % grp_size != 0:
+            raise ValueError("n_features (%d) is not a multiple of the desired"
+                             " group size (%d)" % (n_features, grp_size))
+        n_groups = n_features // grp_size
+        grp_ptr = grp_size * np.arange(n_groups + 1)
+        grp_indices = np.arange(n_features)
+    elif isinstance(groups, list) and isinstance(groups[0], int):
+        grp_indices = np.arange(n_features).astype(np.int32)
+        grp_ptr = np.cumsum(np.hstack([[0], groups]))
+    elif isinstance(groups, list) and isinstance(groups[0], list):
+        grp_sizes = np.array([len(ls) for ls in groups])
+        grp_ptr = np.cumsum(np.hstack([[0], grp_sizes]))
+        grp_indices = np.array([idx for grp in groups for idx in grp])
+    else:
+        raise ValueError("Unsupported group format.")
+    return grp_ptr.astype(np.int32), grp_indices.astype(np.int32)
+
+
+@njit
+def primal(alpha, y, X, w):
+    r = y - X @ w
+    p_obj = (r @ r) / (2 * len(y))
+    return p_obj + alpha * np.sum(np.abs(w))
+
+
+@njit
+def primal_grp(alpha, y, X, w, grp_ptr, grp_indices):
+    r = y - X @ w
+    p_obj = (r @ r) / (2 * len(y))
+    for g in range(len(grp_ptr) - 1):
+        w_g = w[grp_indices[grp_ptr[g]:grp_ptr[g + 1]]]
+        p_obj += alpha * norm(w_g, ord=2)
+    return p_obj
+
+
+@njit
+def cd_epoch(X, G, grads, w, alpha, lipschitz):
+    n_features = X.shape[1]
+    for j in range(n_features):
+        if lipschitz[j] == 0.:
+            continue
+        old_w_j = w[j]
+        w[j] = ST(w[j] + grads[j] / lipschitz[j], alpha / lipschitz[j])
+        if old_w_j != w[j]:
+            grads += G[j, :] * (old_w_j - w[j]) / len(X)
+
+
+@njit
+def bcd_epoch(X, G, grads, w, alpha, lipschitz, grp_indices, grp_ptr):
+    n_groups = len(grp_ptr) - 1
+    for g in range(n_groups):
+        if lipschitz[g] == 0.:
+            continue
+        idx = grp_indices[grp_ptr[g]:grp_ptr[g + 1]]
+        old_w_g = w[idx].copy()
+        w[idx] = BST(w[idx] + grads[idx] / lipschitz[g], alpha / lipschitz[g])
+        diff = old_w_g - w[idx]
+        if np.any(diff != 0.):
+            grads += diff @ G[idx, :] / len(X)
+
+
+def lasso(X, y, alpha, max_iter, tol, check_freq=10):
+    p_obj_prev = np.inf
+    n_features = X.shape[1]
+    # Initialization
+    grads = X.T @ y / len(y)
+    G = X.T @ X
+    lipschitz = np.zeros(n_features, dtype=X.dtype)
+    for j in range(n_features):
+        lipschitz[j] = (X[:, j] ** 2).sum() / len(y)
+    w = np.zeros(n_features)
+    # CD
+    for n_iter in range(max_iter):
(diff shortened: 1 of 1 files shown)


mathurinm:
@PABannier thanks a lot ! I tried it and it really shines on the data by @sehoff

Caveat:  the stopping criteria are not the same. Monitoring the primal decrease is much looser than checking the duality gap. I adjusted the tolerance manually to obtain similar results

sehoff:
Thanks a lot! I tried it out, and can confirm sizable speedups. I use warm-starts for celer, so I guess the figures are even biased towards celer.

**Note**: in the current setting of the gram-solver: ```res = gram_group_lasso(X, y, a, groups=grps, tol=1e-10, max_iter=10_000, check_freq=10)``` with very small alphas (alpha_max/1000), the solver does not convergence after ```max_iter```. 

Do you have any suggestion on a reasonable decrease in the tolerance? Because for ```tol=1e-9``` it reaches convergence after ~3500 iterations (the data is slightly different to the one I provided)
speed_comparison_celer_gram
.

skglm/solvers/gram.py
```diff
@@ -0,0 +1,96 @@
...
+def primal_grp(alpha, y, X, w, grp_ptr, grp_indices):
+    r = y - X @ w
+    p_obj = (r @ r) / (2 * len(y))
+    for g in range(len(grp_ptr) - 1):
+        w_g = w[grp_indices[grp_ptr[g]:grp_ptr[g + 1]]]
+        p_obj += alpha * norm(w_g, ord=2)
+    return p_obj
+
+
+def gram_lasso(X, y, alpha, max_iter, tol, check_freq=10):
```
mathurinm, line 26: @PABannier can you support warm start here and in gram_group_lasso() ?

mathurinm: by that I mean add `w_init=None` as is done in celer_path, if `w = np.zeros(n_features) if w_init is None else w_init.copy()`

mathurinm:
@sehoff hard to tell without the data but consider increasing max_iter, probably the solver is not very far from convergence.You can run it in verbose mode to see if you stop far from convergence or not.

It's easy to add warm start to the gram solvers, it should be done shortly. Beware in your comparison that the stopping criteria are not the same.

PABannier:
Quick update:
- [ ] Gram CD
- [ ] Gram BCD
- [ ] Gram CD - FISTA
- [ ] Gram BCD - FISTA

skglm/solvers/gram.py
```diff
@@ -0,0 +1,235 @@
+import numpy as np
+from numba import njit
+from numpy.linalg import norm
+from celer.homotopy import _grp_converter
```
Badr-MOUFAD, line 4: Consider adding ``celer`` to requirements

skglm/solvers/gram.py
```diff
@@ -0,0 +1,235 @@
...
+    for g in range(len(grp_ptr) - 1):
+        w_g = w[grp_indices[grp_ptr[g]:grp_ptr[g + 1]]]
+        p_obj += alpha * norm(w_g * weights[g], ord=2)
+    return p_obj
+
+
+@njit
+def dual(alpha, norm_y2, theta, y):
+    d_obj = - np.sum((y / (alpha * len(y)) - theta) ** 2)
+    d_obj *= 0.5 * alpha ** 2 * len(y)
```
Badr-MOUFAD, line 27: @mathurinm, 
shouldn't use the same convention as in celer
``theta = y - Xw`` instead of ``theta = (y - Xw) / alpha``

mathurinm: Good point ! This PR will not be merged as is, we don't use the gap in skglm; it's just to do a bench with the same stopping criterion as in celer. So not a problem here

skglm/solvers/gram.py
```diff
@@ -0,0 +1,235 @@
...
+    return w
+
+
+def gram_fista_group_lasso(X, y, alpha, groups, max_iter, tol, w_init=None,
+                           weights=None, check_freq=100):
+    n_features = X.shape[1]
+    norm_y2 = y @ y
+    grp_ptr, grp_indices = _grp_converter(groups, X.shape[1])
+    n_groups = len(grp_ptr) - 1
+    grp_size = n_features // n_groups
```
mathurinm, line 185: is this used elsewhere ? thiqs is valid only if groups are all of same size

skglm/solvers/gram.py
```diff
@@ -0,0 +1,235 @@
...
+@njit
+def primal(alpha, r, w, weights):
+    p_obj = (r @ r) / (2 * len(r))
+    return p_obj + alpha * np.sum(np.abs(w * weights))
+
+
+@njit
+def primal_grp(alpha, norm_r2, r, w, grp_ptr, grp_indices, weights):
+    p_obj = norm_r2 / (2 * len(r))
+    for g in range(len(grp_ptr) - 1):
```
mathurinm, line 18: in celer we needed to skip if weights[g] was np.inf because otherwise we ended up with 0 * np.inf which was nan

same in primal

skglm/solvers/gram.py
```diff
@@ -0,0 +1,235 @@
...
+from numpy.linalg import norm
+from celer.homotopy import _grp_converter
+
+from skglm.utils import BST, ST, BST_vec, ST_vec
+
+
+@njit
+def primal(alpha, r, w, weights):
+    p_obj = (r @ r) / (2 * len(r))
+    return p_obj + alpha * np.sum(np.abs(w * weights))
```
mathurinm, line 12: you need to loop and skip infinite weights to avoid nans

skglm/solvers/gram.py
PABannier: nope just here. indeed i'm writing a custom `prox_l21` function that handles multi-size groups.

mathurinm:
@sehoff support for wieghts (inifnite ones too) is here. Let us know how it works !

PABannier:
@sehoff For very small alphas, I'd recommend using FISTA instead. Have a look at the `gram_fista_group_lasso` function!

sehoff:
> @sehoff support for weights (inifnite ones too) is here. Let us know how it works !
> @sehoff For very small alphas, I'd recommend using FISTA instead. Have a look at the gram_fista_group_lasso function!

Here are the results of my comparisons, where I used the data I provided in the dropbox, and a ```tol=1e-8``` in all solvers. Furthermore, the results I show refer to Case 3.1., i.e., with one weight set to infinity. As a side remark, this highlights that *all three* solvers can handle infinite weights! 
**Bottomline**: especially for (very) small alphas the Gram-based solvers outperform the one in celer significantly. So depending on the grid one eventually searches, each solver has its advantages.
**Comparison 1:**
speed_comparison_celer_gram
**Comparison 2:** *Note: I leave out celer here, because for alpha_max*1e-3 it takes quite long for convergence.*
speed_comparison_bcd_fista

PABannier:
@sehoff indeed, Celer is particularly efficient in settings where `n_features >> n_samples`. It implements a working set strategy, that is particularly useful in high-regularization regime (where there are few active features). In your first figure, for low level of regularization, since your design matrix has way more samples than features, Celer working set strategy is less useful.

Besides, for your data, the Gram solver has a cheaper update since `X.T @ X` is of size (n_features, n_features) (X.T @ X being a useful quantity at every gradient update).

PABannier:
@sehoff Thanks for the plots. They are very insightful for us. By the way, if you ever find yourself in need for an automated way of benchmarking optimization routines, look at https://github.com/benchopt/benchopt

PABannier:
@mathurinm do we want to keep this PR open? FISTA and Gram-solver have been merged

mathurinm:
This one supports groups while the merged PR doesn't, so it does not hurt to leave it open IMO

mathurinm:
note: this is a PR from the `skglm` repo, instead it should be done from a branch of your fork @PABannier

PR #15, opened 2022-05-04
Remove the depreciated call to multiclass.coef in SparseLogisticRegression 

### LINKED ISSUE #13: coef_ and intercept_ are deprecated for OneVsRestClassifier
The following assignment will fail in scikit-learn 1.1:
https://github.com/scikit-learn-contrib/skglm/blob/main/skglm/estimators.py#L917

`coef_` and `intercept_` have been deprecated and are removed in scikit-learn 1.1.
Testing the RC on the code, break one of the tests.

QB3:
Remove the depreciated call to multiclass.coef in SparseLogisticRegression, fix #13

### DIFF
--- a/skglm/estimators.py
+++ b/skglm/estimators.py
@@ -914,7 +914,7 @@ class SparseLogisticRegression(LogReg_sklearn):
             self.coef_ = np.empty([n_classes, X.shape[1]])
             self.intercept_ = 0.
             multiclass = OneVsRestClassifier(self).fit(X, y)
-            self.coef_ = multiclass.coef_
+            self.coef_ = np.array([clf.coef_[0] for clf in multiclass.estimators_])
             self.n_iter_ = max(clf.n_iter_ for clf in multiclass.estimators_)
         return self
 


QB3:
If I understood correctly this was the only change to make

mathurinm:
Thanks ! Do the the pass with the new release candidate of sklearn ?

skglm/estimators.py
```diff
@@ -914,7 +914,7 @@ def fit(self, X, y):
             self.coef_ = np.empty([n_classes, X.shape[1]])
             self.intercept_ = 0.
             multiclass = OneVsRestClassifier(self).fit(X, y)
-            self.coef_ = multiclass.coef_
+            self.coef_ = np.array([clf.coef_[0] for clf in multiclass.estimators_])
```
glemaitre, line 917: To keep full compatibility, you can use the original code of scikit-learn:

```python
        coefs = [e.coef_ for e in self.estimators_]
        if sp.issparse(coefs[0]):
            return sp.vstack(coefs)
        return np.vstack(coefs)
```

I see that we were expecting some sparse matrix sometimes.

glemaitre:
> Thanks ! Do the the pass with the new release candidate of sklearn ?

It will :)

mathurinm (approved):

PR #16, opened 2022-05-05
ENH flexible gram solver with penalty and using datafit

mathurinm:
This is a smaller version of #4 : only without groups, but reusing more code and supporting any penalty

### DIFF
--- a/skglm/solvers/cd_solver.py
+++ b/skglm/solvers/cd_solver.py
@@ -99,7 +99,7 @@ def cd_solver_path(X, y, datafit, penalty, alphas=None,
     # X_dense, X_data, X_indices, X_indptr = _sparse_and_dense(X)
 
     if alphas is None:
-        raise ValueError('alphas should be passed explicitely')
+        raise ValueError('alphas should be passed explicitly')
         # if hasattr(penalty, "alpha_max"):
         #     if sparse.issparse(X):
         #         grad0 = construct_grad_sparse(
@@ -246,8 +246,8 @@ def cd_solver(
     obj_out : array, shape (n_iter,)
         Objective value at every outer iteration.
 
-    stop_crit : array, shape (n_alphas,)
-        Value of stopping criterion at convergence along the path.
+    stop_crit : float
+        Value of stopping criterion at convergence.
     """
     if ws_strategy not in ("subdiff", "fixpoint"):
         raise ValueError(f'Unsupported value for ws_strategy: {ws_strategy}')
--- a/skglm/solvers/cd_solver.py
+++ b/skglm/solvers/cd_solver.py
@@ -2,6 +2,8 @@ import numpy as np
 from numba import njit
 from scipy import sparse
 from sklearn.utils import check_array
+from skglm.datafits.single_task import Quadratic
+from skglm.solvers.gram import cd_gram_quadratic
 
 
 def cd_solver_path(X, y, datafit, penalty, alphas=None,
@@ -109,6 +111,7 @@ def cd_solver_path(X, y, datafit, penalty, alphas=None,
     # else:
         # alphas = np.sort(alphas)[::-1]
 
+    n_samples = len(y)
     n_alphas = len(alphas)
 
     coefs = np.zeros((n_features, n_alphas), order='F', dtype=X.dtype)
@@ -144,10 +147,16 @@ def cd_solver_path(X, y, datafit, penalty, alphas=None,
                 w = np.zeros(n_features, dtype=X.dtype)
                 Xw = np.zeros(X.shape[0], dtype=X.dtype)
 
-        sol = cd_solver(
-            X, y, datafit, penalty, w, Xw,
-            max_iter=max_iter, max_epochs=max_epochs, p0=p0, tol=tol,
-            use_acc=use_acc, verbose=verbose, ws_strategy=ws_strategy)
+        if isinstance(datafit, Quadratic) and n_samples > n_features * 10:
+            # XXX: does n_samples > n_features * 10 look correct?
+            sol = cd_gram_quadratic(
+                X, y, penalty, max_epochs=max_epochs, tol=tol, w_init=None,
+                ws_strategy=ws_strategy, verbose=verbose)
+        else:
+            sol = cd_solver(
+                X, y, datafit, penalty, w, Xw,
+                max_iter=max_iter, max_epochs=max_epochs, p0=p0, tol=tol,
+                use_acc=use_acc, verbose=verbose, ws_strategy=ws_strategy)
 
         coefs[:, t] = w
         stop_crits[t] = sol[-1]
--- a/skglm/solvers/cd_solver.py
+++ b/skglm/solvers/cd_solver.py
@@ -150,7 +150,7 @@ def cd_solver_path(X, y, datafit, penalty, alphas=None,
                 Xw = np.zeros(X.shape[0], dtype=X.dtype)
 
         if (isinstance(datafit, (Quadratic, Quadratic_32)) and n_samples > n_features
-            and n_features < 10_000):
+                and n_features < 10_000):
             # Gram matrix must fit in memory
             sol = cd_gram_quadratic(
                 X, y, penalty, max_epochs=max_epochs, tol=tol, w_init=None,
--- a/skglm/solvers/cd_solver.py
+++ b/skglm/solvers/cd_solver.py
@@ -161,11 +161,11 @@ def cd_solver_path(X, y, datafit, penalty, alphas=None,
             if (hasattr(penalty, "alpha_max") and penalty.alpha /
                     penalty.alpha_max(datafit.Xty) < 1e-3) or solver == "fista":
                 sol = fista_gram_quadratic(
-                    X, y, penalty, max_epochs=max_epochs, tol=tol, w_init=None,
+                    X, y, penalty, max_epochs=max_epochs, tol=tol, w_init=coef_init,
                     verbose=verbose)
             else:
                 sol = cd_gram_quadratic(
-                    X, y, penalty, max_epochs=max_epochs, tol=tol, w_init=None,
+                    X, y, penalty, max_epochs=max_epochs, tol=tol, w_init=coef_init,
                     verbose=verbose)
             w = sol[0]
         else:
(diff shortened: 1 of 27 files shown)


PABannier:
@mathurinm ready for a quick review ;)

skglm/solvers/cd_solver.py
```diff
@@ -144,10 +152,27 @@ def cd_solver_path(X, y, datafit, penalty, alphas=None,
                 w = np.zeros(n_features, dtype=X.dtype)
                 Xw = np.zeros(X.shape[0], dtype=X.dtype)
 
-        sol = cd_solver(
-            X, y, datafit, penalty, w, Xw,
-            max_iter=max_iter, max_epochs=max_epochs, p0=p0, tol=tol,
-            use_acc=use_acc, verbose=verbose, ws_strategy=ws_strategy)
+        if (isinstance(datafit, (Quadratic, Quadratic_32)) and n_samples > n_features
+                and n_features < 10_000) or solver in ("cd_gram", "fista"):
+            # Gram matrix must fit in memory hence the restriction n_features < 1e5
```
mathurinm, line 157:
```suggestion
            # Gram matrix must fit in memory hence the restriction n_features < 1e4
```

skglm/solvers/cd_solver.py
```diff
@@ -144,10 +152,27 @@ def cd_solver_path(X, y, datafit, penalty, alphas=None,
...
+            if not isinstance(datafit, (Quadratic, Quadratic_32)):
```
mathurinm, line 158: I think this bit is unreachable because the check is already performed L155

skglm/solvers/cd_solver.py
```diff
@@ -214,11 +239,11 @@ def cd_solver(
...
     Returns
     -------
-    alphas : array, shape (n_alphas,)
-        The alphas along the path where models are computed.
+    w : array, shape (n_features,)
+        Coefficients.
 
-    coefs : array, shape (n_features, n_alphas)
-        Coefficients along the path.
+    obj_out : array, shape (n_iter,)
```
mathurinm, line 245: do we really return this? or the optimality condition violation instead

skglm/solvers/gram.py
```diff
@@ -0,0 +1,272 @@
...
+                print(f"Epoch {epoch + 1}, objective {p_obj:.10f}, "
+                      f"stopping crit {stop_crit:.2e}")
+            if stop_crit <= tol:
+                break
+            obj_out.append(p_obj)
+    return w, np.array(obj_out), stop_crit
+
+
+@njit
+def _cd_epoch_gram(G, grads, w, datafit, penalty, n_samples, n_features):
```
mathurinm, line 100: I'm not sure we need to document private methods (maintenance has a cost).

We could rename G (not explicit, needs to be documented) to XtX and remove the docstring.

WDYT ?

skglm/solvers/gram.py
```diff
@@ -0,0 +1,272 @@
...
+    """Run an epoch of coordinate descent in place with gradient update using Gram.
+
+    Parameters
+    ----------
+    G : array, shape (n_features, n_features)
+        Gram matrix.
+
+    grads : array, shape (n_features,)
```
mathurinm, line 108: `grads` is called `grad` in other parts of the code I believe (it's a single gradient, not many of them)

skglm/solvers/gram.py
```diff
@@ -0,0 +1,272 @@
...
+        stepsize = 1 / lc[j]
+        w[j] = penalty.prox_1d(old_w_j + grads[j] / lc[j], stepsize, j)
+        diff = old_w_j - w[j]
+        if diff != 0:
+            for i in range(G_indptr[j], G_indptr[j + 1]):
+                grads[G_indices[i]] += diff * G_data[i] / n_samples
+
+
+def fista_gram_quadratic(X, y, penalty, max_epochs=1000, tol=1e-4, w_init=None,
+                         ws_strategy="subdiff", verbose=False):
```
mathurinm, line 185: there is no working set with gram so stop_crit may be better

but anyway, I propose to keep the code simple and use only subdiff_distance anyway. Would that be ok ?

skglm/solvers/gram.py
```diff
@@ -0,0 +1,272 @@
...
+            continue
+        old_w_j = w[j]
+        stepsize = 1 / lc[j] if lc[j] != 0 else 1000
+        w[j] = penalty.prox_1d(old_w_j + grads[j] * stepsize, stepsize, j)
+        if old_w_j != w[j]:
+            grads += G[j, :] * (old_w_j - w[j]) / n_samples
+
+
+@njit
+def _cd_epoch_gram_sparse(G_data, G_indptr, G_indices, grads, w, datafit, penalty,
```
mathurinm, line 138: don't bother with this, G is likely not to be sparse even if X is. Cast X.T @ X to dense L65 if X is sparse

skglm/solvers/cd_utils.py
```diff
@@ -0,0 +1,138 @@
...
+    """
+    grad = np.zeros(ws.shape[0])
+    for idx, j in enumerate(ws):
+        grad[idx] = datafit.gradient_scalar_sparse(
+            data, indptr, indices, y, Xw, j)
+    return grad
+
+
+@njit
+def prox_vec(penalty, z, stepsize, n_features):
```
mathurinm, line 118: arf, I though we had `penalty.prox`

make this function private, remove n_features (access as z.shape[1]) 

we need a reflection on solvers, but probably all penalties will need to implement it. We can do so in basepenalty, but I fear looping over all coordinates will be slower than performing it in one step as `ST_vec` does

skglm/solvers/gram.py
```diff
@@ -0,0 +1,272 @@
...
+    construct_grad, construct_grad_sparse, dist_fix_point, prox_vec)
+
+
+def cd_gram_quadratic(X, y, penalty, max_epochs=1000, tol=1e-4, w_init=None,
+                      ws_strategy="subdiff", verbose=0):
+    r"""Run a coordinate descent solver using Gram update for quadratic datafit.
+
+    This solver should be used when n_samples >> n_features. It does not implement any
+    working set strategy and iteratively updates the gradients (n_features,) instead of
+    the residuals (n_samples,).
```
mathurinm, line 16:
```suggestion
    the prediction Xw (n_samples,).
```

skglm/solvers/gram.py
PABannier: Sure what the function does is quite explicit. I added docstrings to be consistent with other private functions.

skglm/solvers/gram.py
PABannier: I cleaned it.

skglm/solvers/cd_utils.py
PABannier:
```
In [16]: %%time
    ...: out = _prox_vec(pen, z, 0.01)
CPU times: user 28 µs, sys: 1e+03 ns, total: 29 µs
Wall time: 34.1 µs

In [17]: %%time
    ...: out2 = ST_vec(z, 0.01)
CPU times: user 23 µs, sys: 0 ns, total: 23 µs
Wall time: 25.7 µs
```

not a big difference, from my experiments. I tried with different thresholds.

skglm/solvers/cd_solver.py
PABannier: I've placed it because the first condition is "`isinstance....` OR solver in ...". If the user manually inputs "cd_gram", I want to catch a wrong datafit, hence L158. Overkill maybe? Should we even expose `solver`? I think it is convenient for benchmarks. 
WDYT?

skglm/solvers/cd_utils.py
mathurinm: with @QB3 we had an issue a while ago on flashcd with finance where this caused a big overhead. Just to keep it in mind

skglm/solvers/cd_solver.py
mathurinm: Ok I understood, thanks. 
Maybe we can indent the first if, breaking line before  `or solver` to make it more visible ?

skglm/solvers/cd_solver.py
```diff
@@ -144,10 +152,27 @@ def cd_solver_path(X, y, datafit, penalty, alphas=None,
...
+        if (isinstance(datafit, (Quadratic, Quadratic_32)) and n_samples > n_features
+                and n_features < 10_000) or solver in ("cd_gram", "fista"):
+            # Gram matrix must fit in memory hence the restriction n_features < 1e4
+            if not isinstance(datafit, (Quadratic, Quadratic_32)):
+                raise ValueError("`cd_gram` and `fista` solvers are only supported " +
+                                 "for `Quadratic` datafits.")
+            if (hasattr(penalty, "alpha_max") and penalty.alpha /
+                    penalty.alpha_max(datafit.Xty) < 1e-3) or solver == "fista":
+                sol = fista_gram_quadratic(
+                    X, y, penalty, max_epochs=max_epochs, tol=tol, w_init=None,
```
mathurinm, line 164: why not using coef_init ?

skglm/solvers/cd_solver.py
PABannier: We do return this. See L371.

skglm/solvers/cd_solver.py
PABannier: I tried to make it more obvious. WDYT?

skglm/solvers/cd_solver.py
```diff
@@ -52,6 +56,9 @@ def cd_solver_path(X, y, datafit, penalty, alphas=None,
     return_n_iter : bool, optional
         If True, number of iterations along the path are returned.
 
+    solver : ('cd_ws'|'cd_gram'|'fista'), optional
```
PABannier, line 59: FISTA is not a CD solver, it's confusing to expose it to the user like this.
@mathurinm WDYT?

PABannier (changes requested):
Overall LGTM.
Tests are missing for the solvers though, I can write some if needed.

skglm/solvers/gram.py
```diff
@@ -0,0 +1,184 @@
...
+            if stop_crit <= tol:
+                break
+            obj_out.append(p_obj)
+    return w, np.array(obj_out), stop_crit
+
+
+@njit
+def _cd_epoch_gram(XtX, grad, w, datafit, penalty, n_samples, n_features):
+    lc = datafit.lipschitz
+    for j in range(n_features):
```
mathurinm, line 92: since we have complete access to `grad` at each iteration, it would be interesting to use a greedy selection rule here: do not pick `j` cyclically, but instead take `j = np.argmax(np.abs(grad))`

One "epoch" in this setting would only be the update of `n_features` coordinates.

Badr-MOUFAD:
closing in favor of #59

PR #19, opened 2022-05-15
[READY] ENH add fit_intercept

### LINKED ISSUE #18: ENH support fit_intercept
Fititng an intercept can drastically change the size of the supports and the time it takes to fit: 
```python
import numpy as np
import time
from libsvmdata import fetch_libsvm
X, y = fetch_libsvm("finance")

from celer import Lasso

for fit_intercept in [False, True]:
    alpha_max = np.max(np.abs(X.T @ (y - fit_intercept*np.mean(y)))) / len(y)
    clf = Lasso(alpha_max / 200, fit_intercept=fit_intercept, verbose=1)

    t0 = time.time()
    clf.fit(X, y)
    t1 = time.time()
    print(f"### fit_intercept={fit_intercept}, time to fit {t1-t0:.2f} s, supp size: {(clf.coef_ != 0).sum()}")
```

To do it without pain my idea is to add a gradient step with `np.ones(features)` in the solver at the end of each CD epoch. That is like adding a column full of ones to X, but without copying it.

One needs to add a way to compute the lipschitz constant of this; I do not see another solution than to make it an attribute of datafit (eg 1 for Quadratic, 1/4 for Logistic)

PABannier:
Closes #18


mathurinm:
I would suggest to start by the algorithmic part first, so that the diff is smaller and we can focus our attention on the hard part. But it's minor, we can always mark files as viewed in the diff.

skglm/estimators.py
```diff
@@ -155,10 +155,10 @@ def path(self, X, y, alphas, coef_init=None, return_n_iter=True, **params):
         path_func = cd_solver_path if y.ndim == 1 else bcd_solver_path
         return path_func(
             X, y, self.datafit, self.penalty, alphas=alphas,
-            coef_init=coef_init, max_iter=self.max_iter,
-            return_n_iter=return_n_iter, max_epochs=self.max_epochs, p0=self.p0,
-            tol=self.tol, use_acc=True, ws_strategy=self.ws_strategy,
-            verbose=self.verbose)
+            fit_intercept=self.fit_intercept, coef_init=coef_init,
```
mathurinm, line 158: can you add it at the end (just keep verbose last) to make the diff shorter ?

skglm/estimators.py
```diff
@@ -1072,13 +1068,8 @@ def fit(self, X, y):
...
 
         X, y = self._validate_data(
-            X,
-            y,
-            accept_sparse="csc",
-            dtype=np.float64,
-            order="C",
-            accept_large_sparse=True,
-        )
+            X, y, accept_sparse="csc", dtype=np.float64, order="C",
```
mathurinm, line 1080: revert

skglm/solvers/cd_solver.py
```diff
@@ -4,7 +4,7 @@
 from sklearn.utils import check_array
 
 
-def cd_solver_path(X, y, datafit, penalty, alphas=None,
+def cd_solver_path(X, y, datafit, penalty, alphas=None, fit_intercept=True,
```
mathurinm, line 8: to discuss: this is the sklearn default setting, but it requires all datafits to implement fit_intercept. And it's usually confusing for the user. I'd set it to False even if it disagrees with sklearn

skglm/solvers/cd_solver.py
```diff
@@ -112,6 +107,8 @@ def cd_solver_path(X, y, datafit, penalty, alphas=None,
     n_alphas = len(alphas)
 
     coefs = np.zeros((n_features, n_alphas), order='F', dtype=X.dtype)
+    if fit_intercept:
+        intercepts = np.zeros((1, n_alphas), order='F', dtype=X.dtype)
```
mathurinm, line 112: there is no need to specify fortran order as the array has only one row, it's both C and F contiguous

skglm/solvers/cd_solver.py
```diff
@@ -145,15 +142,17 @@ def cd_solver_path(X, y, datafit, penalty, alphas=None,
...
             use_acc=use_acc, verbose=verbose, ws_strategy=ws_strategy)
 
         coefs[:, t] = w
+        if fit_intercept:
+            intercepts[0, t] = sol[1]
         stop_crits[t] = sol[-1]
 
         if return_n_iter:
-            n_iters[t] = len(sol[1])
+            n_iters[t] = len(sol[2])
```
mathurinm, line 156: this means the signature of cd_solver is changed. It can break existing code, I would avoid this

skglm/solvers/cd_solver.py
```diff
@@ -187,6 +187,9 @@ def cd_solver(
     Xw : array, shape (n_samples,)
         Model fit.
 
+    fit_intercept : bool
+        Fit an intercept.
```
mathurinm, line 192: Whether or not to fit an intercept

skglm/solvers/multitask_bcd_solver.py
```diff
@@ -7,7 +7,7 @@
 
 
 def bcd_solver_path(
-        X, Y, datafit, penalty, alphas=None,
+        X, Y, datafit, penalty, alphas=None, fit_intercept=True,
```
mathurinm, line 10: Id address this in another PR

skglm/datafits/single_task.py
```diff
@@ -84,6 +84,9 @@ def full_grad_sparse(
             grad[j] = (XjTXw - self.Xty[j]) / n_samples
         return grad
 
+    def update_intercept(self, y, Xw):
```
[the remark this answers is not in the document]
Klopfe: I agree with the name, however, the stepsize is equal to n_samples / n_samples so 1 but the gradient remains divided by n_samples right?

skglm/solvers/cd_solver.py
Klopfe: agreed!

skglm/solvers/cd_solver.py
Klopfe: then we have stop_crits[t] = sol[-1] so I cannot add the intercept at the end either. 
Should i do return coefs, n_iters, intercept stop_crits? 
Does not make lots of sense either to me

mathurinm: can you return a coefs which is longer by 1 if `fit_intercept` ? This way the signature does not change. And we split in the output of the solver

skglm/datafits/single_task.py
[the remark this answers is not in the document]
mathurinm: Yes, you are right. And it makes sense as what you have written currently means that we center the residuals. Great !

skglm/solvers/cd_solver.py
```diff
@@ -355,7 +357,9 @@ def cd_solver(
                             print("Early exit")
                         break
         obj_out.append(p_obj)
-    return w, intercept, np.array(obj_out), stop_crit
+    if fit_intercept:
```
mathurinm, line 360: since we need to extrapolate the intercept, what do you think of using a `w` of length `n_features + 1` inside the solver too ?

Klopfe: so if fit_intercept=True, we would instantiate w of size n_features + 1 from the beginning? 
isn't it gonna be a problem for the warm_start in solver_path?

skglm/solvers/cd_solver.py
```diff
@@ -278,8 +283,13 @@ def cd_solver(
...
+                _cd_epoch(X, y, w, Xw, datafit, penalty, ws, fit_intercept, intercept)
 
+            # update intercept
+            if fit_intercept:
+                intercept_old = intercept
+                intercept -= datafit.update_intercept(y, Xw)
+                Xw += (intercept - intercept_old)
             # 3) do Anderson acceleration on smaller problem
             # TODO optimize computation using ws
             if use_acc:
```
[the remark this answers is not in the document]
Klopfe: yes, that would be nice indeed

mathurinm:
Yes, sklearn supports intercept fitting too

skglm/tests/test_estimators.py
```diff
@@ -96,12 +96,15 @@ def test_check_estimator(estimator_name):
...
 @pytest.mark.parametrize("estimator_name", dict_estimators_ours.keys())
 @pytest.mark.parametrize('X', [X, X_sparse])
-def test_estimator(estimator_name, X):
+def test_estimator(estimator_name, X, fit_intercept=False):
     if estimator_name == "GeneralizedLinearEstimator":
         pytest.skip()
     estimator_sk = dict_estimators_sk[estimator_name]
     estimator_ours = dict_estimators_ours[estimator_name]
 
+    estimator_ours.fit_intercept = fit_intercept
```
mathurinm, line 105: this affects the dict of estimators (that's why I don't like this design @QB3)

if the test fails, we don't know precisely why.

@Klopfe can you add `test_intecepts.py` with a few tests of cd_solver over there (comapred with sklearn Lasso, Huber and Logreg for example), not encapsulated inside our estimators ?

Klopfe: yes I will do that!

Klopfe:
So, trying to make Lasso inherit from GeneralizedLinearModel, it does not pass check_estimator of sklearn anymore because of what is put in the comment of test_estimators.py
i.e.:
# Currently, `GeneralizedLinearEstimator` does not pass sklearn's `check_estimator`
# tests. Indeed, jitclasses which `GeneralizedLinearEstimator` depends upon (both the
# datafit and penalty objects are jitclasses) are not serializable ("pickleable"). It is
# a non-trivial well-known issue in Numba.
# For more information, see: https://github.com/numba/numba/issues/1846 .

Klopfe:
En fait, on ne peut pas mettre nos classes Jit en attribut d'une classe, sinon ça fait péter check_estimator.
Du coup, on devrait définir un fit pour chaque estimateur en créant les instantes de datafit et penalty dans le fit et pas comme un attribut mais ça devient lourd

skglm/datafits/single_task.py
```diff
@@ -230,6 +236,9 @@ def full_grad_sparse(
             grad[j] = yXjyXTw - 1
         return grad
 
+    def intercept_update_step(self, y, Xw):
+        pass
```
mathurinm, line 240: would it be better to raise an error instead of ignoring this silently ?

skglm/solvers/cd_solver.py
```diff
@@ -134,23 +128,23 @@ def cd_solver_path(X, y, datafit, penalty, alphas=None,
         else:
             if coef_init is not None:
                 w = coef_init.copy()
-                supp_size = penalty.generalized_support(w).sum()
+                supp_size = penalty.generalized_support(w[:n_features]).sum()
                 p0 = max(supp_size, p0)
                 if supp_size:
-                    Xw = X @ w
+                    Xw = X @ w[:n_features] + fit_intercept * w[-1]
```
mathurinm, line 134: nice !

skglm/solvers/cd_solver.py
```diff
@@ -235,6 +232,17 @@ def cd_solver(
     stop_crit = np.inf  # initialize for case n_iter=0
 
     is_sparse = sparse.issparse(X)
+
+    if fit_intercept and len(w) != n_features + 1:
+        raise ValueError(
+            "Shape of w should be %i." % (n_features + 1)
+            + " Got %i" % len(w))
+
+    if not fit_intercept and len(w) != n_features:
```
mathurinm, line 241: very nice

mathurinm: maybe merge with above: if len(w) != n_features + intercept ?

and add a clear message saying when fit_intercept=True w needs be of shape n_feats  + 1 (written explicitely, not as its value)

QB3:
It remains to fix the grid-search tests, @Klopfe was thinking adding a `set_param` method for each estimator.

mathurinm:
what do you have in mind ?

Is it not better to set them by calling them ``penalty__alpha`` ? As in https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.GridSearchCV.html#sklearn.model_selection.GridSearchCV.set_params

Klopfe:
Bizarre, pour moi sur mon mac pytest passe. J'ai même regardé les versions de sklearn et des dépendances et tout semble. Quelqu'un peut essayer de pull et de lancer les tests en local svp?

Klopfe:
> Bizarre, pour moi sur mon mac pytest passe. J'ai même regardé les versions de sklearn et des dépendances et tout semble. Quelqu'un peut essayer de pull et de lancer les tests en local svp?

nvm, avec Python 3.9 je vois l'erreur

Klopfe:
So fit_intercept is up and running with the new API for our estimator. Let me know what you think.
What needs to be addressed (maybe in another PR)
- [ ]  not easy to find a way for the new API with Weighted Lasso (to discuss)
- [ ] SVM will also have to be addressed separately, as we solve the dual it is not easy to integrate right away
- [ ] Add LogReg and SVC to the grid_search test

QB3:
+1 to address the SVM separately since we will not add a `fit_intercept` for this estimator

Klopfe:
> So fit_intercept is up and running with the new API for our estimator. Let me know what you think. What needs to be addressed (maybe in another PR)
> 
> * [x]   not easy to find a way for the new API with Weighted Lasso (to discuss)
> * [ ]  SVM will also have to be addressed separately, as we solve the dual it is not easy to integrate right away
> * [x]  Add LogReg and SVC to the grid_search test

Done! :D

skglm/datafits/single_task.py
```diff
@@ -334,3 +340,14 @@ def full_grad_sparse(
                     grad_j += - X_data[i] * np.sign(tmp) * self.delta
             grad[j] = grad_j / n_samples
         return grad
+
+    def intercept_update_step(self, y, Xw):
+        n_samples = len(y)
+        update = 0.
+        for i in range(n_samples):
+            tmp = y[i] - Xw[i]
```
mathurinm, line 348: cosmetic comment: `tmp` can be avoided. WDYT of `residual`?

skglm/datafits/single_task.py
```diff
@@ -334,3 +340,14 @@ def full_grad_sparse(
...
+            if abs(tmp) < self.delta:
+                update -= tmp
+            else:
+                update -= np.sign(tmp) * self.delta
+        return update / n_samples
```
mathurinm, line 353: just to double check the maths: our definition of the Huber function is 1-smooth, so their is no additional stepsize (apart from the n_samples normalization) ?

skglm/penalties/base.py
```diff
@@ -60,3 +60,14 @@ def is_penalized(self, n_features):
     @abstractmethod
     def generalized_support(self, w):
         r"""Return a mask which is True for coefficients in the generalized support."""
+
+    def get_params(self, deep=False):
```
mathurinm, line 64: `deep` is ignored here, is it deliberate?

skglm/solvers/cd_solver.py
```diff
@@ -31,7 +31,10 @@ def cd_solver_path(X, y, datafit, penalty, alphas=None,
     alphas : ndarray
         List of alphas where to compute the models.
 
-    coef_init : ndarray, shape (n_features,) | None, optional, (default=None)
+    fit_intercept : bool
+        Whether or not to fit an intercept.
+
+    coef_init : ndarray, shape (n_features + 1,) | None, optional, (default=None)
```
mathurinm, line 37: n_features + 1 even if no intercept is used?

I'd be explicit: `shape (n_features,) or (n_features + 1,)`.

skglm/solvers/cd_solver.py
```diff
@@ -64,7 +67,7 @@ def cd_solver_path(X, y, datafit, penalty, alphas=None,
     alphas : array, shape (n_alphas,)
         The alphas along the path where models are computed.
 
-    coefs : array, shape (n_features, n_alphas)
+    coefs : array, shape (n_features + 1, n_alphas)
```
mathurinm, line 70: same as above

skglm/solvers/cd_solver.py
```diff
@@ -111,17 +105,13 @@ def cd_solver_path(X, y, datafit, penalty, alphas=None,
...
+    coefs = np.zeros((n_features + fit_intercept, n_alphas), order='F', dtype=X.dtype)
     stop_crits = np.zeros(n_alphas)
 
     if return_n_iter:
         n_iters = np.zeros(n_alphas, dtype=int)
 
     for t in range(n_alphas):
-
-        alpha = alphas[t]
-        penalty.alpha = alpha  # TODO this feels it will break sklearn compat
```
mathurinm, line 124: I don't see where `alpha` is changed in the case of a path computation, is it working just because we always call it with a single value?

skglm/solvers/cd_solver.py
```diff
@@ -134,23 +124,23 @@ def cd_solver_path(X, y, datafit, penalty, alphas=None,
...
                 Xw = np.zeros(X.shape[0], dtype=X.dtype)
 
         sol = cd_solver(
-            X, y, datafit, penalty, w, Xw,
+            X, y, datafit, penalty, w, Xw, fit_intercept=fit_intercept,
             max_iter=max_iter, max_epochs=max_epochs, p0=p0, tol=tol,
             use_acc=use_acc, verbose=verbose, ws_strategy=ws_strategy)
 
-        coefs[:, t] = w
+        coefs[:, t] = sol[0]
```
mathurinm, line 143: `w` should still work, because it is of size`n_features + fit_intercept`. 
Otherwise it means we have no way to warm start the intercept, which is not great

skglm/solvers/cd_solver.py
```diff
@@ -215,7 +208,7 @@ def cd_solver(
 
     Returns
     -------
-    coefs : array, shape (n_features, n_alphas)
+    coefs : array, shape (n_features + 1, n_alphas)
```
mathurinm, line 211: same remark as above

skglm/solvers/cd_solver.py
```diff
@@ -235,38 +228,43 @@ def cd_solver(
     stop_crit = np.inf  # initialize for case n_iter=0
 
     is_sparse = sparse.issparse(X)
+
+    if len(w) != n_features + fit_intercept:
+        raise ValueError(
+            "Shape of w should be n_features + 1 when fit_intercept=True.")
```
mathurinm, line 234: nice
Can you add `" expected %d, got %d" ` to make the message crystal clear for the user?

skglm/datafits/single_task.py
mathurinm: It would also be clearer IMO if it was Xw[i] - y[i]   and we removed the minus L350 and L352

Klopfe: You're completely right, my bad. Thanks :)

skglm/tests/test_intercept.py
```diff
@@ -0,0 +1,61 @@
+from skglm.solvers.cd_solver import cd_solver
```
mathurinm, line 1: refactor imports:
standard/numpy/pytest/etc imports 

blank line

skglm imports

skglm/tests/test_intercept.py
```diff
@@ -0,0 +1,61 @@
...
+from skglm.datafits import Quadratic, Logistic, Huber
+from skglm.penalties import L1, WeightedL1
+from skglm.utils import make_correlated_data
+from sklearn.linear_model import HuberRegressor, Lasso, LogisticRegression
+import numpy as np
+import pytest
+from skglm.utils import compiled_clone
```
mathurinm, line 8: merge with other imports from skglm.utils

mathurinm (changes requested):
One thing I don't understand is how alpha varies in Lasso path, otherwise looks neat, thanks all!

skglm/datafits/single_task.py
Klopfe: @mathurinm tu as raison, pour le moment alpha ne varie pas dans le path, et ce n'est pas testé d'ailleurs...
J'aurai éventuellement besoin qu'on s'appelle pour fix ça si tu as du temps à m'accorder

skglm/solvers/cd_solver.py
Badr-MOUFAD: Yes indeed, it breaks when fitting with multiple alphas

skglm/solvers/cd_solver.py
```diff
@@ -235,38 +228,43 @@ def cd_solver(
...
-            grad = construct_grad(X, y, w, Xw, datafit, all_feats)
+            grad = construct_grad(X, y, w[:n_features], Xw, datafit, all_feats)
 
         if ws_strategy == "subdiff":
-            opt = penalty.subdiff_distance(w, grad, all_feats)
+            opt = penalty.subdiff_distance(w[:n_features], grad, all_feats)
         elif ws_strategy == "fixpoint":
-            opt = dist_fix_point(w, grad, datafit, penalty, all_feats)
+            opt = dist_fix_point(w[:n_features], grad, datafit, penalty, all_feats)
         stop_crit = np.max(opt)
```
Badr-MOUFAD, line 247: Perhaps I'm mistaken, but I can't see how we include intercept when checking convergence.

mathurinm:
This PR has become a bit too large and has been divided into #54 and #55

PR #22, opened 2022-05-16
[READY] DOC - Add GitHub link to source code

Badr-MOUFAD:


Badr-MOUFAD:
### Original problem

Links are miscreated for jitclasses as there is no reference to the wrapped class. Hence, we fail to get its source file as well as the line where it is declared

Example of the generated link for ``skglm.datafits.Quadratic``:  
``.../skglm/blob/a2ef5d1/venv/lib/site-packages/numba/experimental/jitclass/base.py#L``

### Suggested solution

jitclasses preserve the reference to the declared methods. Hence we can get the source/line of one method among them.


Example of the generated link for ``skglm.datafits.Quadratic``:  
``.../skglm/blob/a2ef5d1/skglm/datafits/single_task.py#L63``


### Limitations

The suggested solution reuses as much code as possible, yet it fails to generate the exact line where the declaration begins as we use one of the methods to create the link.

Badr-MOUFAD:
Since the datafits and penalties are jitted at fit time, following https://github.com/scikit-learn-contrib/skglm/pull/44, the links to the source code are now correctly generated.

mathurinm (approved):

mathurinm:
Thanks @Badr-MOUFAD, it's a nice functionality!

PR #24, opened 2022-05-16
FIX - scikit-learn ``v1.1.0`` breaks doc

### LINKED ISSUE #23: doc building fails because of scikit-learn update
max_iter = 0 is no longer supported in sklearn 1.1, it raises:
```
ValueError: max_iter == 0, must be >= 1.
```

we can use max_iter=1 it's not important

Badr-MOUFAD:
fixes #23


mathurinm:
Thank you @Badr-MOUFAD

PR #21, opened 2022-05-16
MNT remove seaborn as a dependency

Badr-MOUFAD:
This brings the following changes

- reduce packages dependencies
- add cache folder to .gitignore
- ``setup.py`` code enhancements
- ~add standardize dev-requirements~

If merged, it fixes #20

PR #25, opened 2022-05-19
MAINT - remove redundancies

Badr-MOUFAD:
As continuity to #21

- [x] use ``benchopt`` utils instead of reinventing them


mathurinm:
See why the build fails : benchopt is not installed. We don't want benchopt as a requirement. Every requirement added has a maintenance cost.

You never know what they may change, and when they do it's additional work on our side as you've seen with sklearn removing support for max_iter=0

PR #26, opened 2022-05-22
ENH add single task group solver

Badr-MOUFAD:
The goal of this PR is to add a [weighted group lasso](https://arxiv.org/abs/1602.06225) estimator: 

$$\\text{datafit: } \\frac{1}{2 n} \\Vert y \\ - \\ X \\beta \\Vert_2^2 \\ \\ , \\  \\text{ penalty: }   \\sum_g \\omega_g \\Vert \\beta_g \\Vert_2 \\ $$

It will proceed as follows 

- [x] implemented a vanilla block cd solver  _(cross-tested successfully)_

Afterward,

- ~[ ] implement the sk estimator~
- [x] add unit testing
- ~[ ] add examples to the gallery~


skglm/penalties/block_separable.py
```diff
@@ -153,3 +154,69 @@ def subdiff_distance(self, W, grad, ws):
...
+
+    def is_penalized(self, _):
+        "Groups of features with inf weights are not penalized."
+        n_feature_groups = [len(g) for g in self.groups]
+        return np.repeat(self.weights, n_feature_groups) != np.inf
+
+    def generalized_support(self, w):
+        return w != 0
+
+    def subdiff_distance(self, w, grad, ws):
```
Badr-MOUFAD, line 220: Here, we should compute the distance between the $\\frac{-\nabla f}{\\alpha}$ and the dual unit ball (i.e. $\\mathcal{B}_{\Omega^D}[0, 1]$)

skglm/penalties/block_separable.py
```diff
@@ -153,3 +154,69 @@ def subdiff_distance(self, W, grad, ws):
...
+            pen_group_term += alpha * (1-tau) * weights[g] * norm(w[g], ord=2)
+
+        return pen_l1_term + pen_group_term
+
+    def prox_1feat(self, value: np.ndarray, stepsize: float, j: int) -> np.ndarray:
+        tau, weight_j = self.tau, self.weights[j]
+        mask_zero_coefs = value != 0
+
+        # ST for each features
+        value[mask_zero_coefs] = ST_vec(
```
Badr-MOUFAD, line 205: This is pretty dangerous here as I do an in-place assignment and hence the changes are reflected directly into the original object. I didn't want to create a new instance as looking back at the code, I find out that we constantly feed ``prox_1feat`` a new vector (``value`` is obtained through an arithmetic operation between two arrays)

Badr-MOUFAD: The changes are reflected directly in the original instance.

Let's say we want to return an array of the iterates $\beta^{(k)}$, then following this design, the elements will be altered.

mathurinm: you should use copy when storing all iterates.
value is the result after a gradient step so it's not an iterate
the current design is more memory efficient

skglm/tests/test_group.py
```diff
@@ -0,0 +1,133 @@
...
+    # usual L1 solver
+    l1_penalty = L1(alpha)
+    usual_quad = Quadratic()
+
+    _, w_l1_solver, *_ = cd_solver_path(X, y, usual_quad, l1_penalty, alphas=[alpha])
+
+    assert_allclose(w_l1_solver.flatten(), w_group_solver, atol=1e-3, rtol=1e-3)
+
+
+def test_group_lasso():
```
Badr-MOUFAD, line 102: @mathurinm, I tested the vanilla cd against celer ``GroupLasso`` and it works well.

I shall move confidently to Anderson extrapolation.

mathurinm: great work !

skglm/datafits/multi_task.py
```diff
@@ -89,3 +89,52 @@ def full_grad_sparse(self, X_data, X_indptr, X_indices, Y, XW):
                     XjTXW[t] += X_data[i] * XW[X_indices[i], t]
             grad[j, :] = (XjTXW - self.XtY[j, :]) / n_samples
         return grad
+
+
+spec_QuadraticGroup = [
```
mathurinm, line 94: Group datafits are not multitask (the response `y` is still a vector, not a matrix) so the code rather belongs to a newly created `datafits/group.py`

skglm/datafits/multi_task.py
```diff
@@ -89,3 +89,52 @@ def full_grad_sparse(self, X_data, X_indptr, X_indices, Y, XW):
...
+    ('lipschitz', float64[:]),
+    ('grp_ptr', int32[:]),
+    ('grp_indices', int32[:])
+]
+
+
+@jitclass(spec_QuadraticGroup)
+class QuadraticGroup(BaseMultitaskDatafit):
+    def __init__(self, grp_ptr: np.ndarray, grp_indices: np.ndarray):
```
mathurinm, line 103: we don't use type annotations in the codebase (yet)

skglm/datafits/multi_task.py
```diff
@@ -89,3 +89,52 @@ def full_grad_sparse(self, X_data, X_indptr, X_indices, Y, XW):
...
+
+spec_QuadraticGroup = [
+    ('lipschitz', float64[:]),
+    ('grp_ptr', int32[:]),
+    ('grp_indices', int32[:])
+]
+
+
+@jitclass(spec_QuadraticGroup)
+class QuadraticGroup(BaseMultitaskDatafit):
```
mathurinm, line 102: Does it need to inherit from BaseMultitask instead of the regular BaseDatafit ?

skglm/datafits/multi_task.py
```diff
@@ -89,3 +89,52 @@ def full_grad_sparse(self, X_data, X_indptr, X_indices, Y, XW):
...
+        n_samples = y.shape[0]
+        return 1 / (2*n_samples) * norm(y - Xw, ord=2) ** 2
+
+    def gradient_j(self, X: np.ndarray, y: np.ndarray,
+                   w: np.ndarray, Xw: np.ndarray, g: int) -> np.ndarray:
+        grp_ptr, grp_indices = self.grp_ptr, self.grp_indices
+        n_samples = y.shape[0]
+
+        grp_p_indices = grp_indices[grp_ptr[g]:grp_ptr[g+1]]
+        X_g = X[:, grp_p_indices]
```
mathurinm, line 130: this is inefficient, it requires building the matrix if groups are not contiguous. Use a for loop instead

skglm/penalties/block_separable.py
```diff
@@ -153,3 +156,124 @@ def subdiff_distance(self, W, grad, ws):
...
+    ('alpha', float64),
+    ('tau', float64),
+    ('weights', float64[:]),
+    ('grp_ptr', int32[:]),
+    ('grp_indices', int32[:])
+]
+
+
+@jitclass(spec_SparseGroupL1)
+class SparseGroupL1(BasePenalty):
```
mathurinm, line 171: Start simple : implement a basic groupL1 first

skglm/solvers/group_cd.py
```diff
@@ -0,0 +1,148 @@
...
+from skglm.penalties.base import BasePenalty
+
+from skglm.penalties.block_separable import SparseGroupL1
+from skglm.datafits.multi_task import QuadraticGroup
+
+from skglm.utils import make_correlated_data, grp_converter
+from celer import GroupLasso
+
+
+FREQ_CHECK = 10
```
mathurinm, line 15: no need to expose them at submodule level, they can be defined inside `group_solver`

skglm/solvers/group_cd.py
```diff
@@ -0,0 +1,148 @@
...
+FREQ_ACC = 10
+
+
+def group_solver(X: np.ndarray, y: np.ndarray,
+                 datafit: BaseMultitaskDatafit, penalty: BasePenalty,
+                 max_iter: int = 1000, max_epochs: int = 100,
+                 stop_tol: float = 1e-7, p0: int = 10, use_acc: bool = True,
+                 verbose: bool = False) -> np.ndarray:
+
+    n_features: int = X.shape[1]
```
mathurinm, line 25: don't use type annotations for consistency with existing code

skglm/solvers/group_cd.py
```diff
@@ -0,0 +1,148 @@
...
+    grp_ptr, grp_indices = penalty.grp_ptr, penalty.grp_indices
+    n_groups: int = len(grp_ptr) - 1
+
+    # init
+    w = np.zeros(n_features, dtype=np.float64)
```
mathurinm, line 30: Is there a particular reason to specify float64 here ? It's the default type

skglm/solvers/group_cd.py
```diff
@@ -0,0 +1,148 @@
...
+    Xw = X @ w  # this would make sense when w is provided
+    datafit.initialize(X, y)
+    residuals = np.array([], dtype=np.float64)
+
+    for k in range(1, max_iter+1):
```
mathurinm, line 35: using range(max_iter)  feels more standard

skglm/solvers/group_cd.py
```diff
@@ -0,0 +1,148 @@
...
+        ws = np.arange(n_groups)
```
mathurinm, line 36: this is a redefinition at each iteration

skglm/solvers/group_cd.py
```diff
@@ -0,0 +1,148 @@
...
+        is_optimal, scores = _check_is_optimal(datafit, penalty, y, X, w, Xw,
+                                               ws, stop_tol)
+
+        if is_optimal:
+            print("Outer Solver: early exit")
+            break
+
+        ws, ws_size = _select_ws(w, n_groups, penalty, scores, p0)
```
mathurinm, line 44: avoid functions which are called only once

skglm/solvers/group_cd.py
```diff
@@ -0,0 +1,148 @@
...
+
+        if verbose == 1:
```
mathurinm, line 46: if verbose == 2 then nothing is printed, use verbose > 0

skglm/solvers/group_cd.py
```diff
@@ -0,0 +1,148 @@
...
+            p_obj = datafit.value(y, w, Xw) + penalty.value(w)
+            print(f"Iteration {k}: {p_obj}, support: {ws_size} / {n_groups}")
+
+        for epoch in range(1, max_epochs+1):
+            # inplace update of w
+            _cycle_group_cd(datafit, penalty, y, X,
```
mathurinm, line 56: harmonize names with existing code for fully separable penalties
also here you're doing bcd not cd

skglm/solvers/group_cd.py
```diff
@@ -0,0 +1,148 @@
...
+
+        if norm(w[grp_g_indices] - old_w_g) != 0:
+            Xw += X_g @ (w[grp_g_indices] - old_w_g)
+
+
+@njit
+def _check_is_optimal(datafit, penalty, y, X, w, Xw, ws, stop_tol):
+    grad = []
+    for g in ws:
+        grad.append(-datafit.gradient_j(X, y, w, Xw, g))
```
mathurinm, line 105: why use a list ?

mathurinm:
@Badr-MOUFAD I suggest you get a simple solver (no ws, no AA, no sparse support) to work in a first PR, that we merge, and we move on step by step in other PRs. 
Otherwise we have to review from scratch from one achievement to the other

Badr-MOUFAD:
@mathurinm,
As you requested, here is the implementation of the vanilla group solver.
I cross-tested it successfully against celer ``GroupLasso`` (have a look on ``tests/test_group.py``)

.github/workflows/main.yml
```diff
@@ -20,6 +20,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
+        python -m pip install git+https://github.com/mathurinm/celer.git
```
mathurinm, line 23: why adding this, normally the code is compiled locally with `pip install .`

skglm/datafits/group_task.py
```diff
@@ -0,0 +1,71 @@
+import numpy as np
```
mathurinm, line 1: the file should be named `group.py` not `group_task.py` as there are no tasks here

skglm/datafits/group_task.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+spec_QuadraticGroup = [
+    ('grp_partition', int32[:]),
+    ('grp_indices', int32[:]),
+    ('lipschitz', float64[:])
+]
+
+
+@ jitclass(spec_QuadraticGroup)
```
mathurinm, line 16: extra space after @

skglm/datafits/group_task.py
```diff
@@ -0,0 +1,71 @@
+import numpy as np
+from numpy.linalg import norm
+from numba.experimental import jitclass
+from numba import int32, float64
+
+from skglm.datafits.base import BaseDatafit
+
+
+spec_QuadraticGroup = [
+    ('grp_partition', int32[:]),
```
mathurinm, line 10: use `grp_ptr` to stay coherent with the scipy sparse format this is inspired from. ptr stands for "pointer" not "partition"

skglm/datafits/group_task.py
```diff
@@ -0,0 +1,71 @@
...
+
+    The datafit reads::
+
+    (1 / (2 * n_samples)) * ||y - X w||^2_2
+
+    where coordinates of ``w`` admits a group partitions.
+
+    Attributes
+    ----------
+    grp_partition : array, shape (n_groups + 1,)
```
mathurinm, line 28: change everywhere to grp_ptr

skglm/datafits/group_task.py
```diff
@@ -0,0 +1,71 @@
...
+    def initialize(self, X, y):
+        n_samples = len(y)
+        grp_partition, grp_indices = self.grp_partition, self.grp_indices
+        n_groups = len(grp_partition) - 1
+
+        lipschitz = np.zeros(n_groups)
+        for g in range(n_groups):
+            grp_g_indices = grp_indices[grp_partition[g]: grp_partition[g+1]]
+            X_g = X[:, grp_g_indices]
+
```
mathurinm, line 50: rm blank line

skglm/datafits/group_task.py
```diff
@@ -0,0 +1,71 @@
...
+            lipschitz[g] = norm(X_g) ** 2 / n_samples
+
+        self.lipschitz = lipschitz
+
+    def value(self, y, w, Xw):
+        n_samples = len(y)
+        return 1 / (2*n_samples) * norm(y - Xw) ** 2
```
mathurinm, line 57: cosmetic: put the division after the square norm to be more concise

skglm/datafits/group_task.py
```diff
@@ -0,0 +1,71 @@
...
+
+        grad_g = np.zeros(len(grp_g_indices))
+        for i, g in enumerate(grp_g_indices):
+            grad_g[i] = self.gradient_scalar(X, y, w, Xw, g)
+
+        return grad_g
+
+    def gradient_scalar(self, X, y, w, Xw, j):
+        n_samples = len(y)
+        return - 1 / n_samples * X[:, j].T @ (y - Xw)
```
mathurinm, line 71: avoid double minuses: Xj @ (Xw - y)

also transposing 1D arrays is not necessary

skglm/penalties/block_separable.py
```diff
@@ -1,6 +1,7 @@
 import numpy as np
-from numpy.linalg.linalg import norm
-from numba import float64
+from numpy.linalg import norm
+
```
mathurinm, line 3: avoid such changes

skglm/penalties/block_separable.py
```diff
@@ -153,3 +154,81 @@ def subdiff_distance(self, W, grad, ws):
...
+]
+
+
+@jitclass(spec_WeightedGroupL1)
+class WeightedGroupL1(BasePenalty):
+    r"""Weighted Group L1 penalty.
+
+    The penalty reads::
+
+        \alpha * \sum_{g} weights[g] * ||w_g||_2
```
mathurinm, line 173: no alpha ?

skglm/penalties/block_separable.py
```diff
@@ -153,3 +154,81 @@ def subdiff_distance(self, W, grad, ws):
...
+            grp_g_indices = grp_indices[grp_partition[g]: grp_partition[g+1]]
+            w_g = w[grp_g_indices]
+
+            sum_weighted_L2 += alpha * weights[g] * norm(w_g)
+
+        return sum_weighted_L2
+
+    def prox_1group(self, value, stepsize, g):
+        """Compute the proximal operator of a given group."""
+        alpha = self.alpha
```
mathurinm, line 212: no need to declare alpha here, directly use self.alpha L213

skglm/solvers/group_bcd_solver.py
```diff
@@ -0,0 +1,100 @@
+import numpy as np
+from numba import njit
+
+
+def grp_bcd_solver(X, y, datafit, penalty,
```
mathurinm, line 5: you can call it bcd_solver, bcd must be fore group (b stands for Block)

skglm/solvers/group_bcd_solver.py
```diff
@@ -0,0 +1,100 @@
...
+    -------
+    w : array, shape (n_features,)
+        Vector that minimizes ``datafit() + penalty()``.
+    """
+    n_samples, n_features = X.shape
+    n_groups = len(penalty.grp_partition) - 1
+
+    # init
+    w = np.zeros(n_features)
+    Xw = np.zeros(n_samples)  # X @ w in general
```
mathurinm, line 42: why in general ?

skglm/solvers/group_bcd_solver.py
```diff
@@ -0,0 +1,100 @@
...
+    Returns
+    -------
+    w : array, shape (n_features,)
+        Vector that minimizes ``datafit() + penalty()``.
+    """
+    n_samples, n_features = X.shape
+    n_groups = len(penalty.grp_partition) - 1
+
+    # init
+    w = np.zeros(n_features)
```
mathurinm, line 41: Can you add support for w_init ?

skglm/solvers/group_bcd_solver.py
```diff
@@ -0,0 +1,100 @@
...
+    verbose : bool, default False
+        Log or not the objective at each iteration.
+
+    Returns
+    -------
+    w : array, shape (n_features,)
+        Vector that minimizes ``datafit() + penalty()``.
+    """
+    n_samples, n_features = X.shape
+    n_groups = len(penalty.grp_partition) - 1
```
mathurinm, line 38: maybe check that penalty has the attributes needed for a group, and raise a ValueError otherwise ? your call

skglm/solvers/group_bcd_solver.py
```diff
@@ -0,0 +1,100 @@
...
+    # init
+    w = np.zeros(n_features)
+    Xw = np.zeros(n_samples)  # X @ w in general
+    datafit.initialize(X, y)
+    all_groups = np.arange(n_groups)
+
+    for k in range(max_iter):
+        prev_p_obj = datafit.value(y, w, Xw) + penalty.value(w)
+        _bcd_epoch(X, y, w, Xw, datafit, penalty, all_groups)
+        current_obj = datafit.value(y, w, Xw) + penalty.value(w)
```
mathurinm, line 49: harmonize obj and p_obj

skglm/solvers/group_bcd_solver.py
```diff
@@ -0,0 +1,100 @@
...
+
+    # init
+    w = np.zeros(n_features)
+    Xw = np.zeros(n_samples)  # X @ w in general
+    datafit.initialize(X, y)
+    all_groups = np.arange(n_groups)
+
+    for k in range(max_iter):
+        prev_p_obj = datafit.value(y, w, Xw) + penalty.value(w)
+        _bcd_epoch(X, y, w, Xw, datafit, penalty, all_groups)
```
mathurinm, line 48: do 10 passes rather than 1 to maintain consistency with other solvers (minor)

skglm/solvers/group_bcd_solver.py
```diff
@@ -0,0 +1,100 @@
...
+            break
+
+    return w
+
+
+@njit
+def _bcd_epoch(X, y, w, Xw, datafit, penalty, ws):
+    """Perform a single BCD epoch on groups in ws.
+
+    Parameters
```
mathurinm, line 65: IMO, we don't need to document private functions. It induces a higher maintenance cost

skglm/solvers/group_bcd_solver.py
```diff
@@ -0,0 +1,100 @@
...
+    ----------
+    X : array, shape (n_samples, n_features)
+        Design matrix.
+    y : array, shape (n_samples,)
+        Target vector.
+    w : array, shape (n_features,)
+        Vector that minimizes ``datafit() + penalty()``.
```
mathurinm, line 72: for example this should be .value(), etc. So I'd remove the whole docstring

skglm/solvers/group_bcd_solver.py
```diff
@@ -0,0 +1,100 @@
...
+    Xw : array, shape (n_samples,)
+        X @ w in general.
+    datafit : QuadraticGroup
```
mathurinm, line 75: same, this doesn't need to be Quadratic

skglm/tests/test_group.py
```diff
@@ -0,0 +1,106 @@
...
+from skglm.utils import grp_converter, make_correlated_data
+
+
+def test_alpha_max():
+    n_samples, n_features = 100, 1000
+    random_state = 1563
+    rnd = np.random.RandomState(random_state)
+    X, y, _ = make_correlated_data(n_samples, n_features, random_state=random_state)
+
+    groups = 20  # contiguous groups of 20 elements
```
mathurinm, line 18: perfect, can you add checks with the other `group` formats ?

skglm/tests/test_group.py
```diff
@@ -0,0 +1,106 @@
...
+def test_equivalence_lasso():
+    n_samples, n_features = 100, 1000
+    random_state = 1123
+    rnd = np.random.RandomState(random_state)
+    X, y, _ = make_correlated_data(n_samples, n_features, random_state=random_state)
+
+    grp_partition, grp_indices = grp_converter(1, n_features)
+    weights = abs(rnd.randn(n_features))
+
+    alpha_max = norm(X.T@y, ord=np.inf) / n_samples
```
mathurinm, line 50: shouldn't weights be involved here?

skglm/tests/test_group.py
```diff
@@ -0,0 +1,106 @@
...
+    np.testing.assert_allclose(celer_lasso.coef_, w_group_solver, atol=1e-4, rtol=1e-3)
+
+
+def test_vs_celer_GroupLasso():
+    n_samples, n_features = 100, 1000
+    random_state = 42
+    rnd = np.random.RandomState(random_state)
+    X, y, _ = make_correlated_data(n_samples, n_features, random_state=random_state)
+
+    groups = 10  # contiguous groups of 10 elements
```
mathurinm, line 76: same here, can you test varying groups too ?

mathurinm:
Great work, now the PR is easier to review and in close to being merged. A few comments left

.github/workflows/main.yml
Badr-MOUFAD: This is to install ``celer`` for CI tests.
I wanted to release from running the test locally and expose the result directly.

skglm/solvers/group_bcd_solver.py
Badr-MOUFAD: case ``w`` provided initially

.github/workflows/main.yml
mathurinm: my bad

debug_script.py
```diff
@@ -0,0 +1,52 @@
...
+from numpy.linalg import norm
+
+from skglm.penalties.block_separable import WeightedGroupL1
+from skglm.datafits.group import QuadraticGroup
+from skglm.solvers.group_bcd_solver import bcd_solver
+
+from skglm.utils import grp_converter, make_correlated_data
+
+
+def _generate_random_grp(n_groups, n_features, random_state=123654):
```
mathurinm, line 11: no need to exposte the parameter random_state. One parameter fewer makes the code easier, you can hardcode it

skglm/solvers/group_bcd_solver.py
```diff
@@ -0,0 +1,86 @@
...
+
+    for k in range(max_iter):
+        prev_p_obj = datafit.value(y, w, Xw) + penalty.value(w)
+        _bcd_epoch(X, y, w, Xw, datafit, penalty, all_groups)
+        current_p_obj = datafit.value(y, w, Xw) + penalty.value(w)
+
+        if verbose:
+            print(f"Iteration {k}: {current_p_obj}")
+
+        if np.abs(current_p_obj - prev_p_obj) < tol:  # naive stopping criterion
```
mathurinm, line 58: use <=

skglm/penalties/block_separable.py
```diff
@@ -153,3 +154,82 @@ def subdiff_distance(self, W, grad, ws):
...
+        alpha, weights = self.alpha, self.weights
+        grp_ptr, grp_indices = self.grp_ptr, self.grp_indices
+
+        scores = np.zeros(len(ws))
+        for idx, g in enumerate(ws):
+            grad_g = grad[idx]
+
+            grp_g_indices = grp_indices[grp_ptr[g]: grp_ptr[g+1]]
+            w_g = w[grp_g_indices]
+            norm_w_g = norm(w_g)
```
Badr-MOUFAD, line 227: @mathurinm, same here

skglm/penalties/block_separable.py
```diff
@@ -153,3 +154,82 @@ def subdiff_distance(self, W, grad, ws):
...
+    def value(self, w):
+        """Value of penalty at vector w."""
+        alpha, weights = self.alpha, self.weights
+        grp_ptr, grp_indices = self.grp_ptr, self.grp_indices
+        n_grp = len(grp_ptr) - 1
+
+        sum_weighted_L2 = 0.
+        for g in range(n_grp):
+            grp_g_indices = grp_indices[grp_ptr[g]: grp_ptr[g+1]]
+            w_g = w[grp_g_indices]
```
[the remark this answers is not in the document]
mathurinm: it's a tradeoff between code simplicity and speed. I'd hope that the cost is dominated by the dot product to compute the gradient. Do you observe speed differences if you loop over elements of the group directly ?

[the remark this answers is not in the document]
Badr-MOUFAD: Hard to tell, but I'm more with code simplicity and readability.

skglm/tests/test_group.py
```diff
@@ -0,0 +1,126 @@
...
+from skglm.penalties.block_separable import WeightedGroupL1
+from skglm.datafits.group import QuadraticGroup
+from skglm.solvers.group_bcd_solver import bcd_solver
+
+from skglm.utils import grp_converter, make_correlated_data
+from celer import GroupLasso, Lasso
+
+
+def _generate_random_grp(n_groups, n_features):
+    all_features = np.arange(n_features)
```
mathurinm, line 15: It seems to me you can generate directly grp_ptr and grp_indices ? grp_indices is a permutation, grp_ptr is a np.random.choice(.. replace=False) that you sort and to which you add 0 and n_features

Because the first things we do after calling this function is to call grp_converter

skglm/tests/test_group.py
```diff
@@ -0,0 +1,126 @@
...
+
+    w_group_solver = bcd_solver(
+        X, y, quad_group, group_penalty, max_iter=10000,
+        verbose=True, tol=0)
+
+    np.testing.assert_almost_equal(norm(w_group_solver), 0, decimal=10)
+
+
+def test_equivalence_lasso():
+    n_samples, n_features = 100, 1000
```
mathurinm, line 59: use smaller datasets for faster tests and easier debug; 30 x 50 is enough

skglm/tests/test_group.py
```diff
@@ -0,0 +1,126 @@
...
+
+    # group solver
+    quad_group = QuadraticGroup(grp_ptr=grp_ptr, grp_indices=grp_indices)
+    group_penalty = WeightedGroupL1(
+        alpha=alpha_max, grp_ptr=grp_ptr,
+        grp_indices=grp_indices, weights=weights)
+
+    w_group_solver = bcd_solver(
+        X, y, quad_group, group_penalty, max_iter=10000,
+        verbose=True, tol=0)
```
mathurinm, line 53: you can silence the solvers (otherwise when they fail the stacktrace is huge)

skglm/solvers/group_bcd_solver.py
```diff
@@ -0,0 +1,86 @@
...
+
+    Parameters
+    ----------
+    X : array, shape (n_samples, n_features)
+        Design matrix.
+
+    y : array, shape (n_samples,)
+        Target vector.
+
+    datafit : QuadraticGroup
```
PABannier, line 17: cosmetic: `instance of QuadraticGroup`

skglm/solvers/group_bcd_solver.py
```diff
@@ -0,0 +1,86 @@
...
+        DataFit object.
+
+    penalty : WeightedGroupL1
```
PABannier, line 20: `instance of WeightedGroupL1`

skglm/datafits/group.py
```diff
@@ -0,0 +1,72 @@
...
+    def value(self, y, w, Xw):
+        n_samples = len(y)
+        return norm(y - Xw) ** 2 / (2 * n_samples)
+
+    def gradient_g(self, X, y, w, Xw, g):
+        grp_ptr, grp_indices = self.grp_ptr, self.grp_indices
+        grp_g_indices = grp_indices[grp_ptr[g]: grp_ptr[g+1]]
+
+        grad_g = np.zeros(len(grp_g_indices))
+        for i, g in enumerate(grp_g_indices):
```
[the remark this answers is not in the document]
Badr-MOUFAD: Totally agree with you. On the other hand, I think ``j, g_j`` is more readable/relatable than ``i, j``

skglm/tests/test_group.py
Badr-MOUFAD: I agree with you but I think the current design enables us to condense 3 test configurations within one unit testing.

PABannier (commented):
a few minor cosmetic comments on my side. Thanks a lot @Badr-MOUFAD !

skglm/tests/test_group.py
mathurinm: what do you mean?

skglm/solvers/group_bcd_solver.py
mathurinm: Btw why should it be such an instance and not a general Penalty?

skglm/datafits/group.py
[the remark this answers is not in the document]
mathurinm: j is reserved for features. use idx if you want, but remain consistent with existing code

skglm/solvers/group_bcd_solver.py
PABannier: you're right: `instance of Penalty`

skglm/penalties/block_separable.py
```diff
@@ -153,3 +154,82 @@ def subdiff_distance(self, W, grad, ws):
...
+
+            sum_weighted_L2 += alpha * weights[g] * norm(w_g)
```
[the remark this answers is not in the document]
Badr-MOUFAD: I will rather use ``norm`` to avoid confusion.

skglm/solvers/group_bcd_solver.py
```diff
@@ -0,0 +1,115 @@
...
+    verbose : bool, default False
+        Amount of verbosity. 0/False is silent.
+
+    Returns
+    -------
+    w : array, shape (n_features,)
+        Solution that minimizes the problem defined by datafit and penalty.
+
+    p_objs_out: array
+        Array of p_obj values at each iteration.
```
PABannier, line 45:
```suggestion
        The objective values at every outer iteration.
```

skglm/solvers/group_bcd_solver.py
```diff
@@ -0,0 +1,115 @@
...
+            if epoch % 10 == 0:
+                current_p_obj = datafit.value(y, w, Xw) + penalty.value(w)
+                stop_crit = np.abs(current_p_obj - prev_p_obj)
+                if stop_crit <= tol:
+                    print("Inner Solver: Early exit")
+                    break
+                prev_p_obj = current_p_obj
+
+        current_p_obj = datafit.value(y, w, Xw) + penalty.value(w)
+        if verbose > 0:
```
PABannier, line 80:
```suggestion
        if max(verbose - 1, 0):
```

skglm/solvers/group_bcd_solver.py
```diff
@@ -0,0 +1,115 @@
...
+            print(f"Iteration {t}: {current_p_obj}")
```
PABannier, line 81:
```suggestion
            print(f"Epoch {epoch + 1}, objective {current_p_obj:.10f}, "
                          f"stopping crit {stop_crit:.2e}")
```

skglm/datafits/group.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def value(self, y, w, Xw):
+        return norm(y - Xw) ** 2 / (2 * len(y))
+
+    def gradient_g(self, X, y, w, Xw, g):
+        grp_ptr, grp_indices = self.grp_ptr, self.grp_indices
+        grp_g_indices = grp_indices[grp_ptr[g]: grp_ptr[g+1]]
+
+        grad_g = np.zeros(len(grp_g_indices))
+        for idx, g_j in enumerate(grp_g_indices):
```
mathurinm, line 61: use j instead of g_j. j is standard in the lib for features, here it's a feature.

skglm/penalties/block_separable.py
```diff
@@ -153,3 +154,81 @@ def subdiff_distance(self, W, grad, ws):
...
+    ('alpha', float64),
+    ('weights', float64[:]),
+    ('grp_ptr', int32[:]),
+    ('grp_indices', int32[:]),
+]
+
+
+@jitclass(spec_WeightedGroupL1)
+class WeightedGroupL1(BasePenalty):
+    r"""Weighted Group L1 penalty.
```
mathurinm, line 169: It's a group L2 penalty in fact, it needs to be renamed here and in the doc

skglm/solvers/group_bcd_solver.py
```diff
@@ -0,0 +1,119 @@
...
+    """
+    n_features = X.shape[1]
+    n_groups = len(penalty.grp_ptr) - 1
+
+    # init
+    w = w_init or np.zeros(n_features)
+    Xw = X @ w
+    datafit.initialize(X, y)
+    all_groups = np.arange(n_groups)
+    p_objs_out = np.array([])
```
mathurinm, line 58: preassign to np.zeros(max_iter)

skglm/solvers/group_bcd_solver.py
```diff
@@ -0,0 +1,119 @@
...
+    for t in range(max_iter):
+        if t == 0:  # avoid computing p_obj twice
+            prev_p_obj = datafit.value(y, w, Xw) + penalty.value(w)
+
+        for epoch in range(max_epochs):
+            _bcd_epoch(X, y, w, Xw, datafit, penalty, all_groups)
+
+            if max(verbose - 1, 0):
+                current_p_obj = datafit.value(y, w, Xw) + penalty.value(w)
+                print(f"\t | Epoch {epoch+1}: {current_p_obj}")
```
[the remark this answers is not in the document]
mathurinm: don't print at every epoch, the obj is updated every 10 epochs only. Move to the if below

skglm/solvers/group_bcd_solver.py
```diff
@@ -0,0 +1,119 @@
...
+
+            if epoch % 10 == 0:
+                current_p_obj = datafit.value(y, w, Xw) + penalty.value(w)
+                stop_crit = np.abs(current_p_obj - prev_p_obj)
+                if stop_crit <= 0.3 * tol:
```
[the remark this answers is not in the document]
Badr-MOUFAD: it's just to align with other solvers

skglm/solvers/group_bcd_solver.py
```diff
@@ -0,0 +1,119 @@
...
+            )
+
+        if stop_crit <= tol:
+            print("Outer solver: Early exit")
+            break
+
+        prev_p_obj = current_p_obj
+        p_objs_out = np.append(p_objs_out, current_p_obj)
+
+    return w, p_objs_out, stop_crit
```
mathurinm, line 95: Check but in other solvers we return an array of stop_crits I believe

skglm/solvers/group_bcd_solver.py
```diff
@@ -0,0 +1,119 @@
...
+            if epoch % 10 == 0:
+                current_p_obj = datafit.value(y, w, Xw) + penalty.value(w)
+                stop_crit = np.abs(current_p_obj - prev_p_obj)
+                if stop_crit <= 0.3 * tol:
+                    print("Early exit")
+                    break
+                prev_p_obj = current_p_obj
+
+        current_p_obj = datafit.value(y, w, Xw) + penalty.value(w)
+        stop_crit = np.abs(current_p_obj - prev_p_obj)
```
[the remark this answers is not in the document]
Badr-MOUFAD: to avoid checking a negative number with ``tol``

skglm/solvers/group_bcd_solver.py
```diff
@@ -0,0 +1,119 @@
...
+
+        lipschitz_g = datafit.lipschitz[g]
+        grad_g = datafit.gradient_g(X, y, w, Xw, g)
+
+        w[grp_g_indices] = penalty.prox_1group(
+            old_w_g - grad_g / lipschitz_g,
+            1 / lipschitz_g, g
+        )
+
+        # update Xw without copying w_g
```
mathurinm, line 115: we do copy w_g, no ? I'm not sure I understand this comment

skglm/solvers/group_bcd_solver.py
```diff
@@ -0,0 +1,119 @@
...
+        for idx, g_j in enumerate(grp_g_indices):
```
mathurinm, line 116: use j

skglm/solvers/group_bcd_solver.py
Badr-MOUFAD: we only return the last stop_crit

https://github.com/scikit-learn-contrib/skglm/blob/832a57a66d16cda53698721000b45cf800ae45f8/skglm/solvers/cd_solver.py#L347

skglm/solvers/group_bcd_solver.py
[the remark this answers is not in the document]
mathurinm: It's a descent method, you should not get negative value right?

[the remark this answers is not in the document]
mathurinm: It seems like this tries to hide an error that should not happen. If you get a negative quantity, there's something else at play (like precision errors, but in that case you've already converged)

skglm/solvers/group_bcd_solver.py
mathurinm: and what is old_w_g if not a copy of w_g ? :D

skglm/solvers/group_bcd_solver.py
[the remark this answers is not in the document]
mathurinm: Ok it's the inner solver. The criterion is rather 0.3 * current outer value for stop_crit, no ?

skglm/solvers/group_bcd_solver.py
Badr-MOUFAD: old_w_g is a copy of the old_w_g

skglm/solvers/group_bcd_solver.py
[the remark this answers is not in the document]
Badr-MOUFAD: That right,
then it's, 

```python
stop_crit = prev_p_obj - current_p_obj
```

skglm/solvers/group_bcd_solver.py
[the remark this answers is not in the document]
Badr-MOUFAD: it's literally the stop criteria that was passed as arg

https://github.com/scikit-learn-contrib/skglm/blob/832a57a66d16cda53698721000b45cf800ae45f8/skglm/solvers/cd_solver.py#L342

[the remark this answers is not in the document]
Badr-MOUFAD: I should rather name it stop_crit_in to avoid ambiguity

skglm/solvers/group_bcd_solver.py
mathurinm: Yes, so there is a copy involved, remove the comment. It's the straightforward way of doing it, it does not deserve a comment

skglm/solvers/group_bcd_solver.py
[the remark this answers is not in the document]
Badr-MOUFAD: here I am dealing with all features

https://github.com/scikit-learn-contrib/skglm/blob/832a57a66d16cda53698721000b45cf800ae45f8/skglm/solvers/cd_solver.py#L338-L345

so it should be ``tol``

skglm/solvers/group_bcd_solver.py
mathurinm: indeed

skglm/solvers/group_bcd_solver.py
```diff
@@ -0,0 +1,121 @@
...
+                current_p_obj = datafit.value(y, w, Xw) + penalty.value(w)
+                stop_crit_in = prev_p_obj - current_p_obj
+
+                if max(verbose - 1, 0):
+                    print(
+                        f"Epoch {epoch+1}: {current_p_obj:.10f}"
+                        f"stopping crit: {stop_crit_in:.2f}"
+                    )
+
+                if stop_crit_in <= 0.3 * tol:
```
mathurinm, line 77: if you're in the inner loop, the criterion is to stop if stop_crit_in is lower than tol, or 0.3 * stop_crit_out

If you're in the outer loop, the criterion is stop_crit(_out) < tol


The current design is inefficient. Once you reach tol * 0.3, you'll exit from the inner loop, but not from the outer, and you'll do that a lot of time.

Replace by <= tol here since you're not using working sets yet anyway.

skglm/tests/test_group.py
```diff
@@ -0,0 +1,116 @@
...
+
+from skglm.penalties.block_separable import WeightedGroupL2
+from skglm.datafits.group import QuadraticGroup
+from skglm.solvers.group_bcd_solver import bcd_solver
+
+from skglm.utils import grp_converter, make_correlated_data
+from celer import GroupLasso, Lasso
+
+
+def _generate_random_grp(n_groups, n_features, shuffle=True):
```
mathurinm, line 13: this is not seeded and makes the test non deterministic (for the shuffle, and for the choice)

skglm/tests/test_group.py
```diff
@@ -0,0 +1,116 @@
...
+    splits[0], splits[-1] = 0, n_features
+
+    groups = [list(all_features[splits[i]: splits[i+1]])
+              for i in range(n_groups)]
+
+    return all_features, splits, groups   # grp_indices, grp_prt, groups
+
+
+@pytest.mark.parametrize("n_groups, n_features, shuffle",
+                         [[10, 50, True], [10, 50, False], [17, 53, False]])
```
mathurinm, line 29: can one configuration be removed in order to have faster tests ? It seems they are redundant

mathurinm:
Thanks @Badr-MOUFAD ! One very nice step :)

mathurinm:
@MPenaR may be interested in this solver

MPenaR:
thank you

PR #27, opened 2022-05-31
FIX wrong return types in docstrings

PABannier:
While reviewing #26 I found that the docstrings for `cd_solver` and `bcd_solver` were wrong. I fixed it.


mathurinm (approved):

mathurinm:
Thank you @PABannier

PR #28, opened 2022-06-03
[READY] ENH - add working set strategy to ``group_bcd_solver``

Badr-MOUFAD:
This continues #26

It will proceed as follows
- [x] Implement working sets
- [x] Benchmark against ``main`` branch

### DIFF
--- a/skglm/penalties/block_separable.py
+++ b/skglm/penalties/block_separable.py
@@ -213,8 +213,9 @@ class WeightedGroupL2(BasePenalty):
         return BST(value, self.alpha * stepsize * self.weights[g])
 
     def subdiff_distance(self, w, grad_ws, ws):
-        """Compute distance of negative gradient restricted to groups in ``ws``
-        to the subdifferential at ``w``.
+        """
+        Compute distance to the subdifferential at ``w`` of 
+        negative gradient restricted to groups in ``ws``.
 
         Note: ``grad_ws`` is a stacked array of ``-``gradients.
         ([-grad_ws_1, -grad_ws_2, ...])
--- a/skglm/penalties/block_separable.py
+++ b/skglm/penalties/block_separable.py
@@ -218,10 +218,13 @@ class WeightedGroupL2(BasePenalty):
         grp_ptr, grp_indices = self.grp_ptr, self.grp_indices
 
         scores = np.zeros(len(ws))
+        grad_ptr = 0
         for idx, g in enumerate(ws):
-            grad_g = grad[idx]
-
             grp_g_indices = grp_indices[grp_ptr[g]: grp_ptr[g+1]]
+
+            grad_g = grad[grad_ptr: grad_ptr + len(grp_g_indices)]
+            grad_ptr += len(grp_g_indices)
+
             w_g = w[grp_g_indices]
             norm_w_g = norm(w_g)
 
--- a/skglm/penalties/block_separable.py
+++ b/skglm/penalties/block_separable.py
@@ -212,8 +212,13 @@ class WeightedGroupL2(BasePenalty):
         """Compute the proximal operator of group ``g``."""
         return BST(value, self.alpha * stepsize * self.weights[g])
 
-    def subdiff_distance(self, w, grad, ws):
-        """Compute distance of negative gradient to the subdifferential at ``w``."""
+    def subdiff_distance(self, w, grad_ws, ws):
+        """Compute distance of negative gradient restricted to groups in ``ws``
+        to the subdifferential at ``w``.
+
+        Note: ``grad_ws`` is a stacked array of ``-``gradients.
+        ([-grad_ws_1, -grad_ws_2, ...])
+        """
         alpha, weights = self.alpha, self.weights
         grp_ptr, grp_indices = self.grp_ptr, self.grp_indices
 
@@ -222,7 +227,7 @@ class WeightedGroupL2(BasePenalty):
         for idx, g in enumerate(ws):
             grp_g_indices = grp_indices[grp_ptr[g]: grp_ptr[g+1]]
 
-            grad_g = grad[grad_ptr: grad_ptr + len(grp_g_indices)]
+            grad_g = grad_ws[grad_ptr: grad_ptr + len(grp_g_indices)]
             grad_ptr += len(grp_g_indices)
 
             w_g = w[grp_g_indices]
--- a/skglm/penalties/block_separable.py
+++ b/skglm/penalties/block_separable.py
@@ -213,9 +213,7 @@ class WeightedGroupL2(BasePenalty):
         return BST(value, self.alpha * stepsize * self.weights[g])
 
     def subdiff_distance(self, w, grad_ws, ws):
-        """
-        Compute distance to the subdifferential at ``w`` of
-        negative gradient restricted to groups in ``ws``.
+        """Compute distance to the subdifferential at ``w`` of negative gradient.
 
         Note: ``grad_ws`` is a stacked array of ``-``gradients.
         ([-grad_ws_1, -grad_ws_2, ...])
--- a/skglm/penalties/block_separable.py
+++ b/skglm/penalties/block_separable.py
@@ -255,8 +255,7 @@ class WeightedGroupL2(BasePenalty):
                 continue
 
             grp_g_indices = grp_indices[grp_ptr[g]: grp_ptr[g+1]]
-            w_g = w[grp_g_indices]
-            if np.any(w_g):
+            if np.any(w[grp_g_indices]):
                 gsupp[g] = True
 
         return gsupp
--- a/skglm/penalties/block_separable.py
+++ b/skglm/penalties/block_separable.py
@@ -242,11 +242,11 @@ class WeightedGroupL2(BasePenalty):
     def generalized_support(self, w):
         grp_indices, grp_ptr = self.grp_indices, self.grp_ptr
         n_groups = len(grp_ptr) - 1
-        is_not_penalized = ~self.is_penalized(n_groups)
+        is_penalized = self.is_penalized(n_groups)
 
         gsupp = np.zeros(n_groups, dtype=np.bool_)
         for g in range(n_groups):
-            if is_not_penalized[g]:
+            if not is_penalized[g]:
                 gsupp[g] = True
                 continue
 
--- a/skglm/penalties/block_separable.py
+++ b/skglm/penalties/block_separable.py
@@ -232,3 +232,24 @@ class WeightedGroupL2(BasePenalty):
                 scores[idx] = norm(grad_g - subdiff)
 
         return scores
+
+    def is_penalized(self, n_groups):
+        return np.ones(n_groups, dtype=np.bool_)
+
+    def generalized_support(self, w):
+        grp_indices, grp_ptr = self.grp_indices, self.grp_ptr
+        n_groups = len(grp_ptr) - 1
+        is_not_penalized = ~self.is_penalized(n_groups)
+
+        gsupp = np.zeros(n_groups, dtype=np.bool_)
+        for g in range(n_groups):
+            if is_not_penalized[g]:
+                gsupp[g] = True
+                continue
+
+            grp_g_indices = grp_indices[grp_ptr[g]: grp_ptr[g+1]]
+            w_g = w[grp_g_indices]
+            if np.any(w_g):
+                gsupp[g] = True
+
+        return gsupp
--- a/skglm/penalties/block_separable.py
+++ b/skglm/penalties/block_separable.py
@@ -214,7 +214,7 @@ class WeightedGroupL2(BasePenalty):
 
     def subdiff_distance(self, w, grad_ws, ws):
         """
-        Compute distance to the subdifferential at ``w`` of 
+        Compute distance to the subdifferential at ``w`` of
         negative gradient restricted to groups in ``ws``.
 
         Note: ``grad_ws`` is a stacked array of ``-``gradients.
--- a/skglm/solvers/group_bcd_solver.py
+++ b/skglm/solvers/group_bcd_solver.py
@@ -1,11 +1,8 @@
 import numpy as np
 from numba import njit
 
-from skglm.datafits.group import QuadraticGroup
-from skglm.penalties.block_separable import WeightedGroupL2
 
-
-def bcd_solver(X, y, datafit, penalty: WeightedGroupL2, w_init=None, p0=2,
+def bcd_solver(X, y, datafit, penalty, w_init=None, p0=2,
                max_iter=1000, max_epochs=100, tol=1e-7, verbose=False):
     """Run a group BCD solver.
 
@@ -141,7 +138,7 @@ def _bcd_epoch(X, y, w, Xw, datafit, penalty, ws):
     return
 
 
-def _construct_grad(X, y, w, Xw, datafit: QuadraticGroup, ws):
+def _construct_grad(X, y, w, Xw, datafit, ws):
     """Compute the -gradient according to each group in ``ws``.
 
     Note: -gradients are stacked in a 1d array (e.g. [-grad_g1, -grad_g2, ...]).
--- a/skglm/solvers/group_bcd_solver.py
+++ b/skglm/solvers/group_bcd_solver.py
@@ -150,15 +150,3 @@ def _construct_grad(X, y, w, Xw, datafit: QuadraticGroup, ws):
         grad_g = -datafit.gradient_g(X, y, w, Xw, g)
         grads = np.concatenate((grads, grad_g))
     return grads
-
-
-def _construct_ws(w, opt, penalty: WeightedGroupL2, p0):
-    """Construct working set."""
-    n_features = len(w)
-    gsupp_size = penalty.generalized_support(w).sum()
-
-    ws_size = max(min(p0, n_features),
-                  min(n_features, 2 * gsupp_size))
-
-    ws = np.argpartition(opt, -ws_size)[-ws_size:]
-    return ws
--- a/skglm/solvers/group_bcd_solver.py
+++ b/skglm/solvers/group_bcd_solver.py
@@ -59,7 +59,7 @@ def bcd_solver(X, y, datafit, penalty, w_init=None, p0=10,
     datafit.initialize(X, y)
     all_groups = np.arange(n_groups)
     p_objs_out = np.zeros(max_iter)
-    stop_crit = 0  # prevent ref before assignment when max_iter == 0
+    stop_crit = 0.  # prevent ref before assign when max_iter == 0
 
     for t in range(max_iter):
         if t == 0:  # avoid computing grad and opt twice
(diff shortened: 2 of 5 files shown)


Badr-MOUFAD:
I did some benchmarks, thanks to [``benchopts``](https://github.com/benchopt/benchopt), against the main branch which doesn't implement working set strategy

Here are my results:

<details>
<summary>
Results for alpha / alpha_max = 0.1
</summary>

rho=01

</details>

<details>
<summary>
Results for alpha / alpha_max = 0.01
</summary>

rho001

</details>

<details>
<summary>
Results for alpha / alpha_max = 0.001
</summary>

rho0001
</details>

@mathurinm @PABannier, shall we proceed forward to adding Anderson acceleration?

skglm/penalties/block_separable.py
```diff
@@ -218,10 +218,13 @@ def subdiff_distance(self, w, grad, ws):
         grp_ptr, grp_indices = self.grp_ptr, self.grp_indices
 
         scores = np.zeros(len(ws))
+        grad_ptr = 0
         for idx, g in enumerate(ws):
-            grad_g = grad[idx]
-
             grp_g_indices = grp_indices[grp_ptr[g]: grp_ptr[g+1]]
+
+            grad_g = grad[grad_ptr: grad_ptr + len(grp_g_indices)]
```
mathurinm, line 225: If I understand correctly  this assumes that groups are contiguous, but it should fail for other ones. Why not use grad_g = grad[grp_g_indices] ?

skglm/penalties/block_separable.py
```diff
@@ -232,3 +235,24 @@ def subdiff_distance(self, w, grad, ws):
...
 
         return scores
+
+    def is_penalized(self, n_groups):
+        return np.ones(n_groups, dtype=np.bool_)
+
+    def generalized_support(self, w):
+        grp_indices, grp_ptr = self.grp_indices, self.grp_ptr
+        n_groups = len(grp_ptr) - 1
+        is_not_penalized = ~self.is_penalized(n_groups)
```
mathurinm, line 245: no need to assign this array, negate the condition L249 instead

skglm/solvers/group_bcd_solver.py
```diff
@@ -119,3 +139,15 @@ def _bcd_epoch(X, y, w, Xw, datafit, penalty, ws):
...
                 Xw += (w[j] - old_w_g[idx]) * X[:, j]
     return
+
+
+def _construct_grad(X, y, w, Xw, datafit: QuadraticGroup, ws):
+    """Compute the -gradient according to each group in ``ws``.
+
+    Note: -gradients are stacked in a 1d array (e.g. [-grad_g1, -grad_g2, ...]).
+    """
+    grads = np.array([])
```
mathurinm, line 146: here, preallocate an array of size n_features and fill it

Badr-MOUFAD: - ``_compute_grad`` act on a working set rather than all groups, hence it's not relevant to init grad with ``n_features``
- groups have not necessarily equal sizes hence computing the actual size of ``grad`` requires additional computation

PABannier: @Badr-MOUFAD can't you do something like this: `grads = np.zeros(len(ws))`?

skglm/solvers/group_bcd_solver.py
```diff
@@ -119,3 +139,15 @@ def _bcd_epoch(X, y, w, Xw, datafit, penalty, ws):
...
+    for g in ws:
+        grad_g = -datafit.gradient_g(X, y, w, Xw, g)
+        grads = np.concatenate((grads, grad_g))
+    return grads
```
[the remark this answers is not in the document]
PABannier: agreed.
@Badr-MOUFAD see: https://github.com/scikit-learn-contrib/skglm/blob/de7af7cc2eb41174f4feebdac69db1c43a1be9e8/skglm/solvers/cd_solver.py#L386
Would it be any different with groups?

skglm/solvers/group_bcd_solver.py
mathurinm: Could you do grads = np.zeros([n_features_in_ws]) ? and a simple computation for `n_features_in_ws` ?

skglm/solvers/group_bcd_solver.py
```diff
@@ -2,7 +2,7 @@
 from numba import njit
 
 
-def bcd_solver(X, y, datafit, penalty, w_init=None,
+def bcd_solver(X, y, datafit, penalty, w_init=None, p0=2,
```
mathurinm, line 5: use 10 for coherence with existing solvers

skglm/solvers/group_bcd_solver.py
```diff
@@ -24,6 +24,9 @@ def bcd_solver(X, y, datafit, penalty, w_init=None,
         Initial value of coefficients.
         If set to None, a zero vector is used instead.
 
+    p0 : int, default 2
```
PABannier, line 27:
```suggestion
    p0 : int, default 10
```

skglm/solvers/group_bcd_solver.py
```diff
@@ -2,7 +2,7 @@
 from numba import njit
 
 
-def bcd_solver(X, y, datafit, penalty, w_init=None,
+def bcd_solver(X, y, datafit, penalty, w_init=None, p0=10,
                max_iter=1000, max_epochs=100, tol=1e-7, verbose=False):
```
PABannier, line 6: `tol=1e-7` or `tol=1e-6`?
Make it consistent with the docstring.

skglm/solvers/group_bcd_solver.py
```diff
@@ -56,51 +59,65 @@ def bcd_solver(X, y, datafit, penalty, w_init=None,
...
-                if stop_crit_in <= tol:
+                if stop_crit_in <= 0.3 * stop_crit:
                     print("Early exit")
                     break
-                prev_p_obj = current_p_obj
 
         current_p_obj = datafit.value(y, w, Xw) + penalty.value(w)
-        stop_crit = prev_p_obj - current_p_obj
+        grad = _construct_grad(X, y, w, Xw, datafit, all_groups)
+        opt = penalty.subdiff_distance(w, grad, all_groups)
```
PABannier, line 100: Why `all_groups` and not `ws`?

skglm/solvers/group_bcd_solver.py
```diff
@@ -56,51 +59,65 @@ def bcd_solver(X, y, datafit, penalty, w_init=None,
...
             break
 
-        prev_p_obj = current_p_obj
         p_objs_out[t] = current_p_obj
 
     return w, p_objs_out, stop_crit
 
 
 @njit
 def _bcd_epoch(X, y, w, Xw, datafit, penalty, ws):
```
PABannier, line 119: Decide whether you make `bcd_epoch` a public or private function.
If private, remove the docstring.
We have decided not to include docstrings for private functions.

skglm/solvers/group_bcd_solver.py
```diff
@@ -119,3 +136,21 @@ def _bcd_epoch(X, y, w, Xw, datafit, penalty, ws):
             if old_w_g[idx] != w[j]:
                 Xw += (w[j] - old_w_g[idx]) * X[:, j]
     return
+
+
+@njit
+def _construct_grad(X, y, w, Xw, datafit, ws):
```
PABannier, line 142: same here.
I'd change the `-gradients are stacked...` docstring into a comment though.

skglm/solvers/group_bcd_solver.py
Badr-MOUFAD: For the outer solver, we check optimality using all groups, right?
we consider ``ws`` only in the inner solver.

PABannier: you're right, my bad.

skglm/penalties/block_separable.py
```diff
@@ -218,10 +218,13 @@ def subdiff_distance(self, w, grad, ws):
         grp_ptr, grp_indices = self.grp_ptr, self.grp_indices
 
         scores = np.zeros(len(ws))
+        grad_ptr = 0
```
mathurinm, line 221: In this function, grad and w don't match: grad contains stacked gradients by group.
This is the only possible design, but it needs to be made more explicit:
- rename `grad` to `grad_ws`
- edit the docstring to reflect that gradient is restricted to ws 
- add a comment to explain that grad contains [grad_g1, grad_g2], etc.

mathurinm: In hindsight this should be clarified for separable penalties too

skglm/penalties/block_separable.py
```diff
@@ -218,10 +218,13 @@ def subdiff_distance(self, w, grad, ws):
...
         for idx, g in enumerate(ws):
-            grad_g = grad[idx]
-
             grp_g_indices = grp_indices[grp_ptr[g]: grp_ptr[g+1]]
+
+            grad_g = grad[grad_ptr: grad_ptr + len(grp_g_indices)]
+            grad_ptr += len(grp_g_indices)
+
             w_g = w[grp_g_indices]
             norm_w_g = norm(w_g)
```
mathurinm, line 229: no need to assign to a variable because it is used only once

skglm/penalties/block_separable.py
```diff
@@ -232,3 +235,24 @@ def subdiff_distance(self, w, grad, ws):
...
+        is_penalized = self.is_penalized(n_groups)
+
+        gsupp = np.zeros(n_groups, dtype=np.bool_)
+        for g in range(n_groups):
+            if not is_penalized[g]:
+                gsupp[g] = True
+                continue
+
+            grp_g_indices = grp_indices[grp_ptr[g]: grp_ptr[g+1]]
+            w_g = w[grp_g_indices]
```
mathurinm, line 254: no need to assign a variable being used only once

skglm/solvers/group_bcd_solver.py
```diff
@@ -2,8 +2,8 @@
 from numba import njit
 
 
-def bcd_solver(X, y, datafit, penalty, w_init=None,
-               max_iter=1000, max_epochs=100, tol=1e-7, verbose=False):
+def bcd_solver(X, y, datafit, penalty, w_init=None, p0=10,
+               max_iter=1000, max_epochs=100, tol=1e-6, verbose=False):
```
mathurinm, line 6: `cd_solver` uses 1e-4, let's harmonize

skglm/solvers/group_bcd_solver.py
```diff
@@ -56,51 +59,65 @@ def bcd_solver(X, y, datafit, penalty, w_init=None,
     datafit.initialize(X, y)
     all_groups = np.arange(n_groups)
     p_objs_out = np.zeros(max_iter)
+    stop_crit = 0.  # prevent ref before assign when max_iter == 0
```
mathurinm, line 62: can you add a check below that penalty is indeed a group penalty (checking it has attribute grp_indices for example) ?

skglm/solvers/group_bcd_solver.py
```diff
@@ -56,51 +59,65 @@ def bcd_solver(X, y, datafit, penalty, w_init=None,
...
     for t in range(max_iter):
-        if t == 0:  # avoid computing p_obj twice
-            prev_p_obj = datafit.value(y, w, Xw) + penalty.value(w)
+        if t == 0:  # avoid computing grad and opt twice
+            grad = _construct_grad(X, y, w, Xw, datafit, all_groups)
+            opt = penalty.subdiff_distance(w, grad, all_groups)
+            stop_crit = np.max(opt)
+
+            if stop_crit <= tol:
+                print("Outer solver: Early exit")
```
mathurinm, line 71: Harmonize with existing code:
```
        if verbose:
            print(f"Stopping criterion max violation: {stop_crit:.2e}")
```

skglm/solvers/group_bcd_solver.py
```diff
@@ -56,51 +59,65 @@ def bcd_solver(X, y, datafit, penalty, w_init=None,
...
 
             if epoch % 10 == 0:
-                current_p_obj = datafit.value(y, w, Xw) + penalty.value(w)
-                stop_crit_in = prev_p_obj - current_p_obj
+                grad_ws = _construct_grad(X, y, w, Xw, datafit, ws)
+                opt_in = penalty.subdiff_distance(w, grad_ws, ws)
+                stop_crit_in = np.max(opt_in)
 
                 if max(verbose - 1, 0):
+                    current_p_obj = datafit.value(y, w, Xw) + penalty.value(w)
```
mathurinm, line 88: is `p_obj` ok ?

skglm/solvers/group_bcd_solver.py
```diff
@@ -56,51 +59,65 @@ def bcd_solver(X, y, datafit, penalty, w_init=None,
...
                     break
-                prev_p_obj = current_p_obj
 
         current_p_obj = datafit.value(y, w, Xw) + penalty.value(w)
-        stop_crit = prev_p_obj - current_p_obj
+        grad = _construct_grad(X, y, w, Xw, datafit, all_groups)
+        opt = penalty.subdiff_distance(w, grad, all_groups)
+        stop_crit = np.max(opt)
 
         if max(verbose, 0):
```
mathurinm, line 103: if verbose

skglm/solvers/group_bcd_solver.py
```diff
@@ -119,3 +136,19 @@ def _bcd_epoch(X, y, w, Xw, datafit, penalty, ws):
             if old_w_g[idx] != w[j]:
                 Xw += (w[j] - old_w_g[idx]) * X[:, j]
     return
+
+
+@njit
+def _construct_grad(X, y, w, Xw, datafit, ws):
+    # compute the -gradient according to each group in ``ws``
+    # note: -gradients are stacked in a 1d array (e.g. [-grad_g1, -grad_g2, ...])
```
mathurinm, line 144: remove e.g.

skglm/solvers/group_bcd_solver.py
Badr-MOUFAD: You assume ``verbose`` positive?

skglm/penalties/block_separable.py
Badr-MOUFAD: ``norm_w_g`` is used twice (L235 and L238)

skglm/solvers/group_bcd_solver.py
Badr-MOUFAD: I implemented a ``_check_group_compatible`` function for that.

I will add also a unit test for that.

mathurinm: no need for a function now, keep it simple with a 2 lines check
@PABannier your thoughts ?

skglm/solvers/group_bcd_solver.py
mathurinm: Why, you see somebody passing verbose = - 1 ? It's a verbosity level

skglm/solvers/group_bcd_solver.py
```diff
@@ -56,51 +62,62 @@ def bcd_solver(X, y, datafit, penalty, w_init=None,
...
-        p_objs_out[t] = current_p_obj
+        p_objs_out[t] = p_obj
 
     return w, p_objs_out, stop_crit
 
 
 @njit
 def _bcd_epoch(X, y, w, Xw, datafit, penalty, ws):
-    """Perform a single BCD epoch on groups in ws."""
+    # perform a single BCD epoch on groups in ``ws``
```
mathurinm, line 120: rm double quotes

skglm/tests/test_group.py
```diff
@@ -26,6 +28,15 @@ def _generate_random_grp(n_groups, n_features, shuffle=True):
     return grp_indices, splits, groups
 
 
+def test_check_group_compatible():
+    l1_penalty = L1(1e-3)
```
mathurinm, line 32: compute alpha as a function of X and y

skglm/solvers/group_bcd_solver.py
Badr-MOUFAD: "never trust user input"

Anyways, I will change it

mathurinm: I'm more "garbage in, garbage out"
we don't check that verbose is an int for example, could be anything else

skglm/tests/test_group.py
mathurinm: I read too fast, you are right

skglm/solvers/group_bcd_solver.py
PABannier: I like the `_check_group_compatible` function, although I'd make it more explicit and put it in a utils file instead.
Also make the message error more explicit: `Penalties is not block-separable...`. `Missing attr...` is not explicit enough IMO.

I'd do something like this:
```
def _check_group_compatible(obj):
    group_attrs = set(['grp_ptr', 'grp_indices'])
    obj_attrs = set(dir(obj))
    if not group_attrs.issubset(obj_attrs):
            raise ValueError("Datafits and penalties must be compatible with group solvers")
```

PABannier: `_check_group_compatible` may be reused by other group solvers in the future, while `_bcd_epoch` and `_compute_grad` are tailored for this solver. That's why I advocate to encapsulate the logic in a function and put it in a `utils` module.

For the implementation, a for loop or set operations are both `O(n)` time in the number of attributes of the obj. In practice, penalties and datafits don't even have 100 attributes, so the effective cost remains relatively cheap.
I suggested this approach for conciseness, but I have no strong opinion on that matter.

skglm/penalties/block_separable.py
```diff
@@ -212,16 +212,23 @@ def prox_1group(self, value, stepsize, g):
         """Compute the proximal operator of group ``g``."""
         return BST(value, self.alpha * stepsize * self.weights[g])
 
-    def subdiff_distance(self, w, grad, ws):
-        """Compute distance of negative gradient to the subdifferential at ``w``."""
+    def subdiff_distance(self, w, grad_ws, ws):
+        """Compute distance to the subdifferential at ``w`` of negative gradient.
```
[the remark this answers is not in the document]
Badr-MOUFAD: I usually use the double quote to highlight that it's a function argument

usage-````

If it's not ok, I can revert this change.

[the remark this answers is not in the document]
PABannier: Looks good to me then. I saw @mathurinm tell you to remove them. @mathurinm any thoughts?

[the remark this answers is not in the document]
PABannier: Then all good for me. Thanks a lot @Badr-MOUFAD !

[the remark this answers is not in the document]
mathurinm: looks great this way!

mathurinm (approved):

mathurinm:
Thanks @Badr-MOUFAD :)

PR #29, opened 2022-06-08
[READY] ENH - add Anderson acceleration to ``group_bcd_solver``

Badr-MOUFAD:
This continues #28 and will proceed as follows

- [x] implement Anderson acceleration
- [x] benchmark against ``main`` branch


skglm/utils.py
```diff
@@ -254,3 +255,46 @@ def check_group_compatible(obj):
...
+            return
+
+        # compute residuals
+        U = np.zeros((n_features, K))
+        for j in range(K):
+            U[:, j] = arr_w[:, j+1] - arr_w[:, j]
+
+        # compte extrapolation coefs
+        ones_K = np.ones(K)
+        inv_UTU_ones = np.linalg.lstsq(U.T @ U, ones_K)[0]
```
Badr-MOUFAD, line 296: @mathurinm, @PABannier
Any idea on handling the case of a singular matrix here (without try: except)

I am not that convinced about using a least-square solution

PABannier: Check how we do it in `cd_solver`. Wrap the call to `np.linalg.inv` in a `try`- `except` block and catch `LinAlgError`.

PABannier: see: https://github.com/scikit-learn-contrib/skglm/blob/main/skglm/solvers/cd_solver.py#L300

Badr-MOUFAD:
I performed a benchmark (``celer`` vs. ``skglm-vanilla`` vs ``skglm-ws`` vs ``skglm-ws-acc``) on a toy dataset, thanks to [``benchopt``](https://github.com/benchopt/benchopt). 
Here are the results:

<details>
<summary>alpha / alpha_max = 0.1</summary>

rho-0 1

</details>

<details>
<summary>alpha / alpha_max = 0.01</summary>

rho-001

</details>

<details>
<summary>alpha / alpha_max = 0.001</summary>

rho-0001

</details>


I'm satisfied with the results. @mathurinm, @PABannier what is your thought about that?

skglm/utils.py
```diff
@@ -254,3 +255,51 @@ def check_group_compatible(obj):
...
+    ('K', int32),
+    ('n_features', int32),
+    ('current_iter', int32),
+    ('arr_w', float64[:, ::1]),
+]
+
+
+@jitclass(spec_AndersonAcceleration)
+class AndersonAcceleration:
+    """Anderson Accelerator."""
```
Badr-MOUFAD, line 270: If confirmed, this shall replace the acceleration procedure in other solvers.

PABannier: I don't think AA should be exposed this way. Also, see how we handle the singular matrix case. No need to catch `Exception("Singular matrix")`, you already have a `LinAlgError` raised by ` numpy` when the matrix is singular.
I wouldn't put AA in a utils file. If we keep this design, I'd put it in a `common` submodule.

 I agree that we could encapsulate the AA logic in some generic function / class to avoid redundancy though. @mathurinm  WDYT?

Badr-MOUFAD:
> @Badr-MOUFAD could you plot the sub-optimality curve? you should have the option when navigating on the HTML file generated by benchopt. Otherwise, we can't really see what's happening . Also make sure you're in semi-log scale.


I deliberately chose a visualization with objective value + linear scale, since the figures don't reach the same final value which (in my opinion) distorted the plots.

With that being said, here are the figures with sub-optimality + log-y scale, as requested


<details>
<summary>alpha / alpha_max = 0.1</summary>

01-sub

</details>

<details>
<summary>alpha / alpha_max = 0.01</summary>

001-sub

</details>

<details>
<summary>alpha / alpha_max = 0.001</summary>

0001-subpng

</details>

PABannier:
@Badr-MOUFAD there's something to understand then. My guess is that you're not solving the same optimization problem. Are you scaling `alpha` in the same fashion (by `n_samples`) ?
Otherwise, there's a problem in the optimization solver.
What happens at `alpha_max`? Do you have flat lines for all 4 solvers?

skglm/utils.py
mathurinm: I am also not a huge fan of having an AA class, but no strong opinion.

What's the benefit in your opinion Badr ?

mathurinm:
> I performed a benchmark (`celer` vs. `skglm-vanilla` vs `skglm-ws` vs `skglm-ws-acc`) on a toy dataset, thanks to [`benchopt`](https://github.com/benchopt/benchopt). Here are the results:
> alpha / alpha_max = 0.1
> 
> rho-0 1
> alpha / alpha_max = 0.01
> alpha / alpha_max = 0.001
> 
> I'm satisfied with the results. @mathurinm, @PABannier what is your thought about that?

The solvers don't converge to the same objective value, that's not normal for a convex problem.

skglm/utils.py
Badr-MOUFAD: AA accelerate algorithms without interfering with how they work, also it's the same for all of them.
Therefore, I think we better separate the logic behind it instead of re-inviting it for every use case.

I opted for a class instead of functions simply because AA holds an internal state (the past K+1 iterates). Also, I ensured that we benefit from jit compilation by implementing it as ``numba`` jit class.

With that being implemented, boosting an algorithm with AA with be as simple as 
1. initializing an instance of the class
2. calling its ``explotate`` method within the loop of the algorithm.

PABannier:
@Badr-MOUFAD Much nicer plots indeed. What do you mean by seeding random weights? In a convex setting, the solution does not depend on the initialization...
I'm not too surprised Celer remains faster than skglm since Numba is not as fast for block-wise operations as it is for coordinate-wise operations.
AA does not seem very useful in this setup. Could you try a larger one? Like 100 samples and 3000 features. Once the support is identified, AA should kick in and accelerate the descent.

Badr-MOUFAD:
> How do you set use_accel to False for celer ?

I altered the API of celer so that to expose the acceleration params. Here is a [useful link](https://github.com/mathurinm/celer/pull/267) to see the diff.

-----
As requested, I pushed the benchmark on a [separate branch](https://github.com/Badr-MOUFAD/skglm/tree/benchmark-group-lasso), for complete details about it.

Badr-MOUFAD:
@mathurinm, indeed there were two problems:

- in place update of ``w`` (it's like there was no acceleration at all)
- update of ``Xw`` (resulted in a divergent solution compared to celer)

I addressed these two problem in the last commit, and thereafter re performed the benchmarks

<details>
<summary>alpha / alpha_max = 1</summary>

fixed-1

</details>

<details>
<summary>alpha / alpha_max = 0.1</summary>

fixed-01

</details>


<details>
<summary>alpha / alpha_max = 0.01</summary>

fixed-001

</details>


<details>
<summary>alpha / alpha_max = 0.001</summary>

fixed-0001

</details>


and look how beautiful are theses benchmarks figures :heart_eyes:!

skglm/solvers/group_bcd_solver.py
```diff
@@ -99,19 +115,6 @@ def bcd_solver(X, y, datafit, penalty, w_init=None, p0=10,
...
-
-        if verbose:
-            print(
-                f"Iteration {t+1}: {p_obj:.10f}, "
-                f"stopping crit: {stop_crit:.2e}"
-            )
-
-        if stop_crit <= tol:
-            break
-
```
PABannier, line 114: I don't understand why you no longer check convergence at the end of one inner loop.

PABannier (commented):
Plots look good! Thanks @Badr-MOUFAD .
As for the AA class, I have no strong opinion on it.

skglm/solvers/group_bcd_solver.py
Badr-MOUFAD: I checked the implementation of ``cd_solver`` as well as ``skglm`` paper and I find out that we only do the check at the beginning.

PABannier (approved):

skglm/tests/test_group.py
```diff
@@ -123,5 +122,43 @@ def test_vs_celer_grouplasso(n_groups, n_features, shuffle):
...
+    n_iter = 0
+    w = np.ones(n_features)
+    for i in range(max_iter):
+        w = rho * w + 1
+
+        if norm(w - w_star, ord=np.inf) < 1e-9:
+            n_iter = i
+            break
+
+    assert n_iter_acc < n_iter  # AA should twice faster
```
mathurinm, line 160: assert breaks without displaying a useful message. Use np.testing.assert_array_less

skglm/utils.py
```diff
@@ -254,3 +254,55 @@ def check_group_compatible(obj):
...
+            inv_UTU_ones = np.linalg.solve(U.T @ U, ones_K)
+        except np.linalg.LinAlgError:
+            return
+        finally:
+            self.current_iter = 0
+
+        # extrapolate
+        C = inv_UTU_ones / (ones_K @ inv_UTU_ones)
+        w[:] = self.arr_w[:, 1:] @ C
+        Xw[:] = self.arr_Xw[:, 1:] @ C
```
mathurinm, line 308: add a comment saying that due to floating point errors, this may cause w and Xw to disagree

skglm/utils.py
```diff
@@ -254,3 +254,55 @@ def check_group_compatible(obj):
...
+    def extrapolate(self, w, Xw):
+        """Inplace update of ``w`` and ``Xw``."""
+        if self.current_iter <= self.K:
+            self.arr_w[:, self.current_iter] = w.copy()
+            self.arr_Xw[:, self.current_iter] = Xw.copy()
+            self.current_iter += 1
+            return
+
+        # compute residuals
+        U = self.arr_w[:, 1:] - self.arr_w[:, :-1]
```
mathurinm, line 294: use np.diff

skglm/utils.py
```diff
@@ -254,3 +254,55 @@ def check_group_compatible(obj):
...
+        self.n_samples, self.n_features = n_samples, n_features
+
+        self.current_iter = 0
+        self.arr_w = np.zeros(shape=(n_features, K+1))
+        self.arr_Xw = np.zeros(shape=(n_samples, K+1))
+
+    def extrapolate(self, w, Xw):
+        """Inplace update of ``w`` and ``Xw``."""
+        if self.current_iter <= self.K:
+            self.arr_w[:, self.current_iter] = w.copy()
```
mathurinm, line 288: no need for copy here nor in the line below

skglm/utils.py
```diff
@@ -254,3 +254,55 @@ def check_group_compatible(obj):
...
+    n_samples: int
+        Number of elements in ``Xw``.
+    """
+
+    def __init__(self, K, n_samples, n_features):
+        self.K = K
+        self.n_samples, self.n_features = n_samples, n_features
+
+        self.current_iter = 0
+        self.arr_w = np.zeros(shape=(n_features, K+1))
```
mathurinm, line 282: it's standard not to specify shape=

skglm/utils.py
```diff
@@ -254,3 +254,55 @@ def check_group_compatible(obj):
...
+        Number of previous iterates to consider for extrapolation.
+
+    n_features : int
+        Number of elements in ``w``.
+
+    n_samples: int
+        Number of elements in ``Xw``.
+    """
+
+    def __init__(self, K, n_samples, n_features):
```
mathurinm, line 277: n_samples and n_features should be inferred and stored at the first call to extrapolate, to make the API lighter

skglm/solvers/group_bcd_solver.py
```diff
@@ -1,10 +1,10 @@
 import numpy as np
 from numba import njit
 
-from skglm.utils import check_group_compatible
+from skglm.utils import AndersonAcceleration, check_group_compatible
 
 
-def bcd_solver(X, y, datafit, penalty, w_init=None, p0=10,
+def bcd_solver(X, y, datafit, penalty, w_init=None, p0=10, use_acc=True, K=5,
```
mathurinm, line 7: Last discussion: do we keep those exposed ? 
One of the claims of the paper is that K does not need to be tuned. Probably we always want to use AA. So we could make the API simpler by not exposing these 2 parameters.

Thoughts @QB3 @PABannier @Klopfe ?

skglm/solvers/group_bcd_solver.py
```diff
@@ -83,6 +96,9 @@ def bcd_solver(X, y, datafit, penalty, w_init=None, p0=10,
         for epoch in range(max_epochs):
             _bcd_epoch(X, y, w, Xw, datafit, penalty, ws)
 
+            if use_acc:  # inplace update of w and Xw
+                accelerator.extrapolate(w, Xw)
```
mathurinm, line 100: that feels nice with a class !

skglm/tests/test_group.py
```diff
@@ -123,5 +122,43 @@ def test_vs_celer_grouplasso(n_groups, n_features, shuffle):
...
+    n_iter = 0
+    w = np.ones(n_features)
+    for i in range(max_iter):
+        w = rho * w + 1
+
+        if norm(w - w_star, ord=np.inf) < 1e-9:
+            n_iter = i
+            break
+
+    np.testing.assert_array_less(n_iter_acc, n_iter)  # AA should twice faster
```
mathurinm, line 160: why twice ?
in practice what do you get for n_iter_acc and n_iter?

skglm/utils.py
```diff
@@ -254,3 +254,54 @@ def check_group_compatible(obj):
...
+
+    Extrapolate the asymptotic VAR ``w`` and ``Xw``
+    based on ``K`` previous iterations.
+
+    Parameters
+    ----------
+    K : int
+        Number of previous iterates to consider for extrapolation.
+
+    Note:
```
mathurinm, line 270: rm this note

skglm/utils.py
```diff
@@ -254,3 +254,54 @@ def check_group_compatible(obj):
...
+            inv_UTU_ones = np.linalg.solve(U.T @ U, self.ones_K)
+        except np.linalg.LinAlgError:
+            return
+        finally:
+            self.current_iter = 0
+
+        # extrapolate
+        C = inv_UTU_ones / np.sum(inv_UTU_ones)
+        # floating point errors may cause w and Xw to disagree
+        w[:] = self.arr_w_[:, 1:] @ C
```
mathurinm, line 306: This updates `w` in place. This means, if extrapolation goes awry, you don't keep the nice behavior of the base sequence `w`. I think it is dangerous.

skglm/tests/test_group.py
Badr-MOUFAD: 13 for AA and 99 for the classical approach. (I meant twice in log scale)

I cannot remember where I found such a result for VAR systems.

skglm/utils.py
```diff
@@ -254,3 +254,54 @@ def check_group_compatible(obj):
...
+    Note:
+    -----
+    the shape of ``w`` and ``Xw`` is inferred in
+    the first call of ``extrapolate``.
+    """
+
+    def __init__(self, K):
+        self.K, self.current_iter = K, 0
+        self.arr_w_, self.arr_Xw_ = None, None
+        self.ones_K = np.ones(K)
```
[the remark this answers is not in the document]
Badr-MOUFAD: Small details matter. IMO, let's define it only once.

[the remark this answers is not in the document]
Badr-MOUFAD: Disagree with you totally, the var name is meaningful in this context. even if it's 0.01% it remains a code and performance opt.

I have no strong opinion. I will declare it implicitly.

skglm/solvers/group_bcd_solver.py
PABannier: agreed.

Klopfe: I agree too

skglm/tests/test_group.py
```diff
@@ -123,5 +122,32 @@ def test_vs_celer_grouplasso(n_groups, n_features, shuffle):
...
+        w = rho * w + 1
+        Xw = X @ w
+
+        if norm(w - w_star, ord=np.inf) < tol:
+            n_iter_acc = i
+            break
+
+    np.testing.assert_allclose(w, w_star)
+    np.testing.assert_allclose(Xw, X @ w_star)
+    np.testing.assert_array_equal(n_iter_acc, 13)
```
mathurinm, line 149: I'm sorry, maybe I was not clear here: I agree with you that we should test that AA performs way better than regular iteration. I prefer testing equality because it checks that there is a stringent difference (13 vs 99), that is also made clear to the reader (while 98 vs 99 would still pass the n_iter < n_iter_acc test).

So we also need to test that n_iter (regular) is equal to 99. Otherwise, checking convergence of AA in 13 iterations here does not check that it's faster than regular iterations.

mathurinm (approved):

mathurinm:
Thanks @Badr-MOUFAD, it looks really clean with a class now and the perfs are impressive

Klopfe:
> Thanks @Badr-MOUFAD, it looks really clean with a class now and the perfs are impressive

yes, thank you @Badr-MOUFAD, amazing work

PABannier:
Thanks a lot @Badr-MOUFAD ! Really great work!

PR #32, opened 2022-06-13
ENH add prox newton for logistic datafit

PABannier:
Prox-Newton offers better performance than CD + AA + WS for the Logistic datafit.
The goal of this PR is to implement Prox-Newton for the Logistic datafit.

So far, I've implemented a prox-Newton solver and I've done a bit refactoring.
I've moved a `sigmoid` function into a utils module.
Besides, some helper functions for the cd solver like `construct_grad` have been moved to a `common` submodule.

EDIT by QB3:
- link toward the `celer` PR for prox-Newton investigation https://github.com/mathurinm/celer/pull/275
- link toward the `benchmark_logregL1` for numerical comparisons https://github.com/PABannier/benchmark_logreg_l1/pull/1


PABannier:
Sounds good indeed! I'm finishing debugging the solver for the Logistic datafit. There remains a few bugs, then we can write in a generic fashion to support other datafits :)

PABannier:
quick update: i've fixed a bunch of mistakes. Now, the solver converges fast, but still not faster than CD...
I suspect the backtrack line search is not optimized speed-wise, with a bunch of heavy computations.
I'll add AA when all the potential mistakes are cleaned out.

PABannier:
Screen Shot 2022-06-21 at 6 19 27 PM

alpha = alpha_max / 20

Prox-Newton is way faster for problems with less than 10_000 features.
It works just fine now.

Badr-MOUFAD:
That looks nice indeed!

Can you pls check the effectiveness of Anderson Acceleration for this use case? This might hint at proving that more formally for a generic prox newton.
Also, can you implement unitest to check the solution against ``scikit-learn``?

skglm/datafits/single_task.py
```diff
@@ -88,13 +88,6 @@ def full_grad_sparse(
 Quadratic, Quadratic_32 = jit_factory(Quadratic, spec_quadratic)
 
 
-@njit
-def sigmoid(x):
-    """Vectorwise sigmoid."""
```
Badr-MOUFAD, line 93: pls revert

skglm/solvers/cd_solver.py
```diff
@@ -347,118 +347,6 @@ def cd_solver(
     return w, np.array(obj_out), stop_crit
 
 
-@njit
-def dist_fix_point(w, grad, datafit, penalty, ws):
-    """Compute the violation of the fixed point iterate scheme.
```
Badr-MOUFAD, line 352: also, revert these

skglm/solvers/prox_newton_solver.py
```diff
@@ -0,0 +1,312 @@
...
+        resolution of problem restricted to features in working set).
+
+    max_epochs : int, optional
+        Maximum number of prox-Newton epochs on each subproblem.
+
+    max_backtrack : int, optinal
+        Maximum number of backtracking steps for the line search.
+
+    min_pn_cd_epochs : int, optional
+        Minimum number of iterations used in the prox-Newton coordinate
```
Badr-MOUFAD, line 49: why ``min_pn_cd_epochs``? I see that it's used only in L264 L290 as an additional condition to break a loop. ``tol`` is sufficient as a break condition?

skglm/solvers/prox_newton_solver.py
```diff
@@ -0,0 +1,312 @@
...
+            w_acc, Xw_acc = accelerator.extrapolate(w, Xw)
+            p_obj = datafit.value(y, w, Xw) + penalty.value(w)
+            p_obj_acc = datafit.value(y, w_acc, Xw_acc) + penalty.value(w_acc)
+
+            if p_obj_acc < p_obj:
+                w[:] = w_acc
+                Xw[:] = Xw_acc
+
+            if epoch % 5 == 0:  # check every 5 epochs, PN epochs are expensive
+                p_obj = datafit.value(y, w[ws], Xw) + penalty.value(w)
```
Badr-MOUFAD, line 150: No need to recompute p_obj here as it's already done in L142

skglm/solvers/prox_newton_solver.py
```diff
@@ -0,0 +1,312 @@
...
+    grad_datafit = np.zeros(n_samples)  # gradient of F(Xw)
+
+    lc = np.zeros(ws_size)  # weighted Lipschitz constants
+    bias = np.zeros(ws_size)
+
+    for i in range(n_samples):
+        prob = 1. / (1. + np.exp(y[i] * Xw[i]))
+        # this supposes that the datafit is scaled by n_samples
+        hessian_diag[i] = prob * (1. - prob) / n_samples
+        grad_datafit[i] = -y[i] * prob / n_samples
```
Badr-MOUFAD, line 194: Why do you hard code the ``grad`` here? Use ``gradient_scalar`` of the ``datafit`` instead

skglm/tests/test_prox_newton.py
```diff
@@ -0,0 +1,73 @@
...
+
+from skglm.penalties import L1, L1_plus_L2
+from skglm.datafits import Logistic
+from skglm.solvers.cd_solver import cd_solver
+from skglm.solvers.prox_newton_solver import prox_newton_solver
+
+from skglm.utils import make_correlated_data
+
+
+n_samples, n_features = 30, 50
```
Badr-MOUFAD, line 16: IMO: Define for every unitest its own params instead of defining them at the beginning.
Indeed,
- we will change the params for every unit test and thereby cover more cases
- when a unitest fails, the code will act as a standalone debug script

skglm/utils.py
```diff
@@ -288,13 +331,18 @@ def extrapolate(self, w, Xw):
...
+            # to inv_UTU_ones.sum() being null.
+            if inv_UtU_ones.sum() == 0:
+                raise np.linalg.LinAlgError
         except np.linalg.LinAlgError:
             return w, Xw
         finally:
             self.current_iter = 0
 
         # extrapolate
-        C = inv_UTU_ones / np.sum(inv_UTU_ones)
```
Badr-MOUFAD, line 298: revert pls

Badr-MOUFAD (commented):
I did a first pass on the PR. It looks good but still needs some refinement.

skglm/solvers/prox_newton_solver.py
PABannier: I'm not sure it computes the same thing. `gradient_scalar` is the grad of the datafit w.r.t. to `w`, while I'm computing the gradient w.r.t. `Xw`. I haven't checked the computations. 

Anyway, I think it's efficient to do it this way since we reuse `prob` to compute `hessian_diag` and `grad_datafit`. WDYT?

skglm/tests/test_prox_newton.py
PABannier: It depends on what we want to test, and how we see the test script evolves. 
If we plan in later PRs to support other datafits, I suppose we are more interested in parametrizing the tests by the datafits/estimators rather than by the scale of the test setup. I adopted the same design as in https://github.com/scikit-learn-contrib/skglm/blob/main/skglm/tests/test_estimators.py . 

No strong opinion on the matter though. @mathurinm WDYT?

skglm/solvers/cd_solver.py
Badr-MOUFAD: I 100% agree with you, but let's do it in another PR.
IMO it makes the diff terribly and hard to follow.

skglm/solvers/prox_newton_solver.py
Badr-MOUFAD: This is how I view the grad considering the datafit: $F(Xw)$ where $F(u) = \frac{1}{n} \sum_i f_i(u_i)$

$\nabla (F(Xw)) = X^T \\nabla F(Xw)$

such that $\\nabla F(u) = (f_i'(u_i))_{1 \leq i \leq n}$

Badr-MOUFAD: IMO, computing the hessian, weighted dot, ... should be handled by the datafit itself and the solver shouldn't interfere/intervene with it at all.

I believe that a PN solver is a just ``cd_solver`` with a different descent direction/step.

PABannier: When calling `gradient_scalar` with a loop over the features, you obtain the gradient of the datafit w.r.t. to the features. Here I'm looping over the samples to obtain the gradient of `F(Xw)`. It's not the same quantity.

As for the design, I agree, but let's keep this PR simple and refactor to handle other datafits in a later PR.

skglm/solvers/prox_newton_solver.py
mathurinm: It's a heuristic used in blitz 
https://github.com/tbjohns/BlitzL1/blob/master/src/solver.cpp#L24
it seemed to matter when I did benchmarks

skglm/solvers/prox_newton_solver.py
```diff
@@ -0,0 +1,312 @@
...
+        Value of stopping criterion at convergence.
+    """
+    if ws_strategy not in ("subdiff", "fixpoint"):
+        raise ValueError(f"Unsupported value for ws_strategy: {ws_strategy}")
+    if not isinstance(datafit, (Logistic, Logistic_32)):
+        raise ValueError("Prox-Newton solver only supports Logistic datafits.")
+    n_features = X.shape[1]
+    if n_features >= 10_000:
+        warnings.warn(
+            "Prox-Newton solver can be prohibitively slow for high-dimensional " +
```
mathurinm, line 86: not sure this warning should be raised : it is coupled with a working set method so the number of features is not really relevant here

skglm/solvers/prox_newton_solver.py
```diff
@@ -0,0 +1,312 @@
...
+
+    is_sparse = sparse.issparse(X)
+    for t in range(max_iter):
+        if is_sparse:
+            grad = datafit.full_grad_sparse(
+                X.data, X.indptr, X.indices, y, Xw)
+        else:
+            grad = construct_grad(X, y, w, Xw, datafit, all_feats)
+
+        if ws_strategy == "subdiff":
```
mathurinm, line 104: I'd use a single ws_strategy for starters. My 2 c.

skglm/solvers/prox_newton_solver.py
```diff
@@ -0,0 +1,312 @@
...
+            if is_sparse:
+                _prox_newton_iter_sparse(
+                    X.data, X.indptr, X.indices, Xw, w, y, penalty, ws,
+                    min_pn_cd_epochs, _max_pn_cd_epochs, max_backtrack, pn_tol)
+            else:
+                _prox_newton_iter(
+                    X, Xw, w, y, penalty, ws, min_pn_cd_epochs, _max_pn_cd_epochs,
+                    max_backtrack, pn_tol)
+
+            # 3) do Anderson acceleration on smaller problem
```
mathurinm, line 140: You can remove this part, in my experience it's not useful to do AA for prox newton

skglm/solvers/prox_newton_solver.py
mathurinm: @Badr-MOUFAD  PN requires quantities that are different from the ones that `cd_solver` needs, and the optimization algorithm is quite different.

skglm/solvers/prox_newton_solver.py
```diff
@@ -0,0 +1,312 @@
...
+    bias = np.zeros(ws_size)
+
+    for i in range(n_samples):
+        prob = 1. / (1. + np.exp(y[i] * Xw[i]))
+        # this supposes that the datafit is scaled by n_samples
+        hessian_diag[i] = prob * (1. - prob) / n_samples
+        grad_datafit[i] = -y[i] * prob / n_samples
+
+    for idx, j in enumerate(ws):
+        lc[idx] = weighted_dot(X, Xw, hessian_diag, j, ignore_b=True)
```
mathurinm, line 197: this is copied from celer but you can do the dot directly.

skglm/solvers/prox_newton_solver.py
```diff
@@ -0,0 +1,312 @@
...
+        X, w, ws, hessian_diag, bias, lc, penalty, min_cd_epochs, max_cd_epochs, tol)
+    step_size = _backtrack_line_search(
+        w, Xw, delta_w, X_delta_w, ws, y, penalty, max_backtrack)
+
+    for idx, j in enumerate(ws):
+        w[j] += step_size * delta_w[idx]
+    Xw += step_size * X_delta_w
+
+
+@njit
```
mathurinm, line 210: As a first version I'd support only dense X to split the work in easier pieces @Badr-MOUFAD @PABannier

skglm/utils.py
```diff
@@ -80,6 +80,49 @@ def prox_2_3(x, u):
...
+@njit
+def weighted_dot(X, b, weights, j, ignore_b=False):
+    """Weighted dot product between X[:, j] and b."""
+    res = 0.
+    if ignore_b:
+        # if ignore_b, weighted dot product between X[:, j] and X[:, j]
+        for i in range(X.shape[0]):
+            res += (X[i, j] ** 2) * weights[i]
+    else:
+        for i in range(X.shape[0]):
```
mathurinm, line 99: you can dot the dot * and .sum no ? and remove this function. I used it in Cython

skglm/tests/test_prox_newton.py
```diff
@@ -0,0 +1,72 @@
...
+tol = 1e-10
+l1_ratio = 0.3
+
+dict_penalties = {}
+
+dict_penalties["L1"] = L1(alpha=alpha)
+dict_penalties["L1_plus_L2"] = L1_plus_L2(alpha=alpha, l1_ratio=l1_ratio)
+
+
+@pytest.mark.parametrize("X", [X])
```
Badr-MOUFAD, line 34: better define ``X`` inside the unitest.

skglm/tests/test_prox_newton.py
```diff
@@ -0,0 +1,72 @@
...
+def test_prox_newton_vs_sklearn(X):
+    datafit = Logistic()
+    if issparse(X):
```
Badr-MOUFAD, line 37: remove the case of sparse

skglm/tests/test_prox_newton.py
```diff
@@ -0,0 +1,72 @@
...
+
+    estimator_sk = LogReg_sklearn(
+        C=1/(alpha * n_samples), fit_intercept=False, tol=tol, penalty='l1',
+        solver='liblinear')
+    estimator_sk.fit(X, y_ind)
+
+    np.testing.assert_allclose(w_newton, np.ravel(estimator_sk.coef_), atol=1e-5)
+
+
+@pytest.mark.parametrize("X", [X])
```
Badr-MOUFAD, line 54: same here for ``X``

skglm/tests/test_prox_newton.py
```diff
@@ -0,0 +1,72 @@
...
+@pytest.mark.parametrize("penalty_name", ["L1", "L1_plus_L2"])
+def test_prox_newton_vs_cd(X, penalty_name):
+    datafit = Logistic()
+    if issparse(X):
```
Badr-MOUFAD, line 58: same for the sparse case

skglm/solvers/prox_newton_solver.py
```diff
@@ -0,0 +1,221 @@
...
+from scipy import sparse
+from numba import njit
+from skglm.datafits import Logistic, Logistic_32
+from skglm.solvers.common import construct_grad
+
+
+def prox_newton_solver(
+    X, y, datafit, penalty, w, Xw, max_iter=50, max_epochs=1000, max_backtrack=10,
+    min_pn_cd_epochs=2, max_pn_cd_epochs=10, p0=10, tol=1e-4, verbose=0
+):
```
Badr-MOUFAD, line 11: cosmetics, bring this to L10

Badr-MOUFAD (commented):
Code looks much better. I think we are almost there.

GJ @PABannier!

skglm/solvers/prox_newton_solver.py
```diff
@@ -0,0 +1,221 @@
...
+        w, Xw, delta_w, X_delta_w, ws, y, penalty, max_backtrack)
+
+    for idx, j in enumerate(ws):
+        w[j] += step_size * delta_w[idx]
+    Xw += step_size * X_delta_w
+
+
+@njit
+def _newton_cd(
+    X, w, feats, hessian_diag, bias, lc, penalty, min_epochs, max_epochs, tol
```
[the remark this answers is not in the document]
PABannier: i've left `feats` to stay consistent with `_cd_epoch` in `cd_solver`.

skglm/solvers/prox_newton_solver.py
PABannier: flake8 raise a `E125 continuation line` warning

Badr-MOUFAD: A tab must be added, 
suggestion for the whole function signature

```suggestion
def prox_newton_solver(
        X, y, datafit, penalty, w, Xw, max_iter=50, max_epochs=1000, max_backtrack=10,
        min_pn_cd_epochs=2, max_pn_cd_epochs=10, p0=10, tol=1e-4, verbose=0):
```

skglm/solvers/prox_newton_solver.py
```diff
@@ -0,0 +1,228 @@
...
+                if stop_crit_in <= tol_in:
+                    if max(verbose - 1, 0):
+                        print("Early exit")
+                    break
+        obj_out.append(p_obj)
+    return w, np.array(obj_out), stop_crit
+
+
+@njit
+def _prox_newton_iter(
```
Badr-MOUFAD, line 149:
```suggestion
def _prox_newton_iter(
        X, Xw, w, y, penalty, ws, min_cd_epochs, max_cd_epochs, max_backtrack, tol):
```

skglm/solvers/prox_newton_solver.py
```diff
@@ -0,0 +1,228 @@
...
+    step_size = _backtrack_line_search(
+        w, Xw, delta_w, X_delta_w, ws, y, penalty, max_backtrack)
+
+    for idx, j in enumerate(ws):
+        w[j] += step_size * delta_w[idx]
+    Xw += step_size * X_delta_w
+
+
+@njit
+def _newton_cd(
```
Badr-MOUFAD, line 181:
```suggestion
def _newton_cd(
        X, w, feats, hessian_diag, bias, lc, penalty, min_epochs, max_epochs, tol):
```

skglm/tests/test_prox_newton.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+n_samples, n_features = 30, 50
+X, y, _ = make_correlated_data(
+    n_samples=n_samples, n_features=n_features, density=0.1, random_state=0)
+y_ind = np.sign(y)
+
+np.random.seed(0)
+
+alpha_max = norm(X.T @ y, ord=np.inf) / n_samples
```
Badr-MOUFAD, line 22:
```suggestion
alpha_max = norm(X.T @ y, ord=np.inf) / (2 * n_samples)
```

``alpha_max`` for logreg is not the same as quad datafit.
in general, it has the following formula  $\\Omega^D(X^T \\nabla F(0))$
(cf. [Gap Safe screening rules for sparsity enforcing penalties](https://arxiv.org/abs/1611.05780) proposition 4)

In our case $\\nabla F(0) = \\frac{1}{2n} y$

Can you add unitest to check that we get a zero solution when we are at ``alpha_max``

Badr-MOUFAD (commented):
With these changes taken into account, LGTM!

skglm/solvers/prox_newton_solver.py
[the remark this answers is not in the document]
Badr-MOUFAD: would be nice if we switch to ``ws`` instead?
IMO, it's less confusing.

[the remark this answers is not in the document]
PABannier: agreed. let's do that in the next PR and change as well in `_cd_epoch`.

skglm/solvers/prox_newton_solver.py
```diff
@@ -0,0 +1,225 @@
...
+        raise ValueError("Prox-Newton solver only supports Logistic datafits.")
+    n_features = X.shape[1]
+    pen = penalty.is_penalized(n_features)
+    unpen = ~pen
+    n_unpen = unpen.sum()
+    obj_out = []
+    all_feats = np.arange(n_features)
+    stop_crit = np.inf  # initialize for case n_iter=0
+
+    accelerator = AndersonAcceleration(K=5)
```
mathurinm, line 82: Is there a benefit in using AndersonAccelerator with PN ? In my experience, no

skglm/solvers/prox_newton_solver.py
```diff
@@ -0,0 +1,225 @@
...
+                delta -= penalty.alpha * abs(delta_w[idx])
+        theta = -y * sigmoid(-y * (Xw + step_size * X_delta_w))
+        delta += X_delta_w @ theta
+        if delta < 1e-7:
+            break
+        step_size = step_size / 2
+    return step_size
+
+
+@njit
```
mathurinm, line 221: this can imported form `single_task.py`

mathurinm:
How does it compare against celer's PN and other solvers ? Can you share a benchopt benchmark @PABannier or @Badr-MOUFAD ?

skglm/solvers/prox_newton_solver.py
PABannier: Yes, I removed it and it was really slow afterwards (https://github.com/scikit-learn-contrib/skglm/pull/32/commits/18b185fbd6b75194e3da18dbd4657c64b203639f).

mathurinm: This is weird. PN iterations break the VAR structure in my experience, see Fig 8 https://www.jmlr.org/papers/volume21/19-587/19-587.pdf 

How does the current solver compare to Blitz and Celer-PN ?

QB3:
Checkpoint on where we are:
Screenshot from 2022-07-14 18-32-21

- here is the benchmark_logregl1 PR to compare skglm with other solvers https://github.com/PABannier/benchmark_logreg_l1/pulls
- here a quick benchmark with the AA deactivated for our PN solver

QB3:
(Without AA) we are currently very slow compared to `sklearn`, `blitz` or `celer`.
@PABannier suspects a line-search problem

PABannier:
The solver does converge eventually, but very slowly. So the step size would be my first guess. Besides, I implemented the backtrack algo from @mathurinm 's thesis. Backtracking from Celer is quite different, but I couldn't understand it fully.

skglm/solvers/prox_newton_solver.py
```diff
@@ -0,0 +1,219 @@
...
+def _newton_cd(
+        X, w, ws, hessian_diag, bias, lc, penalty, min_epochs, max_epochs, tol):
+    delta_w, X_delta_w = np.zeros(len(ws)), np.zeros(X.shape[0])
+    for epoch in range(max_epochs):
+        sum_sq_hess_diff = 0
+        for idx, j in enumerate(ws):
+            stepsize = 1/lc[idx] if lc[idx] != 0 else 1000
```
QB3, line 186: Do you have an intuition on the hardcoded value `1000` for the stepsize here?

PABannier: same as here https://github.com/scikit-learn-contrib/skglm/blob/main/skglm/solvers/cd_solver.py#L380

QB3:
> The solver does converge eventually, but very slowly. So the step size would be my first guess. Besides, I implemented the backtrack algo from @mathurinm 's thesis. Backtracking from Celer is quite different, but I couldn't understand it fully.

Maybe this is the first thing to disantangle then, we could try to:
- 1 make sure that this is a slow convergence problem (i.e. profiling to check that the gradient evaluation steps are the most costly)
- 2 if this is indeed a slow convergence pb, we could try to understand and convert together `celer` line-search code

mathurinm:
Yes I would copy the LS code from celer

mathurinm:
> Checkpoint on where we are: Screenshot from 2022-07-14 18-32-21
> 
>     * here is the benchmark_logregl1 PR to compare skglm with other solvers https://github.com/PABannier/benchmark_logreg_l1/pulls
> 
>     * here a quick benchmark with the AA deactivated for our PN solver

what is the legend here?

skglm/solvers/prox_newton_solver.py
```diff
@@ -0,0 +1,269 @@
...
+                aux[i] = 1. - 1. / (1. + exp_Xw[i])
+        # compute deriv
+        deriv_l1 = 0.
+        for j in ws:
+            w_j = w[j] + step_size * delta_w[j]
+            if w_j == 0.:
+                deriv_l1 -= abs(delta_w[j])
+            else:
+                deriv_l1 += w_j / abs(w_j) * delta_w[j]
+        deriv_loss = X_delta_w @ aux
```
QB3, line 256: As spotted by @PABannier we might a normalization here

QB3:
Checkpoint:
- perf of `skglm` are similar to `celer-pn`
- however coefficients are not the same along the iterations (even with ws desactivated + tol inner smaller for `celer-pn)`

mathurinm:
the stopping criterion are not the same for the inner problems

QB3:
> the stopping criterion are not the same for the inner problems

Yes I also changed the inner stopping criterion from `celer-pn,` through ` eps_inner = 0.0000001` (https://github.com/mathurinm/celer/pull/275/commits/54b5361cb1ad6373654022c1d00bb925ae110728).

Despite this change, iterates are not the same

mathurinm:
Closing in favor of #51

PR #33, opened 2022-06-21
[READY] MAINT - use ``AndersonAcceleration`` class in ``cd_solver``

Badr-MOUFAD:
This aims to refactor ``cd_solver``. It will proceed as follows:

- [x] AA class in ``cd_solver``
- [x] hide ``use_acc``, ``K`` from the API
- [x] perform benchmark against main
  [(link to the benchmark results)](https://badr-moufad.github.io/share-benchmarks/cd_solver/benchmark_aa_single_benchopt_run_2022-07-25_23h48m09.html)


skglm/solvers/multitask_bcd_solver.py
```diff
@@ -254,32 +253,19 @@ def bcd_solver(
...
-                            W[:] = W_acc
-                            XW[:] = Xw_acc
-                    except np.linalg.LinAlgError:
-                        if max(verbose - 1, 0):
-                            print("----------Linalg error")
+                W_acc, XW_acc = accelerator.extrapolate(W.ravel(), XW.ravel())
+
+                p_obj = datafit.value(Y, W, XW) + penalty.value(W)
+                p_obj_acc = (datafit.value(Y, W_acc.reshape(-1, n_tasks),
+                                           XW_acc.reshape(-1, n_tasks))
```
Badr-MOUFAD, line 262: @mathurinm, I m not sure whether I pass in a view or a copy.

Link to [reshape Numpy doc ](https://numpy.org/doc/stable/reference/generated/numpy.reshape.html#:~:text=Returns,the%20returned%20array.)

Badr-MOUFAD:
Here are [the results of the benchmark](https://badr-moufad.github.io/share-benchmarks/) to ensure that switching to AA class won't deteriorate the performance of the solver.

I'm satisfied with the results for the ``cd_solver``, yet not that convinced with those of the multitask solver. I suspect that this https://github.com/scikit-learn-contrib/skglm/pull/33#discussion_r905250846 is the reason behind but still quite unsure.

PABannier:
i think there's an issue with the curves @Badr-MOUFAD . All the suboptimality gaps are going up (Lasso and MTL).

Badr-MOUFAD:
@PABannier, that was a nice catch indeed! I was blinded by the shape of the figures.
There was a bug in the computation of the objective. In fact, I use solver path in the benchmarks, which outputs an array of the coefs along the path and not a single coef.

I fixed that and here are [the new figures](https://badr-moufad.github.io/share-benchmarks/).

PABannier:
LGTM! For the performance on the MTL, remember that Numba is not as performing as in the single task case. MTL is not very fast actually.

skglm/solvers/cd_solver.py
```diff
@@ -282,44 +272,16 @@ def cd_solver(
...
-                        p_obj = datafit.value(y, w, Xw) + penalty.value(w)
-                        # p_obj = datafit.value(y, w, Xw) +penalty.value(w[ws])
-                        Xw_acc = X[:, ws] @ w_acc[ws]
-                        # TODO : managed penalty.value(w[ws])
-                        p_obj_acc = datafit.value(
-                            y, w_acc, Xw_acc) + penalty.value(w_acc)
-                        if p_obj_acc < p_obj:
-                            w[:] = w_acc
-                            Xw[:] = Xw_acc
+            w_acc, Xw_acc = accelerator.extrapolate(w, Xw)
```
mathurinm, line 275: This design is not optimal: K - 1 times out of K, we just return w and Xw, so there is no need to compute p_obj_acc. A third variable, boolean, should be returned, and checking of Xw_acc done only if acceleration is performed

mathurinm (changes requested):

QB3:
Just to let you know, in my experience computing the quantity `Xwacc`  using the `K` last `Xw` (as it is done here  https://github.com/scikit-learn-contrib/skglm/blob/9b7f3ef89c5261002ace2d804245a6d934023cb5/skglm/utils.py#L300) often yielded numerical errors (as you noticed) but also slower convergence. I usually had to recompute `Xwacc` using `X @ wacc`

Badr-MOUFAD:
I pulled main to integrate changes that address https://github.com/scikit-learn-contrib/skglm/pull/33#discussion_r919887164
Here are the results of the [benchmark](https://badr-moufad.github.io/share-benchmarks/multitask_solver/benchmark_aa_multi_benchopt_run_2022-07-18_14h09m20.html) on datasets with sizes (100, 1000) and (100, 3000)

> ... often yielded numerical errors (as you noticed) ...

@QB3 can you pls hint me at those numerical errors?


With that being said, I see two areas for improving MTL

- [ ] Extend AA class to handle the case of ``W`` matrix.
  so far, we get around that by flattening ``W`` which results in extra-memory copies of ``W``
  (c.f. [L248-249](https://github.com/Badr-MOUFAD/skglm/blob/9b27d8ab60b950ab70f06d445a79a3fd7e08ec28/skglm/solvers/multitask_bcd_solver.py#L248-L249)  and [L253-253](https://github.com/Badr-MOUFAD/skglm/blob/9b27d8ab60b950ab70f06d445a79a3fd7e08ec28/skglm/solvers/multitask_bcd_solver.py#L252-L253))

- [ ] Atler AA to accelerate only the features in ``ws``.
This is relevant in the case of a solution with small support  

--------------

If these seem right, I shall tackle them in a separate PR

skglm/solvers/cd_solver.py
```diff
@@ -283,44 +273,18 @@ def cd_solver(
...
+            if is_extrapolated:  # avoid computing p_obj for un-extrapolated w, Xw
+                p_obj = datafit.value(y, w, Xw) + penalty.value(w)
+                p_obj_acc = datafit.value(y, w_acc, Xw_acc) + penalty.value(w_acc)
+
+                if p_obj_acc < p_obj:
+                    w[:], Xw[:] = w_acc, Xw_acc
+                    p_obj = p_obj_acc
 
             if epoch % 10 == 0:
                 # TODO : manage penalty.value(w, ws) for weighted Lasso
```
mathurinm, line 287: can you move the todo to keep it next to the line of code it applies to (the p_obj computation)

skglm/solvers/multitask_bcd_solver.py
```diff
@@ -307,6 +285,7 @@ def bcd_solver(
                         if max(verbose - 1, 0):
                             print("Early exit")
                         break
+        p_obj = datafit.value(Y, W[ws, :], XW) + penalty.value(W)
```
mathurinm, line 288: here W is restricted to ws, above it's not. Is one better than the other ?

Badr-MOUFAD:
For conciseness, this PR will focus on ``_cd_solver`` only.
MLT deserves its own PR.

skglm/solvers/cd_solver.py
```diff
@@ -264,10 +258,6 @@ def cd_solver(
         ws = np.argpartition(opt, -ws_size)[-ws_size:]
         # is equivalent to ws = np.argsort(opt)[-ws_size:]
 
-        if use_acc:
-            last_K_w = np.zeros([K + 1, ws_size])
```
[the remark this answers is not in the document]
Badr-MOUFAD: solved by passing w[ws] to AA instead of the entire vector

Badr-MOUFAD:
Here are the [result of the benchmark](https://badr-moufad.github.io/share-benchmarks/cd_solver/benchmark_aa_single_benchopt_run_2022-07-25_13h19m55.html) for the cd_solver.
@mathurinm, I will perform a benchmark on the finance dataset to answer https://github.com/scikit-learn-contrib/skglm/pull/33#discussion_r928651847

skglm/solvers/cd_solver.py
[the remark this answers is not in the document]
Badr-MOUFAD: @mathurinm,
link to [benchmark on finance dataset](https://badr-moufad.github.io/share-benchmarks/cd_solver_finance/benchmark_aa_single_benchopt_run_2022-07-25_13h43m34.html)

PABannier:
LGTM @Badr-MOUFAD !

QB3:
Thanks a lot for the PR @Badr-MOUFAD !

Implementation with the class seems to yield faster results, do you think it is because `X_w_acc` is not recomputed anymore when extrapolating?

Screenshot from 2022-07-25 10-16-41

QB3:
> Here are the [result of the benchmark](https://badr-moufad.github.io/share-benchmarks/cd_solver_simulated/benchmark_aa_single_benchopt_run_2022-07-25_13h19m55.html) for the cd_solver.

just to be 100% sure that there is no regression with respect to main, can you make the same benchmark on rcv1, new20, finance, with lambda_max / 10, lambda_max / 100, lambda_max / 1000

Badr-MOUFAD:
@QB3, as requested, here is the [full benchmark](https://badr-moufad.github.io/share-benchmarks/cd_solver/benchmark_aa_single_benchopt_run_2022-07-25_23h48m09.html) on finance, rvc1, news20, and simulated data.

> Implementation with the class seems to yield faster results, do you think it is because X_w_acc is not recomputed anymore when extrapolating?

Frankly, I haven't profiled the AA code. But recomputing ``Xw_acc`` is certainly a reason among (potentially) others.

QB3:
Thanks you so much for the hard work and the benchmark @Badr-MOUFAD .

One last (collective) question: do we want to remove the parameters for the activation / deactivation of Anderson acceleration? as well as the number of iterations? (I have no strong takes, I just want to be sure everyone is OK with it).
Apart from this LGTM!

Badr-MOUFAD:
> One last (collective) question: do we want to remove the parameters for the activation / deactivation of Anderson acceleration? as well as the number of iterations? (I have no strong takes, I just want to be sure everyone is OK with it).

actually, AA is used either way as it's one of the features of ``skglm`` solvers. We've previously had a [small discussion](https://github.com/scikit-learn-contrib/skglm/pull/29#discussion_r901695736) about that.

One relevant use case of exposing these parameters is when debugging/comparing solvers. Yet, we can still manage that in a separate branch and keep the API simple. we had also [a discussion](https://github.com/mathurinm/celer/pull/267) on that with @mathurinm.

PABannier:
I think the API should be kept as minimal as possible.

QB3:
Should we merge?

mathurinm:
The benchmark results look good, conflict need to be fixed and then we can do a final round of reviews

mathurinm (approved):

PR #34, opened 2022-06-21
[WIP] ENH - Handle low rank residual matrix in ``AndersonAcceleration``

Badr-MOUFAD:
This aims to handle the case of singular residual matrix in ``AndersonAccelration`` class

According to [Dual Extrapolation for Sparse Generalized Linear Models](https://arxiv.org/abs/1907.05830) (cf. Proposition 7),
if the matrix of residuals is not full rank (`U.T @ U` singular) then we can extrapolate using the K-1 previous iteration.

The following proposes an implementation of this. It will proceed as follow

- [x] handle the case of singular `U.T @ U`
- [x] implement unitest
- [x] perform benchmark against main

--------------

P.S. @mathurinm does this count as a citation :smile:

### DIFF
--- a/debug_script.py
+++ b/debug_script.py
@@ -1,0 +1,48 @@
+r"""Example with:
+
+u_{n+1} = \begin{cases}
+            u_{n}            & if n \bmod 3
+            \rho u_{n} + b   & otherwise
+          \end{cases}
+"""
+import numpy as np
+from numpy.linalg import norm
+
+from debug_script_utils import AA_singular, AA_usual
+
+
+def assess_AA(aa_class):
+    max_iter, tol = 1000, 1e-9
+    n_features = 5
+    X = np.eye(n_features)
+
+    b = 1
+    np.random.seed(0)
+    rho = np.random.rand(n_features)
+    w_star = b / (1 - rho)
+
+    accelerator = aa_class(K=5)
+    w = np.ones(n_features)
+    Xw = X @ w
+    for i in range(max_iter):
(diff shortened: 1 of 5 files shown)


mathurinm:
Thanks ! I think #33 should be done before, while AA is still relatively simple ?

Badr-MOUFAD:
Here are the result of the benchmark

<details>
<summary>alpha / alpha_max = 1</summary>

sin-1


</details>

<details>
<summary>alpha / alpha_max = 0.1</summary>

sin-2


</details>


<details>
<summary>alpha / alpha_max = 0.01</summary>

sin-3


</details>


<details>
<summary>alpha / alpha_max = 0.001</summary>

sin-4


</details>


I'm not that convinced with the result. I believe that handling the singularity case will pay off only if the opt problem at hand yields a lot of low-rank residual matrices.

I made an example that shows that handling the singularity is not as beneficial as we might think:

$$
u_{n+1} = \\begin{cases}
            u_{n}            & \\text{ if } n \\bmod 3 \\\\
            \\rho \\ u_{n} + b   & \\text{ otherwise }
         \\end{cases}
$$

In this particular case, handling the singularity case deteriorate the convergence speed (more iterations are required than a usual AA). Refer to ``debug_script`` for more details.

I'm not really in favor of complexifying the code for a minor detail.
What do you suggest @mathurinm? I believe in your foresight.

mathurinm:
> I'm not really in favor of complexifying the code for a minor detail.

Completely agree. Let's close this then, thanks !

PR #39, opened 2022-07-12
[WIP] Investigate low running time of MultiTask Lasso 

QB3:
In some cases our MultiTask Lasso has some slow cases compared to sklearn.
The goal of this PR is to investigate some suspicious low running time for the MultiTask Lasso. 

With @PABannier:
- [x]  We added an example to compare and profile the MultiTask Lasso
- [x] We manage to remove the numba warning for the MTL  
- [x] We found out that applying the prox was the most costly operation for 1 epoch, up to 50 % of the time of 1 epoch. The prox operation was more costly than accessing the gradients. As in MNE, using in place operation for the update of the vector `W[j, :]`, we managed to decrease the computing time of the proximal operator computation.

Now, the (rank one) update of the residuals seems to be one of the main bottleneck of the `_bcd_epoch` function. We cannot use dgemm rountines as in MNE because numba does not support them.

### DIFF
--- a/examples/compare_time_mtl.py
+++ b/examples/compare_time_mtl.py
@@ -1,0 +1,95 @@
+# Authors: Quentin Bertrand
+#          Pierre-Antoine Bannier
+"""
+=============================================
+Timing comparison with scikit-learn for MultiTask Lasso
+=============================================
+Compare time to solve medium scale MultiTask Lasso problems with scikit-learn.
+"""
+
+
+import time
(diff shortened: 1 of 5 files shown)


mathurinm:
do we want to keep this open or it'll be better to start from scratch given the conflicts @QB3 @PABannier ?

QB3:
This is pretty old, I think it will be better to start from scratch

PR #41, opened 2022-07-14
ENH - Avoid computing ``p_obj`` for un accelerated ``w``, ``Xw``

### LINKED ISSUE #40: ENH AndersonAccelerator.accelerate() does not create new points most of the time
This means computation of `p_obj_acc` is not needed. It's a big bottleneck in datasets such as Finance where the subproblems are very small compared to `n_features`.

Returning an additional boolean variable as the output of `accelerate` would allow us to skip the `p_obj_acc` computation in cases where it's not needed

Badr-MOUFAD:
this closes #40 and proceeds as follows:

- [x] Alter ``AndersonAcceleration`` to return ``is_extrapolated``
- [x] handle case ``is_extrapolated`` to avoid extra ``p_obj`` computation
- [ ] perform benchmark against main to ensure no performance loss

### DIFF
--- a/skglm/solvers/group_bcd_solver.py
+++ b/skglm/solvers/group_bcd_solver.py
@@ -91,13 +91,15 @@ def bcd_solver(X, y, datafit, penalty, w_init=None, p0=10,
             # inplace update of w and Xw
             _bcd_epoch(X, y, w, Xw, datafit, penalty, ws)
 
-            w_acc, Xw_acc = accelerator.extrapolate(w, Xw)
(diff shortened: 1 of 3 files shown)


mathurinm (approved):
Go for me if it looks OK compared to main !

Badr-MOUFAD:
@mathurinm, @QB3, Here are the [results of the benchmark](https://badr-moufad.github.io/share-benchmarks/grp_cd_solver/benchmark_enh_aa_benchopt_run_2022-07-14_13h00m40.html) against the main branch.

PR #42, opened 2022-07-14
NOMRG Evaluating space allocation in BCD solver

PABannier:
In the wake of #39 , this PR profiles the memory allocated by the multi-task BCD solver. It appears that arrays are copied when the prox is evaluated, which makes the procedure very slow when `n_samples` grow.

This PR consists in quantifying the gain in speed due to modifying arrays in place.


mathurinm:
same as #39: do we want to keep this open or it'll be better to start from scratch given the conflicts @PABannier ?

PR #44, opened 2022-07-22
API design with penalty and datafit jitted inside fit

mathurinm:
This attempts to simplify the API of datafits by jit complying classes when fitting models.

Currently, we have a distinct jit class for each type (e.g. ``Quadratic``, ``Quadratic_32``)


test_design.py
```diff
@@ -20,6 +20,22 @@
 from skglm.solvers import cd_solver_path, bcd_solver_path
 
 
+def jit_closure():
+    compiled_classes = dict()
```
PABannier, line 24: I assume you can't use `lru_cache` from `functools` because spec is unhashable, right?

debug_script.py
```diff
@@ -0,0 +1,16 @@
+import numpy as np
```
Badr-MOUFAD, line 1: I appreciate your idea of making a free variable where to put jit compiled classes. but still, we need to hard code the spec of classes which then leaves us with some boilerplate code.

``debug_script`` shows that datafit breaks down when its spec doesn't match the types of ``X`` and ``y`` (the inputs).
IMO, we should opt for a solution (still haven't thought about one yet) where it's up to the class to manage types-conflicts

Hence the final look should be something like

```python
datafit = Quadratic(dtype, **params)
```

mathurinm: The specs of the classes are already hardcoded: https://github.com/scikit-learn-contrib/skglm/blob/main/skglm/penalties/separable.py#L11

The spec can be updated dynamically using the trick we have to generate Quadratic32. The script is just an example to show that it can work.
I don't think it is necessary to pass dtype when instanciating `datafit`. It can be inferred from X at the time of `fit`.

debug_script.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+X, y, _ = make_correlated_data()
+
+X = X.astype(np.float32)
+y = y.astype(np.float32)
+
+print(X.dtype, y.dtype)
+
+# This will throw an error
```
mathurinm, line 13: This makes little sense for me. I am using different classes in `test_design`, that are not jitted.

It's easy to modify `our_jit_compiler` to handle `X_dtype` 

Then we implement `compile_and_clone` which takes an instance, compiles its class with `our_jit_compiler`, and copies parameters from the original instance

test_design.py
mathurinm: I forgot about lru_cache ! Did you test it here ? We can make `spec` a tuple if this helps

debug_script.py
mathurinm: I don't really understand this example: it does not use the new code I propose. The same exact code would fail on main, right ?

Maybe I'm not clearly exposing the design I want to try. I will code the rest later today and share at that point.

test_design.py
PABannier: I tried it yesterday but didn't push because of the unhashable issue. If specs can be a tuple, it's more Pythonic :) I'm having a look

PABannier:
I pushed a streamlined version of `our_jit_compiler` that uses built-in LRU cache.
The performance remains the same:
Screen Shot 2022-07-23 at 11 00 45 AM

Minor change to make it work: change all specs to tuples instead of lists.

debug_script.py
Badr-MOUFAD: The example is to illustrate what problem this PR could solve

mathurinm: Let's do things one at a time. First, avoid having jitted classes as attributes, in order to be usable in GridSearchCV. Then, as you suggest it's a good idea to update the spec dynamically depending on the type of X being passed to fit.

mathurinm:
@PABannier thanks!

I have merged my design proposal into the existing code. It seems to be working OK.

For now I did it only for L1 and Quadratic. It should be straightforward to extend to other datafits and penalties.
It removes then need for the `*_32` classes, and the need to hardcode them into generalized estimator.

Something needs to be fixed to handle `path` correctly (there is duplicated code currently that I could not fix easily)

WDYT of the design @QB3 @Klopfe ?

PABannier:
Looks very clean to me. The kind of PR that removes a lot of hurdles at once.
You mentioned an issue with `path` @mathurinm , what is it?

mathurinm:
I was surprised at first that fit does not call `self.path` (there is duplicated code in these functions). But path is designed for coherence with the Lasso (I think) and does not use penalty nor datafit.

I think in the end we can/should just call `self.path` in fit without defining the jitted penalty and datafit there, just define them in the path. That should be OK. @Badr-MOUFAD if you read this :)

skglm/datafits/base.py
```diff
@@ -31,6 +32,23 @@ def spec_to_float32(spec):
     return spec32
 
 
+@lru_cache
```
Badr-MOUFAD, line 35: This throws an error for python <= 3.7. 

```shell
TypeError: Expected maxsize to be an integer or None
```
It was corrected starting from 3.8

```suggestion
@lru_cache()
```

skglm/estimators.py
```diff
@@ -153,8 +156,11 @@ def path(self, X, y, alphas, coef_init=None, return_n_iter=True, **params):
             The number of iterations along the path. If return_n_iter is set to `True`.
         """
         path_func = cd_solver_path if y.ndim == 1 else bcd_solver_path
+        penalty = compiled_clone(self.penalty)
+        datafit = compiled_clone(self.datafit, to_float32=X.dtype is np.float32)
```
Badr-MOUFAD, line 160: This doesn't test for equality

```python
X = np.ones(5, dtype=np.float32)
X.dtype is np.float32  # False
```
```suggestion
        datafit = compiled_clone(self.datafit, to_float32=X.dtype == np.float32)
```

Hence, ``test_design.py`` threw an error when I added the case of float32

skglm/penalties/separable.py
```diff
@@ -8,18 +8,21 @@
...
     def __init__(self, alpha):
         self.alpha = alpha
 
+    def get_spec(self):
+        spec_L1 = (
+            ('alpha', float64),
+        )
+        return spec_L1
+
+    def params_to_dict(self):
```
Badr-MOUFAD, line 23: is this method to align with scikit learn API?

Badr-MOUFAD (commented):
I liked the design. Big appreciation @mathurinm!

Before we proceed forward, we need enhance ``spec_to_float32`` non float attribute (e.g. spec of [`QuadraticGroup` ](https://github.com/scikit-learn-contrib/skglm/blob/8a581b01c4a81b317b778b728c85f2adf545c90a/skglm/datafits/group.py#L9-L13) ). Indeed, in this case, it returns a wrong spec (cf. debug_script).

skglm/datafits/single_task.py
```diff
@@ -85,7 +95,7 @@ def full_grad_sparse(
         return grad
 
 
-Quadratic, Quadratic_32 = jit_factory(Quadratic, spec_quadratic)
+# Quadratic, Quadratic_32 = jit_factory(Quadratic, spec_quadratic)
```
mathurinm, line 98:
```suggestion
```

skglm/estimators.py
```diff
@@ -234,11 +232,15 @@ def fit(self, X, y):
         if not self.warm_start or not hasattr(self, "coef_"):
             self.coef_ = None
 
-        path_func = cd_solver_path if y.ndim == 1 else bcd_solver_path
         X_ = yXT if isinstance(self.datafit, QuadraticSVC) else X
 
+        path_func = cd_solver_path if y.ndim == 1 else bcd_solver_path
+        penalty = compiled_clone(self.penalty)
+        datafit = compiled_clone(self.datafit, to_float32=X.dtype is np.float32)
```
mathurinm, line 239:
```suggestion
        datafit = compiled_clone(self.datafit, to_float32=X.dtype == np.float32)
```

skglm/penalties/separable.py
mathurinm: we need it because we get the instance, compile its class (not the instance itself), then need to reinstantiate one instance of the compiled class with the same parameters as the original object

skglm/tests/test_estimators.py
```diff
@@ -191,8 +192,8 @@ def assert_deep_dict_equal(expected_attr, estimator):
                 np.testing.assert_allclose(v, v_est)
             else:
                 assert v == v_est
-    df_reg, df_clf = Quadratic(), Logistic()
-    pen_reg, pen_clf = L1(4.), MCPenalty(2., 3.)
+    df_reg, df_clf = compiled_clone(Quadratic()), compiled_clone(Logistic())
+    pen_reg, pen_clf = compiled_clone(L1(4.)), compiled_clone(MCPenalty(2., 3.))
```
mathurinm, line 196: It's regular instances which should be passed rather than compiled ones, no?

Badr-MOUFAD: They are not compiled hence don't have a Lipschitz attribute.

mathurinm: Compilation is not related to having a lipschitz attribute.

The whole purpose of this PR is to pass non jitted classes to GeneralizedLinearEstimator. If you must compile them before, it means something is weird, right?

skglm/tests/test_estimators.py
```diff
@@ -230,6 +231,7 @@ def test_grid_search(estimator_name):
 
 
 if __name__ == '__main__':
-    Datafit, Penalty, Estimator = Logistic, L1, SparseLogisticRegression
-    pen_args = [alpha]
-    test_generic_estimator(Datafit, Penalty, True, Estimator, pen_args)
+    # Datafit, Penalty, Estimator = Logistic, L1, SparseLogisticRegression
```
mathurinm: to clean

test_design.py
```diff
@@ -0,0 +1,31 @@
+import numpy as np
```
mathurinm: rm this file if done

skglm/datafits/base.py
```diff
@@ -24,34 +25,63 @@ def spec_to_float32(spec):
...
+@lru_cache()
+def jit_cached_compile(klass, spec, to_float32=False):
+    """Jit compile class and cache compilation.
 
     Parameters
     ----------
-    Datafit : datafit class, inheriting from BaseDatafit
-        A datafit class, to be compiled.
+    klass : class
+        Un instantiated Datafit or Penalty.
```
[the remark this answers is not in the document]
Badr-MOUFAD: This is to indicate that ``Datafit``, resp ``Penalty``, is passed and not ``Datafit()``, resp ``Penalty()``.

skglm/tests/test_estimators.py
mathurinm: Ok, so this is a non standard behavior. Imagine the person reading this piece of code in 2 years: usually we don't need to compile, but here we do. It's not super obvious why. So we should make their life easier by adding a comment, cf my proposition above

skglm/estimators.py
```diff
@@ -234,11 +231,11 @@ def fit(self, X, y):
         if not self.warm_start or not hasattr(self, "coef_"):
             self.coef_ = None
 
-        path_func = cd_solver_path if y.ndim == 1 else bcd_solver_path
         X_ = yXT if isinstance(self.datafit, QuadraticSVC) else X
 
-        _, coefs, kkt = path_func(
-            X_, y, self.datafit, self.penalty, alphas=[self.penalty.alpha],
+        # TODO merge with self.path, handle penalty not an argument in self.path
```
mathurinm, line 236:
```suggestion
```

mathurinm (commented):
Last thing to do is to add `get_spec` and `params_to_dict` to BaseDatafit and BasePenalty, to add docstring explaining what these methods do. then LGTM!

skglm/utils.py
```diff
@@ -1,10 +1,93 @@
...
+            else:
+                dtype32 = dtype
+        else:
+            raise ValueError(f"Unknown spec type {dtype}")
+        spec32.append((name, dtype32))
+    return spec32
+
+
+@lru_cache()
+def jit_cached_compile(klass, spec, to_float32=False):
```
QB3, line 43: I am not sure ti understand why we need a separated `jit_cached_compile` function, why not directly doing this in the `compile_clone` function? (I am sure there is a good reason, just asking)

skglm/utils.py
```diff
@@ -1,10 +1,93 @@
+from functools import lru_cache
```
QB3, line 1: Nitpick, but I would tend to put these functions into a separated `utils_jit.py` or `utils_numba.py`

skglm/utils.py
mathurinm: because we use lru_cache on this function, but not on the cloning one (that gets fed different potentially different instances every time), while this one receives classes

skglm/utils.py
mathurinm: this file indeed deserves a refactoring, but I'd do it later: there are math utils, the Anderson class, etc that should be put in separate files

skglm/utils.py
QB3: I see, so the key point is the caching of with respect to `klass` and `spec` instead of `instance`

mathurinm: yes exactly

Badr-MOUFAD (approved):

PR #49, opened 2022-08-06
NOMRG - Prox Newton Solver à la ``skglm``

Badr-MOUFAD:
A modular prox newton solver that uses ``sklgm`` stopping criteria.

This proceeds as follows
- [x] implement PN solver for dense and sparse datasets
- [x] Apply and benchmark PN solver for LogReg
- Apply PN solver for other problems
    - [ ] Poisson regression
    - [ ] ...

--------
[Link to the benchmark results](https://badr-moufad.github.io/share-benchmarks/pn_sparse/benchmark_logreg_benchopt_run_2022-08-05_15h03m05.html) on ``news20.binary``, ``rcv1.binary``, and simulated datasets.

Refer to ``benchmark_logreg`` to reproduce. To do so:

- ``pip install -r requirements.txt``
- run ``benchmark_entry.py`` file.


mathurinm:
Thanks @Badr-MOUFAD, results are a great improvement over celer. 
It seems as expected, AA does not bring anything in this solver, so we can remove it ?
I would not add other datafits (Poisson) in this PR, but in a subsequent one.

One last thing to investigate is that we seem to be slower than blitz in some settings (though we improve a lot over celer's implementation):
image
Do you have any idea of where the difference comes from?

benchmark_entry.py
```diff
@@ -0,0 +1,9 @@
+from benchopt.benchmark import Benchmark
```
mathurinm, line 1: all of these files make the diff harder to parse and the PR harder to review. You can have a separate benchmark PR in your fork of benchopt_logreg_l1

skglm/prototype_PN/L1_penalty.py
```diff
@@ -0,0 +1,68 @@
+import numpy as np
```
mathurinm, line 1: you can directly edit the `L1` class. It will better highlight which changes are needed, and reduce the amount of duplicated us for reviewers to read

skglm/prototype_PN/acceleration.py
```diff
@@ -0,0 +1,52 @@
...
+
+spec_AA = (
+    ('K', int32),
+    ('current_iter', int32),
+    ('arr_w_', float64[:, ::1]),
+    ('arr_Xw_', float64[:, ::1])
+)
+
+
+@jitclass(spec_AA)
```
mathurinm, line 13: Is it clear that this class needs to be jitted ?

Badr-MOUFAD: Yes, it must be jitted as it is used inside a numba jitted function.
That being said, AA doesn't bring much to the solver so we won't need it.

Badr-MOUFAD:
There is no need to keep this PR alive since, so far, we have a PN solver in https://github.com/scikit-learn-contrib/skglm/pull/51 that significantly improves this PR's solver and particularly performs better than the state-of-the-art, ``blitz`` and ``scikit-learn:liblinear`` to name a few.

PR #50, opened 2022-08-11
NOMERG - ``py_blitz`` a python numba replica of Blitz

Badr-MOUFAD:
This provides a python/numba re-implementation of Blitz.

It's an (exact) replica of Blitz in the sense that it:
- screens features and builds working sets as described and implemented in Blitz 
- uses the same stopping criteria for the three loops (iterations, epochs, and CD iterations)
- implements similar code optimizations to avoid redundant computations 

For one iteration, both solvers, ``py_blitz`` and Blitz, have the same behavior (same objective, gap, ..., for every epoch and CD iteration). Yet, for several iterations, their behavior may differ. After inspecting that, this mismatch is due to sorting algorithms failing to produce the same arg sort for an array with some of its elements equal (e.g. ``[0, 0, 5, 0, 1, 3, 2]``). This results in having working sets with different features order. After inforcing a specific order, the two solvers behave similarly.

### Experiment on simulated[100, 200], rho=0.1

<details>
<summary>Without enforcing and order</summary>

- ``py_blitz``

py_blitz_ordinar

- Blitz

blitz-ordinar

</details>

<details>
<summary>Enforcing and order</summary>

- ``py_blitz``

py_blitz_enforce

-  Blitz

blitz_enforcing

</details>

----------------------------
checklist:
- [x] Implement a python numba replica of Blitz
- [x] ensure py_blitz and Blitz behave similarly
- [x] perform benchmark to compare both solvers

 
[Link to](https://badr-moufad.github.io/share-benchmarks/pn_solver_all/benchmark_logreg_benchopt_run_2022-08-14_10h57m32.html) all benchmark results (sparse case included)

PR #51, opened 2022-08-15
[READY] ENH - Add modular Prox Newton solver

Badr-MOUFAD:
Following the discussion in [#49](https://github.com/scikit-learn-contrib/skglm/pull/49#issuecomment-1208340139) and the investigation on #50,  this adds an improved version of the suggested PN solver in #49 and applies it to Logistic Regression.

Other problems will be handled in upcoming PRs.

The improvements are code optimization that reduces redundant computations, namely the computation of ``grads`` and the update of ``Xw``. 

Here are the [results of the benchmarks](https://badr-moufad.github.io/share-benchmarks/pn_a_la_skgml_fastmath/benchmark_logreg_l1_benchopt_run_2022-08-20_20h12m26.html), which, in addition, show the performance of the current solver ``PN-skglm-improved`` compared to the previous one #49 ``PN-skglm``.

[Repo](https://github.com/Badr-MOUFAD/benchmark_logreg_l1/tree/PN-improved) to reproduce/try other benchmarks

--------------

**P.S**: the presence of two solvers ``pn_solver``, and ``pn_solver_improved`` is just for benchmarking purposes.


skglm/datafits/single_task.py
```diff
@@ -126,6 +126,11 @@ def get_spec(self):
     def params_to_dict(self):
         return dict()
 
+    def raw_hessian(self, y, Xw):
+        """Compute hessian of datafit w.r.t ``Xw``."""
+        exp_yXw = np.exp(-y * Xw)
+        return exp_yXw / (1 + exp_yXw) ** 2 / len(y)
```
agramfort, line 132: numba does not have https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.expit.html#scipy-special-expit ?

Badr-MOUFAD: Yes indeed. But unfortunately, numba is not compatible with ``scipy`` and thereby throws an error at JIT compilation time.

skglm/datafits/single_task.py
```diff
@@ -126,6 +126,15 @@ def get_spec(self):
     def params_to_dict(self):
         return dict()
 
+    def raw_grad(self, y, Xw):
+        """Compute gradient of datafit w.r.t ``Xw``."""
+        return -y / (1 + np.exp(y * Xw)) / len(y)
+
+    def raw_hessian(self, y, Xw):
+        """Compute hessian of datafit w.r.t ``Xw``."""
```
mathurinm, line 134: cosmit: Hessian, and it's the diagonal of the Hessian

skglm/penalties/separable.py
```diff
@@ -33,14 +33,17 @@ def prox_1d(self, value, stepsize, j):
     def subdiff_distance(self, w, grad, ws):
         """Compute distance of negative gradient to the subdifferential at w."""
         subdiff_dist = np.zeros_like(grad)
+        is_w_ws_provided = len(w) == len(ws)
```
mathurinm, line 36: what about calling this function with `ws=np.arange(n_features)` as we currently do in other parts of the code?

skglm/solvers/pn_solver.py
```diff
@@ -0,0 +1,295 @@
+import numpy as np
```
mathurinm, line 1: I would go for a more explicit filename: `prox_newton.py`. The `_solver` can be omitted IMO (we are in the `solvers` folder).

skglm/solvers/pn_solver.py
```diff
@@ -0,0 +1,295 @@
...
+from scipy.sparse import issparse
+from numba import njit
```
mathurinm, line 3: really cosmetic: inside each import block we put shorter import first otherwise it summons @tommoral

skglm/solvers/pn_solver.py
```diff
@@ -0,0 +1,295 @@
...
+
+from skglm.solvers.common import construct_grad, construct_grad_sparse
+
+
+def pn_solver(X, y, datafit, penalty, w_init=None, p0=10,
+              max_iter=20, max_epochs=1000, tol=1e-4, verbose=0):
+    """Run a Prox Newton solver.
```
mathurinm, line 10: maybe add refs to Blitz/celer ? and add that it's combined with working sets (to explain `p0`)

skglm/solvers/pn_solver.py
```diff
@@ -0,0 +1,295 @@
...
+
+    Parameters
+    ----------
+    X : array, shape (n_samples, n_features)
```
mathurinm, line 14:
```suggestion
    X : array or sparse CSC matrix, shape (n_samples, n_features)
```

skglm/solvers/pn_solver.py
```diff
@@ -0,0 +1,295 @@
...
+
+    penalty : instance of BasePenalty
+        Penalty object.
+
+    w_init : array, shape (n_features,), default None
+        Initial value of coefficients.
+        If set to None, a zero vector is used instead.
+
+    p0 : int, default 10
+        Minimum number of groups to be included in the working set.
```
mathurinm, line 31:
```suggestion
        Minimum number of features to be included in the working set.
```

skglm/solvers/pn_solver.py
```diff
@@ -0,0 +1,295 @@
...
+
+    max_iter : int, default 20
+        Maximum number of iterations.
```
mathurinm, line 34: this can be a bit mysterious to the user as there are many loops in the algorithm

skglm/solvers/pn_solver.py
```diff
@@ -0,0 +1,295 @@
...
+    Returns
+    -------
+    w : array, shape (n_features,)
+        Solution that minimizes the problem defined by datafit and penalty.
+
+    objs_out: array (max_iter,)
+        The objective values at every outer iteration.
+
+    stop_crit: float
+        The value of the stop criterion.
```
mathurinm, line 54:
```suggestion
        The value of the stopping criterion when the solver stops.
```

skglm/solvers/pn_solver.py
```diff
@@ -0,0 +1,295 @@
...
+    verbose : bool, default False
+        Amount of verbosity. 0/False is silent.
+
+    Returns
+    -------
+    w : array, shape (n_features,)
+        Solution that minimizes the problem defined by datafit and penalty.
+
+    objs_out: array (max_iter,)
+        The objective values at every outer iteration.
```
mathurinm, line 51: see remark above on `iterations`

skglm/solvers/pn_solver.py
```diff
@@ -0,0 +1,295 @@
...
+    n_samples, n_features = X.shape
+    w = np.zeros(n_features) if w_init is None else w_init
+    Xw = np.zeros(n_samples) if w_init is None else X @ w_init
+    all_features = np.arange(n_features)
+    stop_crit = 0.
+    p_objs_out = []
+
+    is_sparse = issparse(X)
+    if is_sparse:
+        X_bundles = (X.data, X.indptr, X.indices)
```
mathurinm, line 65: this is a clean design, if it does not slow anything down, we could adopt it in other parts of the code in another PR

skglm/solvers/pn_solver.py
```diff
@@ -0,0 +1,295 @@
...
+        # similar to np.argsort()[-ws_size:] but without sorting
+        ws = np.argpartition(opt, -ws_size)[-ws_size:]
+        grad_ws = grad[ws]
+
+        for epoch in range(max_epochs):
+            tol_in = 0.3 * stop_crit
+
+            # find descent direction
+            if is_sparse:
+                (
```
mathurinm, line 103: cosmetic: `delta_w_ws, X_delta_w_ws = _compute_descent_direction(`  and break line there, right after the opening parenthesis

skglm/solvers/pn_solver.py
```diff
@@ -0,0 +1,295 @@
...
+                    print("|—— Inner loop early exit")
+                break
+
+        p_obj = datafit.value(y, w, Xw) + penalty.value(w)
+        p_objs_out.append(p_obj)
+    return w, p_objs_out, stop_crit
+
+
+@njit
+def _compute_descent_direction(X, y, w_epoch, Xw_epoch, grad_ws, datafit, penalty,
```
mathurinm, line 149: cosmetic suggestion to be shorter: `_descent_direction`

skglm/solvers/pn_solver.py
```diff
@@ -0,0 +1,295 @@
...
+                               ws, max_cd_iter, tol):
+    # Given:
+    #   - b = \nabla   F(X w_epoch)
```
mathurinm, line 152: nitpitck: the dash could be mistaken for a minus. Maybe use 1) and 2) ?

skglm/solvers/pn_solver.py
```diff
@@ -0,0 +1,295 @@
...
+    #   - D = \nabla^2 F(X w_epoch)   <------>  raw_hess
+    # Minimize for delta_w = w - w_epoch:
```
mathurinm, line 154: minimize *quadratic approximation* ?

skglm/solvers/pn_solver.py
mathurinm: also too many spaces between nabla and F I believe

skglm/solvers/pn_solver.py
```diff
@@ -0,0 +1,295 @@
...
+    for idx, j in enumerate(ws):
+        lipschitz[idx] = raw_hess @ X[:, j] ** 2
+
+    cached_grads = np.zeros(len(ws))
+    X_delta_w_ws = np.zeros(X.shape[0])
+    w_ws = w_epoch[ws]
+
+    for cd_iter in range(max_cd_iter):
+        for idx, j in enumerate(ws):
+
```
mathurinm, line 169: rm blank line?

skglm/solvers/pn_solver.py
```diff
@@ -0,0 +1,295 @@
...
+            if lipschitz[idx] == 0:
+                continue
+
+            cached_grads[idx] = grad_ws[idx] + X[:, j] @ (raw_hess * X_delta_w_ws)
+            old_w_idx = w_ws[idx]
+            stepsize = 1 / lipschitz[idx]
+
+            w_ws[idx] = penalty.prox_1d(
+                old_w_idx - stepsize * cached_grads[idx],
+                stepsize, j
```
mathurinm, line 180: can it fit on the previous line ?

skglm/solvers/pn_solver.py
```diff
@@ -0,0 +1,295 @@
...
+        if cd_iter % 5 == 0:
+            opt = penalty.subdiff_distance(w_ws, cached_grads, ws)
+            if np.max(opt) <= tol:
+                break
+
+    # descent direction
+    return w_ws - w_epoch[ws], X_delta_w_ws
+
+
+# sparse version of func above
```
mathurinm, line 195: sparse version of  `_compute_descent_direction` (otherwise, if we reorganize the file the order may change and this comment becomes invalid)

skglm/solvers/pn_solver.py
```diff
@@ -0,0 +1,295 @@
...
+    return w_ws - w_epoch[ws], X_delta_w_ws
+
+
+@njit
+def _backtrack_line_search(X, y, w, Xw, datafit, penalty, delta_w_ws, X_delta_w_ws,
+                           ws, max_backtrack_iter):
+    # inplace update of w and Xw
+    # return grad_ws of the last w and Xw
+    step, prev_step = 1., 0.
+    prev_penalty_val = penalty.value(w[ws])
```
mathurinm, line 250: doesn't this need to be updated ?

skglm/solvers/pn_solver.py
```diff
@@ -0,0 +1,295 @@
...
+        if stop_crit < 0:
+            break
+        else:
+            prev_step = step
+            step /= 2
+
+    return grad_ws
+
+
+# sparse version of func above
```
mathurinm, line 270: same as above

skglm/solvers/pn_solver_improved.py
```diff
@@ -0,0 +1,349 @@
...
+    """Compute grad of datafit restricted to ``ws`` in case ``X`` sparse."""
+    raw_grad = datafit.raw_grad(y, Xw)
+    grad = np.zeros(len(ws))
+    for idx, j in enumerate(ws):
+        grad[idx] = sparse_xj_dot(X_data, X_indptr, X_indices, j, raw_grad)
+    return grad
+
+
+@njit(fastmath=True)
+def sparse_xj_dot(data, indptr, indices, j, other):
```
mathurinm, line 319: cosmetic: X_data, X_indices, X_indptr for consistency with existing codebase

skglm/solvers/pn_solver_improved.py
```diff
@@ -0,0 +1,349 @@
...
+    """Compute ``X[:, j] @ other`` in case ``X`` sparse."""
+    res = 0.
+    for i in range(indptr[j], indptr[j+1]):
+        res += data[i] * other[indices[i]]
+    return res
+
+
+@njit(fastmath=True)
+def sparse_weighted_dot(data, indptr, indices, j, other, weights):
```
mathurinm, line 328: same, X_data

mathurinm (commented):
Thanks for the huge work @Badr-MOUFAD.

So the main acceleration factors are:
- fastmath
- blitz-like number of iterations depending on outer iteration 

are there some other?

Also I'm unsure of what we do in the backtracking line search: prev_penalty_val is not updated ?
The LS that is coded in celer is wrong and only works because, miraculously, step_size=1 always satisfies the descent condition

skglm/solvers/pn_solver.py
tomMoral: :smiling_imp:

skglm/datafits/single_task.py
Badr-MOUFAD: The Hessian is diagonal as the datafit is assumed to be separable, in particular for the logreg:

$$F(u) = \sum_{i=1}^n log(1+e^{-y_i u_i})$$

mathurinm: yes I agree, but the Hessian is still a n by n matrix, while here you return only its diagonal entries 
anyways, not very important. I meant that you may want to clarify that you return only the diagonal because the Hessian is diagonal.

skglm/penalties/separable.py
Badr-MOUFAD: Yes indeed, but here what differs is that I pass in ``w_ws``  instead of ``w``

https://github.com/Badr-MOUFAD/skglm/blob/baea36a2fcc3ff5b6aaac308965a4179bb3cee74/skglm/solvers/pn_solver.py#L226

skglm/solvers/pn_solver.py
Badr-MOUFAD: ``prev_penalty_val`` refers to the value of the penalty in the epoch loop.
sorry, perhaps the name of the var is a bit misleading. I renamed the variable to avoid confusion.

skglm/solvers/pn_solver.py
Badr-MOUFAD: Should I put a link to Blitz/celer code?

mathurinm (approved):

skglm/penalties/separable.py
QB3: in order to be 100% sure and clean, instead of creating `w_ws`, one could create a copy `w_copy` of the same size of `w`, and then  call `penalty.subdiff_distance(w_copy, cached_grads, ws)`. I think this is what we do in the rest of the code.

If this does not decrease performance, I would vote for this option.

skglm/solvers/prox_newton.py
```diff
@@ -0,0 +1,338 @@
...
+            opt = penalty.subdiff_distance(w_ws, cached_grads, ws)
+            if np.max(opt) <= tol:
+                break
+
+    # descent direction
+    return w_ws - w_epoch[ws], X_delta_w_ws
+
+
+@njit
+def _backtrack_line_search(X, y, w, Xw, datafit, penalty, delta_w_ws,
```
QB3, line 235: This seems like a very clean backtracking line-search to me.
In addition you managed to remove the (unclear) penalty subgradient call from blitz (https://github.com/tbjohns/BlitzL1/blob/8c1d2f6251307021ff4c8e2027df2d92b10d7f63/src/solver.cpp#L236).

If we say we minimize f + g
From the code I guess you are finding a point such that  g(t + delta x) - g(t) + t \nabla f (x) \Delta x < 0, which yields a descent direction since the functions are convex (slide 4 of https://www.di.ens.fr/~aspremon/PDF/ENSAE/Newton.pdf).

IMO this should be explained in the doc of the `_backtrack_line_search` function.

skglm/solvers/prox_newton.py
```diff
@@ -0,0 +1,338 @@
...
+                           X_delta_w_ws, ws):
+    # inplace update of w and Xw
+    # return grad_ws of the last w and Xw
+    step, prev_step = 1., 0.
+    epoch_penalty_val = penalty.value(w[ws])
```
QB3, line 240: > `epoch_penalty_val`

I would change then name, maybe for `penalty_val_old`

skglm/solvers/prox_newton.py
```diff
@@ -0,0 +1,338 @@
...
+
+    for backtrack_iter in range(MAX_BACKTRACK_ITER):
+        stop_crit = -epoch_penalty_val
+        w[ws] += (step - prev_step) * delta_w_ws
+        Xw += (step - prev_step) * X_delta_w_ws
+
+        grad_ws = _construct_grad(X, y, w, Xw, datafit, ws)
+        stop_crit += step * grad_ws @ delta_w_ws
+        stop_crit += penalty.value(w[ws])
```
QB3, line 249: For the sake of clarity I would remove l243 and change l249 for smth like 

`stop_crit += penalty.value(w[ws]) - penalty_val_old`

skglm/solvers/prox_newton.py
QB3: I think we should also add a clear reference to the backtrack line search we are using, maybe Nocedal and Liu Chap. 3.

skglm/solvers/prox_newton.py
```diff
@@ -0,0 +1,338 @@
...
+
+        if stop_crit < 0:
```
QB3, line 251: Finally, since we are computing the descent direction approximately, is it possible that the backtracking line search does not provide a better objective? In other words what is happening if `MAX_BACKTRACK_ITER` is attained? Should we handle this corner case? any idea @mathurinm @Klopfe ?

QB3: In addition, I think we might need to think about to scope of the penalties taken into account, is it possible to use this type of criterion for non-convex penalties? At least maybe we could add a warning in the doc.

skglm/solvers/prox_newton.py
QB3: In addition, sadly you need to call 
`penalty.value(w)` because of the weighted Lasso penalty possibility. (Btw we might want to change this in another PR, and call penalty.value(w, ws))

skglm/solvers/prox_newton.py
```diff
@@ -0,0 +1,338 @@
...
+        else:
+            prev_step = step
+            step /= 2
+
+    return grad_ws
+
+
+# sparse version of _backtrack_line_search
+@njit
+def _backtrack_line_search_s(X_data, X_indptr, X_indices, y, w, Xw, datafit,
```
QB3, line 262: all the comments on ` _backtrack_line_searc` also holds for this function

QB3 (commented):
Thank you so much for the incredible work, here are a few (sometimes hard) comments

skglm/penalties/separable.py
mathurinm: To avoid the creation of a potentially large array (`w_copy`), @Badr-MOUFAD would it work to pass the restricted w, the gradient, and  `ws=np.arange(len(restricted_w)` ?

I'm proposing this because I'm strongly in favor of leaving this function untouched and finding a workaround directly in the solver.

skglm/solvers/prox_newton.py
```diff
@@ -0,0 +1,338 @@
...
+
+
+EPS_TOL = 0.3
+MAX_CD_ITER = 20
+MAX_BACKTRACK_ITER = 20
+
+
+def prox_newton(X, y, datafit, penalty, w_init=None, p0=10,
+                max_iter=20, max_epochs=1000, tol=1e-4, verbose=0):
+    """Run a Prox Newton solver.
```
mathurinm, line 13: I may help to Add refs to the blitz and celer (JMLR) papers and codebases. You can put (see References at this line, and a reference section at the end).

skglm/solvers/prox_newton.py
```diff
@@ -0,0 +1,338 @@
...
+
+    Parameters
+    ----------
+    X : array or sparse CSC matrix shape (n_samples, n_features)
```
mathurinm, line 17:
```suggestion
    X : array or sparse CSC matrix, shape (n_samples, n_features)
```

skglm/solvers/prox_newton.py
mathurinm: also specify that it is combined with working sets

skglm/solvers/prox_newton.py
```diff
@@ -0,0 +1,338 @@
...
+        If set to None, a zero vector is used instead.
+
+    p0 : int, default 10
+        Minimum number of features to be included in the working set.
+
+    max_iter : int, default 20
+        Maximum number of outer iterations.
+
+    max_epochs : int, default 1000
+        Maximum number of epochs.
```
mathurinm, line 40:
```suggestion
        Maximum number of CD epochs to find the descent direction.
```

skglm/solvers/prox_newton.py
```diff
@@ -0,0 +1,338 @@
...
+        gsupp_size = penalty.generalized_support(w).sum()
+        ws_size = max(min(p0, n_features),
+                      min(n_features, 2 * gsupp_size))
+        # similar to np.argsort()[-ws_size:] but without sorting
+        ws = np.argpartition(opt, -ws_size)[-ws_size:]
+
+        grad_ws = grad[ws]
+        tol_in = EPS_TOL * stop_crit
+
+        for epoch in range(max_epochs):
```
mathurinm, line 103: I thought epochs referred to the number of CD passes to find the descent direction, but in fact it's the number of PN updates on the subproblems. Is `max_pn_iter` a better name @Badr-MOUFAD ?

skglm/solvers/prox_newton.py
```diff
@@ -0,0 +1,338 @@
...
+        Initial value of coefficients.
+        If set to None, a zero vector is used instead.
+
+    p0 : int, default 10
+        Minimum number of features to be included in the working set.
+
+    max_iter : int, default 20
+        Maximum number of outer iterations.
+
+    max_epochs : int, default 1000
```
mathurinm, line 39: todo change name and description

skglm/solvers/prox_newton.py
```diff
@@ -0,0 +1,338 @@
...
+    # Minimize quadratic approximation for delta_w = w - w_epoch:
+    #  b.T @ X @ delta_w + \
+    #  1/2 * delta_w.T @ (X.T @ D @ X) @ delta_w + penalty(w)
+    raw_hess = datafit.raw_hessian(y, Xw_epoch)
+
+    lipschitz = np.zeros(len(ws))
+    for idx, j in enumerate(ws):
+        lipschitz[idx] = raw_hess @ X[:, j] ** 2
+
+    cached_grads = np.zeros(len(ws))
```
mathurinm, line 159: do you know why we call this cached_grads ?

skglm/solvers/prox_newton.py
```diff
@@ -0,0 +1,338 @@
...
+            stepsize = 1 / lipschitz[idx]
+
+            w_ws[idx] = penalty.prox_1d(
+                old_w_idx - stepsize * cached_grads[idx], stepsize, j)
+
+            if w_ws[idx] != old_w_idx:
+                X_delta_w_ws += (w_ws[idx] - old_w_idx) * X[:, j]
+
+        if cd_iter % 5 == 0:
+            opt = penalty.subdiff_distance(w_ws, cached_grads, ws)
```
mathurinm, line 180: so the optimality condition is approximated (to avoid recomputing the gradients fully); but the gradients are not exact (all of them but 1 are from older updates). Right?

skglm/solvers/prox_newton.py
```diff
@@ -0,0 +1,338 @@
...
+# sparse version of _compute_descent_direction
+@njit
+def _descent_direction_s(X_data, X_indptr, X_indices, y, w_epoch,
+                         Xw_epoch, grad_ws, datafit, penalty, ws, tol):
+    raw_hess = datafit.raw_hessian(y, Xw_epoch)
+
+    lipschitz = np.zeros(len(ws))
+    for idx, j in enumerate(ws):
+        # equivalent to: lipschitz[idx] += raw_hess * X[:, j] ** 2
+        lipschitz[idx] = _sparse_squared_weighted_norm(X_data, X_indptr, X_indices,
```
mathurinm, line 197: cosmetic: I prefer breaking the line after the (

skglm/solvers/prox_newton.py
```diff
@@ -0,0 +1,338 @@
...
+
+    for cd_iter in range(MAX_CD_ITER):
+        for idx, j in enumerate(ws):
+            # skip when X[:, j] == 0
+            if lipschitz[idx] == 0:
+                continue
+
+            cached_grads[idx] = grad_ws[idx]
+            # equivalent to cached_grads[idx] += X[:, j] @ (raw_hess * X_delta_w_ws)
+            cached_grads[idx] += _sparse_weighted_dot(X_data, X_indptr, X_indices, j,
```
mathurinm, line 212: same cosmetic breakline

skglm/solvers/prox_newton.py
```diff
@@ -0,0 +1,338 @@
...
+
+
+@njit
+def _backtrack_line_search(X, y, w, Xw, datafit, penalty, delta_w_ws,
+                           X_delta_w_ws, ws):
+    # inplace update of w and Xw
+    # return grad_ws of the last w and Xw
+    step, prev_step = 1., 0.
+    epoch_penalty_val = penalty.value(w[ws])
+
```
mathurinm, line 243: I would add a comment: first try a step of 1, then 1/2, then 1/4, etc.

skglm/solvers/prox_newton.py
Badr-MOUFAD: I'm a bit confused @mathurinm,
we perform epochs to solve a subproblem (problem restricted to ws)?
This is the naming convention we adopt for other solvers

skglm/tests/test_datafits.py
```diff
@@ -29,3 +29,27 @@ def test_huber_datafit():
...
+    Xw = X @ w
+
+    log_datafit = Logistic()
+    grad = log_datafit.raw_grad(y, Xw)
+    hess = log_datafit.raw_hessian(y, Xw)
+
+    np.testing.assert_equal(grad.shape, (n_samples,))
+    np.testing.assert_equal(hess.shape, (n_samples,))
+
+    exp_yXw = np.exp(-y * Xw)
```
mathurinm, line 49: this can be error prone: this variable name implies that it's value is `np.exp(y * Xw)`

skglm/tests/test_pn_solver.py
```diff
@@ -0,0 +1,50 @@
+import pytest
+import numpy as np
+from itertools import product
+
+from skglm.penalties import L1
+from skglm.datafits import Logistic
+from skglm.solvers.prox_newton import prox_newton
+from sklearn.linear_model import LogisticRegression
```
mathurinm, line 8: move this import in the above block

skglm/solvers/prox_newton.py
Badr-MOUFAD: cached_grads are the grads of the quadratic approximation, and we use them in 
- soft thresh update
- check convergence when solving the quadratic approximation

the cached prefix refers to the fact that we are computing them only once: When performing the soft thresh updates.

@mathurinm, do you have a better name?

skglm/solvers/prox_newton.py
Badr-MOUFAD: no, it's not
refer to https://github.com/scikit-learn-contrib/skglm/pull/51#discussion_r952329694

skglm/tests/test_datafits.py
Badr-MOUFAD: yes indeed!

any suggestion?

skglm/penalties/separable.py
Badr-MOUFAD: Yes absolutely, it will work, (actually I initially implemented it this way)
However, this trick won't work if we are using a penalty that depends on the feature e.g. WeightedL1

mathurinm: indeed, and such a bug will be nasty to catch.
So I'd go with the solution @QB3 proposed, instantiate a full array of shape  `n_features,`

skglm/solvers/prox_newton.py
```diff
@@ -0,0 +1,351 @@
...
+    objs_out: array (max_iter,)
+        The objective values at every outer iteration.
+
+    stop_crit: float
+        The value of the stopping criterion when the solver stops.
+
+    References
+    ----------
+    .. [1] M. Massias, A. Gramfort, J. Salmon
```
mathurinm, line 61: Celer-PN is detailed in the JMLR paper : https://arxiv.org/abs/1907.05830 (put this link as there is an error in the JMLR website version)

skglm/solvers/prox_newton.py
```diff
@@ -0,0 +1,351 @@
...
+        "Celer: a Fast Solver for the Lasso wit Dual Extrapolation", ICML 2018,
+        http://proceedings.mlr.press/v80/massias18a.html
+
+    .. [2] Johnson, T. B. and Guestrin, C.
+        "Blitz: A principled meta-algorithm for scaling sparse optimization",
+        ICML, pp. 1171-1179, 2015.
+        https://proceedings.mlr.press/v37/johnson15.html
```
mathurinm, line 68: also add a link to the codebase

skglm/solvers/prox_newton.py
mathurinm: One epoch means one pass of CD (or SGD) over the dataset (all features/all samples for SGD)

skglm/solvers/prox_newton.py
mathurinm: every time we update Xw, the gradient wrt any feature changes, no?
So if you compute gradient wrt to feature 1, update feature, compute gradient wrt feature 2, update feat 2, then hasn't the gradient for feature 1?

skglm/tests/test_datafits.py
mathurinm: exp_minusyXw
it's in a test so it's not critical

skglm/solvers/prox_newton.py
QB3: `old_grads` ?

skglm/solvers/prox_newton.py
```diff
@@ -0,0 +1,351 @@
...
+@njit
+def _backtrack_line_search(X, y, w, Xw, datafit, penalty, delta_w_ws,
+                           X_delta_w_ws, ws):
+    # 1) find step such that:
+    #   penalty(w + step * delta_w) - penalty(w) +
+    #   step * \nabla datafit(w + step * delta_w) @ delta_w < 0
+    # ref: https://www.di.ens.fr/~aspremon/PDF/ENSAE/Newton.pdf
+    # 2) inplace update of w and Xw and return grad_ws of the last w and Xw
+    step, prev_step = 1., 0.
+    old_penalty_val = penalty.value(w[ws])
```
QB3, line 254: I still that calling `penalty.value` on `w[ws]` will cause problems for the weighted l1 norm, and currently one has to write `penalty.value(w)`

skglm/solvers/prox_newton.py
```diff
@@ -0,0 +1,351 @@
...
+
+    # try step = 1, 1/2, 1/4, ...
+    for backtrack_iter in range(MAX_BACKTRACK_ITER):
+        w[ws] += (step - prev_step) * delta_w_ws
+        Xw += (step - prev_step) * X_delta_w_ws
+
+        grad_ws = _construct_grad(X, y, w, Xw, datafit, ws)
+        stop_crit = penalty.value(w[ws]) - old_penalty_val
```
QB3, line 262: same here

skglm/solvers/prox_newton.py
mathurinm: Yes. Let's do it as QBE says and put a todo to update in another PR

skglm/solvers/prox_newton.py
Badr-MOUFAD: ``quadratic_grads``?

mathurinm: you need to insist on the fact that they are not up to date gradients

skglm/solvers/prox_newton.py
```diff
@@ -0,0 +1,351 @@
...
+
+    return grad_ws
+
+
+# sparse version of _backtrack_line_search
+@njit
+def _backtrack_line_search_s(X_data, X_indptr, X_indices, y, w, Xw, datafit,
+                             penalty, delta_w_ws, X_delta_w_ws, ws):
+    step, prev_step = 1., 0.
+    epoch_penalty_val = penalty.value(w[ws])
```
QB3: same here

skglm/solvers/prox_newton.py
```diff
@@ -0,0 +1,351 @@
...
+
+    for backtrack_iter in range(MAX_BACKTRACK_ITER):
+        stop_crit = -epoch_penalty_val
+        w[ws] += (step - prev_step) * delta_w_ws
+        Xw += (step - prev_step) * X_delta_w_ws
+
+        grad_ws = _construct_grad_sparse(X_data, X_indptr, X_indices,
+                                         y, w, Xw, datafit, ws)
+        stop_crit += step * grad_ws.T @ delta_w_ws
+        stop_crit += penalty.value(w[ws])
```
QB3: same here

skglm/solvers/prox_newton.py
```diff
@@ -0,0 +1,369 @@
...
+                )
+
+            if stop_crit_in <= tol_in:
+                if max(verbose-1, 0):
+                    print("Early exit")
+                break
+
+        p_obj = datafit.value(y, w, Xw) + penalty.value(w)
+        p_objs_out.append(p_obj)
+    return w, p_objs_out, stop_crit
```
mathurinm, line 153:
```suggestion
    return w, np.asarray(p_objs_out), stop_crit
```

skglm/solvers/prox_newton.py
```diff
@@ -0,0 +1,369 @@
...
+
+    verbose : bool, default False
+        Amount of verbosity. 0/False is silent.
+
+    Returns
+    -------
+    w : array, shape (n_features,)
+        Solution that minimizes the problem defined by datafit and penalty.
+
+    objs_out : array (max_iter,)
```
mathurinm, line 53:
```suggestion
    objs_out : array, shape (n_iter,)
```

skglm/solvers/prox_newton.py
mathurinm: I have put a TODO. We can see what happens with non convex penalties experimentally later if that is OK

skglm/solvers/prox_newton.py
```diff
@@ -0,0 +1,369 @@
...
+            step /= 2
+    else:
+        pass  # TODO
+
+    return grad_ws
+
+
+@njit
+def _construct_grad(X, y, w, Xw, datafit, ws):
+    # Compute grad of datafit restricted to ws
```
mathurinm, line 320:
```suggestion
    # Compute grad of datafit restricted to ws. This function avoids 
    # recomputing raw_grad for every j, which is costly for logreg
```

skglm/tests/test_pn_solver.py
```diff
@@ -0,0 +1,50 @@
...
+from skglm.penalties import L1
+from skglm.datafits import Logistic
+from skglm.solvers.prox_newton import prox_newton
+from skglm.utils import make_correlated_data, compiled_clone
+
+
+@pytest.mark.parametrize('X_density', [1, 0.5])
+def test_alpha_max(X_density):
+    n_samples, n_features = 10, 20
+    X, y, _ = make_correlated_data(n_samples, n_features, X_density=X_density)
```
mathurinm, line 15:
```suggestion
    X, y, _ = make_correlated_data(
        n_samples, n_features, X_density=X_density, random_state=2)
```

skglm/tests/test_pn_solver.py
```diff
@@ -0,0 +1,50 @@
...
+    log_datafit = compiled_clone(Logistic())
+    l1_penalty = compiled_clone(L1(alpha_max))
+    w = prox_newton(X, y, log_datafit, l1_penalty)[0]
+
+    np.testing.assert_equal(w, 0)
+
+
+@pytest.mark.parametrize("rho, X_density", product([1e-1, 1e-2], [1., 0.5]))
+def test_pn_vs_sklearn(rho, X_density):
+    n_samples, n_features = 10, 20
```
mathurinm, line 29:
```suggestion
    n_samples, n_features = 11, 19
```

PR #52, opened 2022-08-19
[READY] DOC - add example of scikit-learn ``GridSearchCV``

Badr-MOUFAD:
This verifies that we do support [scikit-learn ``GridSearchCV``](https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.GridSearchCV.html) and thereby adds an example to the documentation to illustrate that.

### DIFF
--- a/examples/grid_search_cv.py
+++ b/examples/grid_search_cv.py
@@ -1,0 +1,35 @@
+"""
+==========================================
+Skglm support of scikit-learn GridSearchCV
+==========================================
+An example that uses scikit-learn``GridSearchCV`` to select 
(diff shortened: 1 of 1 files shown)


mathurinm:
@Badr-MOUFAD thanks! the example name must start with `plot_` otherwise it is not run (that's why there is a running time of 0s here: https://output.circle-artifacts.com/output/job/ce6fb6a7-a421-48ad-b1e3-554d3621231d/artifacts/0/dev/auto_examples/grid_search_cv.html#sphx-glr-auto-examples-grid-search-cv-py)

In addition can you add a plot inside the example, to make the example visual ? (and so that sphinx picks it up as a thumbnail)

PR #54, opened 2022-08-22
[READY] remove inheritance from sklearn Lasso

mathurinm:
split of #19 with only the part concerning the removal of sklearn Lasso inheritance


skglm/estimators.py
```diff
@@ -1,22 +1,25 @@
...
-from sklearn.linear_model import ElasticNet as ElasticNet_sklearn
-from sklearn.linear_model import LogisticRegression as LogReg_sklearn
 from sklearn.linear_model import MultiTaskLasso as MultiTaskLasso_sklearn
-from sklearn.linear_model._base import LinearModel
+from sklearn.linear_model._base import LinearModel, RegressorMixin,\
+    LinearClassifierMixin, SparseCoefMixin, BaseEstimator
+
 from sklearn.svm import LinearSVC as LinearSVC_sklearn
 from sklearn.preprocessing import LabelEncoder
 from sklearn.multiclass import OneVsRestClassifier
```
mathurinm, line 20: rm blank line

skglm/estimators.py
```diff
@@ -26,6 +29,120 @@
 )
 from skglm.utils import compiled_clone
 from skglm.solvers import cd_solver_path, bcd_solver_path
+from skglm.solvers.cd_solver import cd_solver
+
+
+def glm_fit(X, y, model, datafit, penalty):
```
mathurinm, line 35: Should we make this private ? Otherwise we need a docstring

skglm/estimators.py
```diff
@@ -26,6 +29,120 @@
...
+    is_classif = False
+
+    for base in model.__class__.__bases__:
+        if base.__name__ in ["ClassifierMixin", "LinearClassifierMixin"]:
+            is_classif = True
+    if y is None:
+        raise ValueError("requires y to be passed, but the target y is None")
+
+    penalty = penalty if penalty else L1(1.)
```
mathurinm, line 70: is this still needed ? it's a bit mysterious so it deserves a comment

skglm/estimators.py
```diff
@@ -26,6 +29,120 @@
...
+            X, y, validate_separately=(check_X_params, check_y_params))
+        X = check_array(X, 'csc', dtype=[np.float64, np.float32],
+                        order='F', copy=False, accept_large_sparse=False)
+        y = check_array(y, 'csc', dtype=X.dtype.type, order='F', copy=False,
+                        ensure_2d=False)
+    if y.ndim == 2 and y.shape[1] == 1:
+        warnings.warn("DataConversionWarning('A column-vector y"
+                      " was passed when a 1d array was expected")
+        y = y[:, 0]
+    penalty = compiled_clone(penalty)
```
mathurinm, line 106: here I would break line to indicate we start fitting after checking input

skglm/estimators.py
```diff
@@ -26,6 +29,120 @@
...
+    datafit = compiled_clone(datafit, to_float32=X.dtype == np.float32)
+    if issparse(X):
+        datafit.initialize_sparse(X.data, X.indptr, X.indices, y)
```
mathurinm, line 109: init should be done after all checks, like checking that X and y shapes match

skglm/estimators.py
```diff
@@ -26,6 +29,120 @@
...
+        model.coef_ = None
+
+    if isinstance(datafit, QuadraticSVC):
+        if issparse:
+            yXT = (X.T).multiply(y)
+        else:
+            yXT = (X * y[:, None]).T
+
+    X_ = yXT if isinstance(datafit, QuadraticSVC) else X
+    # TODO allow warm start
```
mathurinm, line 131: This can be done L97, with an else following https://github.com/scikit-learn-contrib/skglm/pull/54/files#diff-5a55ebbdbb198f456b4549cc660c2b1e6fc838fd05424e8f4b7775239aed32b9R96

skglm/estimators.py
```diff
@@ -26,6 +29,120 @@
...
+    w = np.zeros(X_.shape[1], dtype=X_.dtype)
+    Xw = np.zeros(X_.shape[0], dtype=X_.dtype)
+
+    # check consistency of weights for WeightedL1
+    if isinstance(penalty, WeightedL1):
+        if len(penalty.weights) != X.shape[1]:
+            raise ValueError(
+                "The size of the WeightedL1 penalty should be n_features")
```
mathurinm, line 139: Make the error message explicit: expected..., got ...

skglm/estimators.py
```diff
@@ -26,6 +29,120 @@
...
+                model.coef_ = np.array(primal_coef).reshape(1, -1)
+        elif n_classes_ > 2:
+            model.coef_ = np.empty([len(model.classes_), X.shape[1]])
+            model.intercept_ = 0
+            multiclass = OneVsRestClassifier(model).fit(X, y)
+            model.coef_ = np.array([clf.coef_[0]
+                                    for clf in multiclass.estimators_])
+            model.n_iter_ = max(
+                clf.n_iter_ for clf in multiclass.estimators_)
+    elif isinstance(datafit, Logistic):
```
mathurinm, line 169: Logistic is for classif, I believe we will never this block ?

skglm/solvers/cd_solver.py
```diff
@@ -345,7 +345,7 @@ def cd_solver(
                             print("Early exit")
                         break
         obj_out.append(p_obj)
-    return w, np.array(obj_out), stop_crit
+    return w, t, stop_crit
```
mathurinm, line 348: this can break existing code. Is it completely needed ? Can you get it from outside by multiplying `len(obj_out)` by 10 ?

skglm/estimators.py
Klopfe: I have added docstring for now. I can also be private function eventually

mathurinm: I think it should be private. Do you envision a case where a user would need it?

Klopfe: Not really, I changed it to a private function :)

skglm/estimators.py
```diff
@@ -26,6 +27,123 @@
...
+        model.coef_ = None
+
+    penalty = compiled_clone(penalty)
+    datafit = compiled_clone(datafit, to_float32=X.dtype == np.float32)
+    if issparse(X):
+        datafit.initialize_sparse(X.data, X.indptr, X.indices, y)
+    else:
+        datafit.initialize(X, y)
+
+    if isinstance(datafit, QuadraticSVC):
```
Badr-MOUFAD, line 93: these cases are special cases (rep WeighedLasso). I think they should be treated inside the model rather than a function with if/else statements.

skglm/estimators.py
```diff
@@ -26,6 +27,123 @@
...
+
+    if model.warm_start and hasattr(model, 'coef_'):
+        w = model.coef_.copy()
+        Xw = X_ @ w
+    else:
+        w = np.zeros(X_.shape[1], dtype=X_.dtype)
+        Xw = np.zeros(X_.shape[0], dtype=X_.dtype)
+
+    # check consistency of weights for WeightedL1
+    if isinstance(penalty, WeightedL1):
```
Badr-MOUFAD, line 110: same remark as ``QuadraticSVC`` in L93

skglm/estimators.py
```diff
@@ -26,6 +27,123 @@
...
+        if issparse:
+            yXT = (X.T).multiply(y)
+        else:
+            yXT = (X * y[:, None]).T
+        X_ = yXT
+    else:
+        X_ = X
+
+    if model.warm_start and hasattr(model, 'coef_'):
+        w = model.coef_.copy()
```
mathurinm, line 103: this fails if coef_ is None

skglm/estimators.py
```diff
@@ -26,6 +27,123 @@
...
+                " got %i" % len(penalty.weights))
+
+    coefs, p_obj, kkt = cd_solver(
+        X_, y, datafit, penalty, w, Xw, max_iter=model.max_iter,
+        max_epochs=model.max_epochs, p0=model.p0,
+        tol=model.tol, use_acc=True, K=5, ws_strategy=model.ws_strategy,
+        verbose=model.verbose)
+
+    model.coef_, model.stop_crit_ = coefs, kkt
+    model.n_iter_ = len(p_obj)
```
mathurinm, line 124: add a comment: n_iter for us means number of subproblems (for sklearn it's number of CD epochs, not the same)

skglm/estimators.py
```diff
@@ -26,6 +27,123 @@
...
+                    primal_coef = ((yXT).multiply(model.coef_[0, :])).T
+                else:
+                    primal_coef = (yXT * model.coef_[0, :]).T
+                primal_coef = primal_coef.sum(axis=0)
+                model.coef_ = np.array(primal_coef).reshape(1, -1)
+        elif n_classes_ > 2:
+            model.coef_ = np.empty([len(model.classes_), X.shape[1]])
+            model.intercept_ = 0
+            multiclass = OneVsRestClassifier(model).fit(X, y)
+            model.coef_ = np.array([clf.coef_[0]
```
mathurinm, line 141: harmonize style with below: break after (

Badr-MOUFAD (commented):
The PR is much simpler to review now. I think we are on the right way to getting rid of sklearn inheritance, though I have the following remark:
- In our estimators, we have a fit and path methods that don't seem to use the same things (``cd_solver`` for path, and ``_glm_fit`` for fit)

and code review.

skglm/estimators.py
```diff
@@ -26,6 +27,123 @@
 )
 from skglm.utils import compiled_clone
 from skglm.solvers import cd_solver_path, bcd_solver_path
+from skglm.solvers.cd_solver import cd_solver
+
+
+def _glm_fit(X, y, model, datafit, penalty):
```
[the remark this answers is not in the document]
QB3: I am on it!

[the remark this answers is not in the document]
QB3: this should be OK now

skglm/estimators.py
QB3: I am not sure that we will be able to put in the datafit for the SVM.
The problem is that the SVM is solved in the dual, if we really want to take care of in this PR, I think the cleanest way would be to add a `solved_in_dual` parameter

mathurinm: You can't treat that in the model because this function is called by GeneralizedLinearEstimator

skglm/estimators.py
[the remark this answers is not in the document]
mathurinm: I am not sure this is relevant. How much duplicated code do you have now ?

skglm/estimators.py
QB3: it is already in the doc of estimators: "max_iter : int, optional The maximum number of iterations (subproblem definitions).", do you want to make a recall here?

mathurinm:
Note that currently I don't think we can use GeneralizeLinearEstimator with a group penalty, has it will not call `group_bcd_solver`. I wonder how the tests don't break

QB3:
> Note that currently I don't think we can use GeneralizeLinearEstimator with a group penalty, has it will not call group_bcd_solver. I wonder how the tests don't break

so running `test_group.py` should fail

skglm/tests/test_estimators.py
```diff
@@ -228,6 +227,15 @@ def test_grid_search(estimator_name):
                                ours_clf.best_params_["alpha"], rtol=1e-3)
 
 
+@pytest.mark.parametrize("estimator_name", ["Lasso", "SVC"])
```
QB3, line 230: is there a reason not to test the warm start for all the estimators?

skglm/estimators.py
```diff
@@ -26,6 +27,123 @@
...
+
+    is_classif = False
+
+    for base in model.__class__.__bases__:
+        if base.__name__ in ["ClassifierMixin", "LinearClassifierMixin"]:
+            is_classif = True
```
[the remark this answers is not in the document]
QB3: I am not sure to understand what you mean here

skglm/estimators.py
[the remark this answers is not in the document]
QB3: should we merge it back then?

QB3:
> Note that currently I don't think we can use GeneralizeLinearEstimator with a group penalty, has it will not call group_bcd_solver. I wonder how the tests don't break

`GeneralizeLinearEstimator` is currently not tested for group estimators, hence it makes sense that tests do not fail.
In `test_group.py` we directly call `group_bcd_solver`.

mathurinm:
OK, great. When we add the `Solver` class we should test all solvers with the `GLE` class

skglm/tests/test_estimators.py
mathurinm: I just wanted the test to be fast. I was thinking if it worked for one estimator it would work for all of them

skglm/estimators.py
mathurinm: Nope you are right, I forgot it was there

skglm/estimators.py
mathurinm: we can have a penalty.check_compatible(X, y)  but that's more work for the user when they define penalties.

skglm/estimators.py
```diff
@@ -1,31 +1,201 @@
...
+        model.coef_ = None
+
+    penalty = compiled_clone(penalty)
+    datafit = compiled_clone(datafit, to_float32=X.dtype == np.float32)
+    if issparse(X):
+        datafit.initialize_sparse(X.data, X.indptr, X.indices, y)
+    else:
+        datafit.initialize(X, y)
+
+    X_ = X
```
mathurinm, line 64: This is not needed in this function, it's reminiscent from glm_classif. It could be confusing. I would merge the two functions back

skglm/estimators.py
```diff
@@ -1,31 +1,201 @@
...
+        # check consistency of weights for WeightedL1
+        if isinstance(penalty, WeightedL1):
+            if len(penalty.weights) != X.shape[1]:
+                raise ValueError(
+                    "The size of the WeightedL1 penalty should be n_features, \
+                    expected %i, got %i" % (X_.shape[1], len(penalty.weights)))
+
+        coefs, p_obj, kkt = cd_solver(
+            X_, y, datafit_jit, penalty, w, Xw, max_iter=model.max_iter,
+            max_epochs=model.max_epochs, p0=X_.shape[1],
```
mathurinm, line 167: @Klopfe why is p0 equal to X_.shape here ? Avoid writing such code without a TODO comment, it can very well go unnoticed, then be merged and it would take a long time to find it back

Klopfe: I don't think that the "why" question is necessary. I did not leave a mistake on purpose. I was debugging the SVC. Now, you're right I should have been more cautious. I will pay attention. next time.

mathurinm: I was asking if if was for debugging purposes or if there was a reason it was still there. Sorry if the tone was not friendly, it was not my intention.

Klopfe: I misunderstood then. It should have been back to p0=10 and it could have been very long to find out. You're right

skglm/estimators.py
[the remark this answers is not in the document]
QB3: done!

skglm/estimators.py
QB3: I went back to `p0=model.p0,` should we put a different value for  the SVC?

skglm/estimators.py
```diff
@@ -1,31 +1,163 @@
...
+    else:
+        solver = cd_solver if y.ndim == 1 else multitask_bcd_solver
+    # TODO this must be replaced by an instance of BaseSolver being passed
+    # so that arguments are attributes of the `solver` object and arguments
+    # do not need to match across solvers
+    # QUESTION should p0 be different for SVC ?
+    coefs, p_obj, kkt = solver(
+        X_, y, datafit_jit, penalty, w, Xw, max_iter=model.max_iter,
+        max_epochs=model.max_epochs, p0=model.p0,
+        tol=model.tol,  # ws_strategy=model.ws_strategy,
```
QB3, line 143: the `ws_strategy` was commented, should we add it back? or do not let the choice to the user?

skglm/tests/test_estimators.py
QB3: OK, I just added all the estimators to the test!

skglm/estimators.py
```diff
@@ -26,6 +27,123 @@
...
+
+def _glm_fit(X, y, model, datafit, penalty):
+
+    is_classif = False
+
+    for base in model.__class__.__bases__:
+        if base.__name__ in ["ClassifierMixin", "LinearClassifierMixin"]:
+            is_classif = True
+
+    if is_classif:
```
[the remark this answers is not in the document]
QB3: was it taken from here ? https://github.com/scikit-learn/scikit-learn/blob/5bd81234e6e6501ddcddbfdfdc80b90a1302af55/sklearn/svm/_base.py#L1177

QB3:
It would be nice to add a comment to the `test_generic_get_params` function in `test_estimators.py`

QB3:
Apart from the few comments left, LGTM

skglm/estimators.py
```diff
@@ -1,31 +1,164 @@
...
+        solver = cd_solver  # TODO to be be replaced by an instance of BaseSolver
+    else:
+        solver = cd_solver if y.ndim == 1 else multitask_bcd_solver
+    # TODO this must be replaced by an instance of BaseSolver being passed
+    # so that arguments are attributes of the `solver` object and arguments
+    # do not need to match across solvers
+    # TODO QUESTIONS
+    # Should p0 be different for SVC?
```
Klopfe, line 139: I think we can leave the p0 as it is for the SVC, I will investigate the slow tests to see if I can find why it happens.

skglm/estimators.py
Klopfe: Maybe we should just do for convex penalties = subdiff and for non-convex: fixed_point until we have time to do more benchmarks. Wdyt?

skglm/estimators.py
```diff
@@ -1,31 +1,164 @@
...
+                primal_coef = ((yXT).multiply(model.coef_[0, :])).T
+            else:
+                primal_coef = (yXT * model.coef_[0, :]).T
+            primal_coef = primal_coef.sum(axis=0)
+            model.coef_ = np.array(primal_coef).reshape(1, -1)
+            model.dual_coef_ = coefs[np.newaxis, :]
+    return model
 
 
 class GeneralizedLinearEstimator(LinearModel):
```
Klopfe, line 164: It looks neat. Well done 👍

skglm/estimators.py
```diff
@@ -193,78 +326,82 @@ def fit(self, X, y):
...
-                self.intercept_ = 0
-                multiclass = OneVsRestClassifier(self).fit(X, y)
-                self.coef_ = np.array([clf.coef_[0]
-                                       for clf in multiclass.estimators_])
-                self.n_iter_ = max(
-                    clf.n_iter_ for clf in multiclass.estimators_)
-        elif isinstance(self.datafit, Logistic):
-            self.coef_ = coefs.T
-        return self
+            return _glm_fit(X, y, self, self.datafit, self.penalty)
```
Klopfe, line 402: since _glm_fit is merged back, we should just return _glm_fit once here.

skglm/estimators.py
```diff
@@ -971,8 +1179,61 @@ def path(self, X, y, alphas, coef_init=None, return_n_iter=True, **params):
...
+                # Workaround for multi_class="multinomial" and binary outcomes
+                # which requires softmax prediction with only a 1D decision.
+                decision_2d = np.c_[-decision, decision]
+            else:
+                decision_2d = decision
+            return softmax(decision_2d, copy=False)
+
 
-class LinearSVC(LinearSVC_sklearn):
+class LinearSVC(LinearClassifierMixin, SparseCoefMixin, BaseEstimator):
```
Klopfe, line 1236: I will add as a TODO for another PR to add a predict_proba for LinearSVC, probably just taking some code from sklearn.

skglm/estimators.py
Klopfe: I fixed it

skglm/estimators.py
```diff
@@ -1,31 +1,164 @@
...
+        multiclass = OneVsRestClassifier(model).fit(X, y)
+        model.coef_ = np.array(
+            [clf.coef_[0] for clf in multiclass.estimators_])
+        if isinstance(datafit, QuadraticSVC):
+            model.dual_coef_ = np.array(
+                [clf.dual_coef_[0] for clf in multiclass.estimators_])
+        model.n_iter_ = max(
+            clf.n_iter_ for clf in multiclass.estimators_)
+        return model
+    if is_classif and n_classes_ <= 2 and isinstance(datafit, QuadraticSVC):
```
mathurinm, line 90: cosmetic: when two if blocks are unrelated (not like an if elif), it often helps to put a blank line between them

mathurinm:
LGTM

skglm/estimators.py
Klopfe: Leave as it is for now and we will see in the future :)

skglm/estimators.py
```diff
@@ -1333,5 +1488,5 @@ def path(self, X, Y, alphas, coef_init=None, **params):
         datafit = compiled_clone(self.datafit, to_float32=X.dtype == np.float32)
         penalty = compiled_clone(self.penalty)
 
-        return bcd_solver_path(X, Y, datafit, penalty, alphas=alphas,
-                               coef_init=coef_init, **params)
+        return multitask_bcd_solver_path(X, Y, datafit, penalty, alphas=alphas,
```
Badr-MOUFAD, line 1491: Is there a reason why we don't call ``_glm_fit`` here?

skglm/estimators.py
```diff
@@ -117,53 +250,53 @@ def __repr__(self):
             % (self.datafit.__class__.__name__, self.penalty.__class__.__name__,
                self.penalty.alpha, self.is_classif))
 
-    def path(self, X, y, alphas, coef_init=None, return_n_iter=False, **params):
-        """Compute regularization path.
+    # def path(self, X, y, alphas, coef_init=None, return_n_iter=False, **params):
```
Badr-MOUFAD, line 253: We have two big chunks of commented code in ``GeneralizedLinearEstimator``, What are we planning to do with them? (remove, keep for future refs, )

Badr-MOUFAD (commented):
Congrats on this amazing job! I think we are almost done with inheritance from scikit-learn estimators.

I have two comments before we merge.

skglm/estimators.py
Klopfe: Here, we want to compute the path for several alphas. Hence, calling `multitask_bcd_solver_path` as it is done in the Lasso case with `cd_solver_path`. So I'm guessing it is all right

skglm/estimators.py
mathurinm: they can be removed

Klopfe: To my opinion, the code in fit can be removed as it can now be found almost completely in `_glm_fit`. I don't really see reasons to keep it but maybe the others will think otherwise.

skglm/estimators.py
Badr-MOUFAD: What about the fit method, why don't call ``_glm_fit``? In fact, we handle the case of MLT (c.f. L135).

Klopfe: You're right but we don't handle multiple penalty values in the `_glm_fit`.

Badr-MOUFAD (approved):

Badr-MOUFAD:
LGTM when it's green!

Klopfe:
Well done everyone

PR #55, opened 2022-08-22
[READY] Fit intercept inside cd_solver

mathurinm:
Second part of #19 split (first part is #54)


skglm/estimators.py
```diff
@@ -141,15 +143,26 @@ def _glm_fit(X, y, model, datafit, penalty):
...
+        # ws_strategy=model.ws_strategy,
         verbose=model.verbose)
+    if y.ndim == 1:
+        model.coef_, model.stop_crit_ = coefs[:n_features], kkt
+        if model.fit_intercept:
+            model.intercept_ = coefs[-1]
+        else:
+            model.intercept_ = 0.
+    else:
+        model.coef_, model.stop_crit_ = coefs[:n_features, :], kkt
```
mathurinm, line 156: you can do `coefs[:n_features]` and so this line becomes the same as for y.ndim==1 and they can be taken out of the if

skglm/estimators.py
```diff
@@ -141,15 +143,26 @@ def _glm_fit(X, y, model, datafit, penalty):
     coefs, p_obj, kkt = solver(
         X_, y, datafit_jit, penalty_jit, w, Xw, max_iter=model.max_iter,
         max_epochs=model.max_epochs, p0=model.p0,
-        tol=model.tol,  # ws_strategy=model.ws_strategy,
+        tol=model.tol, fit_intercept=model.fit_intercept,
+        # ws_strategy=model.ws_strategy,
         verbose=model.verbose)
+    if y.ndim == 1:
+        model.coef_, model.stop_crit_ = coefs[:n_features], kkt
+        if model.fit_intercept:
```
mathurinm, line 151: you can try a one liner: `model.intercept_ = coefs[-1] if model.fit_intercept else 0.`

skglm/solvers/cd_solver.py
```diff
@@ -235,38 +230,43 @@ def cd_solver(
     stop_crit = np.inf  # initialize for case n_iter=0
 
     is_sparse = sparse.issparse(X)
+
+    if len(w) != n_features + fit_intercept:
+        raise ValueError(
+            "Shape of w should be n_features + 1 when fit_intercept=True.")
```
mathurinm, line 236: can you add the usual expected %d, got %d ?

skglm/solvers/cd_solver.py
```diff
@@ -235,38 +230,43 @@ def cd_solver(
...
         else:
-            grad = construct_grad(X, y, w, Xw, datafit, all_feats)
+            grad = construct_grad(X, y, w[:n_features], Xw, datafit, all_feats)
 
         if ws_strategy == "subdiff":
-            opt = penalty.subdiff_distance(w, grad, all_feats)
+            opt = penalty.subdiff_distance(w[:n_features], grad, all_feats)
         elif ws_strategy == "fixpoint":
-            opt = dist_fix_point(w, grad, datafit, penalty, all_feats)
+            opt = dist_fix_point(w[:n_features], grad, datafit, penalty, all_feats)
```
mathurinm, line 248: as @Badr-MOUFAD pointed out, we don't take the intercept into account in the optimality condition. Maybe let's just write it as a comment here so that we don't forget

skglm/solvers/cd_solver.py
```diff
@@ -276,15 +276,22 @@ def cd_solver(
...
+                _cd_epoch(X, y, w[:n_features], Xw, datafit, penalty, ws)
 
+            # update intercept
+            if fit_intercept:
+                intercept_old = w[-1]
+                w[-1] -= datafit.intercept_update_step(y, Xw)
+                Xw += (w[-1] - intercept_old)
             # 3) do Anderson acceleration on smaller problem
             # TODO optimize computation using ws
             if use_acc:
```
mathurinm, line 291: I forgot, why don't we use `AndersonAccelerator` here ?

skglm/solvers/group_bcd_solver.py
```diff
@@ -62,9 +66,12 @@ def group_bcd_solver(X, y, datafit, penalty, w_init=None, Xw_init=None, p0=10,
     n_samples, n_features = X.shape
     n_groups = len(penalty.grp_ptr) - 1
 
-    w = np.zeros(n_features) if w_init is None else w_init
+    w = np.zeros(n_features + fit_intercept) if w_init is None else w_init
     Xw = np.zeros(n_samples) if w_init is None else Xw_init
-    # Xw = X @ w
+    if len(w) != n_features + fit_intercept:
+        raise ValueError(
+            "Shape of w should be n_features + 1 when fit_intercept=True.")
```
mathurinm, line 73: ditto

skglm/solvers/group_bcd_solver.py
```diff
@@ -95,6 +102,12 @@ def group_bcd_solver(X, y, datafit, penalty, w_init=None, Xw_init=None, p0=10,
             # inplace update of w and Xw
             _bcd_epoch(X, y, w, Xw, datafit, penalty, ws)
```
mathurinm, line 103: This is harmless because of ws, but for clarity probably w[:n_features] should be passed

skglm/solvers/group_bcd_solver.py
```diff
@@ -95,6 +102,12 @@ def group_bcd_solver(X, y, datafit, penalty, w_init=None, Xw_init=None, p0=10,
...
 
+            # update intercept
+            if fit_intercept:
+                intercept_old = w[-1]
+                w[-1] -= datafit.intercept_update_step(y, Xw)
+                Xw += (w[-1] - intercept_old)
+
             w_acc, Xw_acc, is_extrapolated = accelerator.extrapolate(w, Xw)
 
             if is_extrapolated:  # avoid computing p_obj for un-extrapolated w, Xw
```
mathurinm, line 113: probably below, the objective should be computed with w[:n_features]. I wonder why it does not fail, maybe it's not tested

skglm/solvers/multitask_bcd_solver.py
```diff
@@ -212,6 +219,10 @@ def multitask_bcd_solver(
     all_feats = np.arange(n_features)
     stop_crit = np.inf  # initialize for case n_iter=0
 
+    if W.shape[0] != n_features + fit_intercept:
+        raise ValueError(
+            "Shape of W should be n_features + 1 when fit_intercept=True.")
```
mathurinm, line 224: ditto

skglm/solvers/multitask_bcd_solver.py
```diff
@@ -234,13 +245,13 @@ def multitask_bcd_solver(
                       min(2 * (norm(W, axis=1) != 0).sum() - n_unpen,
                           n_features))
         opt[unpen] = np.inf  # always include unpenalized features
-        opt[norm(W, axis=1) != 0] = np.inf  # TODO check
+        opt[norm(W[:n_features, :], axis=1) != 0] = np.inf  # TODO check
```
mathurinm, line 248: The second : can be ommitted

skglm/tests/test_group.py
```diff
@@ -132,6 +132,44 @@ def test_vs_celer_grouplasso(n_groups, n_features, shuffle):
     np.testing.assert_allclose(model.coef_, w, atol=1e-5)
 
 
+@pytest.mark.parametrize("n_groups, n_features, shuffle",
+                         [[15, 50, False]])
```
mathurinm, line 136: if a single set of parameters is used, we can get rid of parametrize and hardcode these.

skglm/tests/test_intercept.py
```diff
@@ -0,0 +1,111 @@
+from skglm.solvers.cd_solver import cd_solver, cd_solver_path
+from skglm.solvers.multitask_bcd_solver import multitask_bcd_solver
+from skglm.datafits import Quadratic, Logistic, Huber, QuadraticMultiTask
+from skglm.penalties import L1, WeightedL1, L2_1
+from skglm.utils import make_correlated_data
+
+from sklearn.linear_model import HuberRegressor, Lasso, LogisticRegression
```
mathurinm, line 7: move non skglm import before skglm imports; linear_model imports can be grouped

skglm/tests/test_intercept.py
```diff
@@ -0,0 +1,111 @@
+from skglm.solvers.cd_solver import cd_solver, cd_solver_path
```
QB3, line 1: Ideally I would avoid to duplicate this test file with the one from `test_estimator.py`

QB3 (commented):
Here is a first round of comments, my main complain is the `test_intercept.py` which I think could / should be merged with `test_estimators.py`

skglm/solvers/cd_solver.py
QB3: because we still have to merge it: https://github.com/scikit-learn-contrib/skglm/pull/33

skglm/datafits/single_task.py
```diff
@@ -87,6 +87,9 @@ def full_grad_sparse(
             grad[j] = (XjTXw - self.Xty[j]) / n_samples
         return grad
 
+    def intercept_update_step(self, y, Xw):
+        return np.sum(Xw - y) / len(Xw)
```
mathurinm, line 91: maybe np.mean is more readable here (very minor)

skglm/datafits/single_task.py
```diff
@@ -169,6 +172,9 @@ def gradient_scalar_sparse(self, X_data, X_indptr, X_indices, y, Xw, j):
             grad -= X_data[i] * y[idx_i] * sigmoid(- y[idx_i] * Xw[idx_i])
         return grad / len(Xw)
 
+    def intercept_update_step(self, y, Xw):
+        return np.sum((- y * sigmoid(- y * Xw))) / (4 * len(Xw))
```
mathurinm, line 176: np.mean / 4 ?

skglm/estimators.py
```diff
@@ -141,15 +143,26 @@ def _glm_fit(X, y, model, datafit, penalty):
...
+        else:
+            model.intercept_ = np.zeros(y.shape[1])
 
-    model.coef_, model.stop_crit_ = coefs, kkt
     model.n_iter_ = len(p_obj)
-    model.intercept_ = 0.
 
     if is_classif and n_classes_ <= 2:
-        model.coef_ = coefs[np.newaxis, :]
+        model.coef_ = coefs[np.newaxis, :n_features]
```
mathurinm, line 165: Use None instead of np.newaxis  (btw they are the same object)

skglm/solvers/cd_solver.py
```diff
@@ -64,7 +67,7 @@ def cd_solver_path(X, y, datafit, penalty, alphas=None,
     alphas : array, shape (n_alphas,)
         The alphas along the path where models are computed.
 
-    coefs : array, shape (n_features, n_alphas)
+    coefs : array, shape (n_features + 1, n_alphas)
```
mathurinm, line 70: typo: `+fit_intercept` instead of +1

skglm/solvers/multitask_bcd_solver.py
```diff
@@ -66,7 +69,7 @@ def multitask_bcd_solver_path(
     alphas : array, shape (n_alphas,)
         The alphas along the path where models are computed.
 
-    coefs : array, shape (n_features, n_tasks, n_alphas)
+    coefs : array, shape (n_features + 1, n_tasks, n_alphas)
```
mathurinm, line 72: + fit_intercept

skglm/tests/test_estimators.py
```diff
@@ -97,27 +97,38 @@ def test_check_estimator(estimator_name):
...
 @pytest.mark.parametrize("estimator_name", dict_estimators_ours.keys())
 @pytest.mark.parametrize('X', [X, X_sparse])
-def test_estimator(estimator_name, X):
+@pytest.mark.parametrize('fit_intercept', [True, False])
+def test_estimator(estimator_name, X, fit_intercept):
     if estimator_name == "GeneralizedLinearEstimator":
         pytest.skip()
     estimator_sk = dict_estimators_sk[estimator_name]
     estimator_ours = dict_estimators_ours[estimator_name]
+    # TODO This seems a bit unusal, maybe to discuss
```
mathurinm, line 106: @QB3  this has potentially harmful side effects: it sets the intercept to True on an object defined outside the function, so if we first call the function with fit_intercept=True, we set to `model.fit_intercept` to True, then the second time the function is called, the model still has `clf.fit_intercept` set to True.

mathurinm: it can be avoided by cloning the estimator when entering the test instead (sklearn.base.clone)

skglm/tests/test_intercept.py
```diff
@@ -0,0 +1,111 @@
...
+    coefs_ours = coefs_ours[:X.shape[1]]
+    coefs_sk = dict_estimator_sklearn[name_estimator].fit(X, y).coef_
+
+    intercept_sk = dict_estimator_sklearn[name_estimator].intercept_
+    np.testing.assert_allclose(coefs_ours, coefs_sk.flatten(), atol=1e-4)
+    np.testing.assert_allclose(intercept_sk, intercept_ours, atol=1e-4)
+
+
+# Test if skglm multitask solver returns the coefficients
+def test_intercept_mtl():
```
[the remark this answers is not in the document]
mathurinm: Agreed, this file can be deleted and its content merged with the existing tests

skglm/estimators.py
QB3: Is there any reason to prefer `None` over `np.newaxis` ? `np.newaxis` seems more explicit to me https://stackoverflow.com/questions/944863/numpy-should-i-use-newaxis-or-none

skglm/solvers/cd_solver.py
```diff
@@ -31,7 +31,10 @@ def cd_solver_path(X, y, datafit, penalty, alphas=None,
     alphas : ndarray
         List of alphas where to compute the models.
 
-    coef_init : ndarray, shape (n_features,) | None, optional, (default=None)
+    fit_intercept : bool
+        Whether or not to fit an intercept.
+
+    coef_init : ndarray, shape (n_features + 1,) | None, optional, (default=None)
```
mathurinm, line 37: shape

skglm/solvers/cd_solver.py
mathurinm, line 70: shape

skglm/solvers/cd_solver.py
```diff
@@ -268,19 +269,27 @@ def cd_solver(
...
             # 3) do Anderson acceleration on smaller problem
             # TODO optimize computation using ws
-            w_acc[ws], Xw_acc[:], is_extrapolated = accelerator.extrapolate(w[ws], Xw)
+            w_acc[:], Xw_acc[:], is_extrapolated = accelerator.extrapolate(w, Xw)
 
             if is_extrapolated:  # avoid computing p_obj for un-extrapolated w, Xw
                 # TODO : manage penalty.value(w, ws) for weighted Lasso
-                p_obj = datafit.value(y, w, Xw) + penalty.value(w)
-                p_obj_acc = datafit.value(y, w_acc, Xw_acc) + penalty.value(w_acc)
+                p_obj = (datafit.value(y, w[:n_features], Xw) +
```
Badr-MOUFAD, line 289: Perhaps we should consider, in another PR removing the ``w`` argument in ``datafit.value``. It's pretty disturbing as ``.value`` does not use it. Also, in our setup we assume that datafit depends on ``Xw``. WDYT?

skglm/solvers/cd_solver.py
```diff
@@ -268,19 +269,27 @@ def cd_solver(
...
+                _cd_epoch(X, y, w[:n_features], Xw, datafit, penalty, ws)
+
+            # update intercept
+            if fit_intercept:
+                intercept_old = w[-1]
+                w[-1] -= datafit.intercept_update_step(y, Xw)
+                Xw += (w[-1] - intercept_old)
 
             # 3) do Anderson acceleration on smaller problem
             # TODO optimize computation using ws
```
Badr-MOUFAD, line 284: performing acceleration while considering the intercept is not straightforward using the ``AndersonAcceleration`` class.
I disabled it here but marked it as a potential improvement

mathurinm: Are you sure it's not straightforward ? The intercept is the last element of `w` so it should be taken into account easily by the accelerator

skglm/solvers/cd_solver.py
mathurinm: I agree it makes sense at first, but unfortunately the `w` argument is needed for SVC

skglm/solvers/cd_solver.py
Badr-MOUFAD: @mathurinm, I mean taking advantage of working sets (accelerating ``w_ws``) while considering intercept in the AndersonAcceleration.

In this setup, we extrapolate w and intercept.

skglm/estimators.py
```diff
@@ -141,15 +143,24 @@ def _glm_fit(X, y, model, datafit, penalty):
...
         X_, y, datafit_jit, penalty_jit, w, Xw, max_iter=model.max_iter,
         max_epochs=model.max_epochs, p0=model.p0,
-        tol=model.tol,  # ws_strategy=model.ws_strategy,
+        tol=model.tol, fit_intercept=model.fit_intercept,
+        # ws_strategy=model.ws_strategy,
         verbose=model.verbose)
+    model.coef_, model.stop_crit_ = coefs[:n_features], kkt
+    if y.ndim == 1:
+        model.intercept_ = coefs[-1] if model.fit_intercept else 0.
+        # if model.fit_intercept:
```
mathurinm, line 152: rm comments

skglm/estimators.py
mathurinm: You're right, np.newaxis is clearer and there are no room constraints here

skglm/solvers/cd_solver.py
mathurinm: not resolved, no ?

skglm/solvers/cd_solver.py
```diff
@@ -226,32 +221,38 @@ def cd_solver(
     obj_out = []
     all_feats = np.arange(n_features)
     stop_crit = np.inf  # initialize for case n_iter=0
-    w_acc, Xw_acc = np.zeros(n_features), np.zeros(n_samples)
+    w_acc, Xw_acc = np.zeros(n_features+fit_intercept), np.zeros(n_samples)
```
mathurinm, line 224: use spaces around + 

they are not mandatory, but we use them everywhere else. Staying consistent allows more efficient `ctrl F` + replace.

skglm/solvers/cd_solver.py
```diff
@@ -226,32 +221,38 @@ def cd_solver(
...
 
     is_sparse = sparse.issparse(X)
+
+    if len(w) != n_features + fit_intercept:
+        raise ValueError(
+            "The size of weights should be n_features + fit_intercept, \
```
mathurinm, line 230: should weights be coefficients here ?

skglm/solvers/group_bcd_solver.py
```diff
@@ -62,9 +66,13 @@ def group_bcd_solver(X, y, datafit, penalty, w_init=None, Xw_init=None, p0=10,
     n_samples, n_features = X.shape
     n_groups = len(penalty.grp_ptr) - 1
 
-    w = np.zeros(n_features) if w_init is None else w_init
+    w = np.zeros(n_features + fit_intercept) if w_init is None else w_init
     Xw = np.zeros(n_samples) if w_init is None else Xw_init
-    # Xw = X @ w
+    if len(w) != n_features + fit_intercept:
+        raise ValueError(
+            "The size of weights should be n_features + fit_intercept, \
```
mathurinm, line 73: same remark as above

skglm/tests/test_estimators.py
```diff
@@ -97,27 +97,38 @@ def test_check_estimator(estimator_name):
...
+    if fit_intercept:
+        estimator_sk.fit_intercpet = True
```
mathurinm, line 108: There is a typo `intercpet`

That's why we try to avoid such designs, they can fail silently.

skglm/tests/test_group.py
```diff
@@ -132,6 +132,44 @@ def test_vs_celer_grouplasso(n_groups, n_features, shuffle):
...
+def test_intercept_grouplasso(n_groups, n_features, shuffle):
+    n_samples = 100
+    rnd = np.random.RandomState(42)
+    X, y, _ = make_correlated_data(n_samples, n_features, random_state=rnd)
+
+    grp_indices, grp_ptr, groups = _generate_random_grp(n_groups, n_features, shuffle)
+    weights = abs(rnd.randn(n_groups))
+
+    alpha_max = 0.
```
mathurinm, line 145: can we define `_alpha_max_group_lasso` in some utils instead of redefining it in the test ? It feels like it would be used in other parts of the codebase

skglm/tests/test_intercept.py
PABannier: +1. Why do we need an extra `test_intercept.py`? Can't we parametrize all tests in `test_estimator.py` with `fit_intercept`?

skglm/tests/test_group.py
```diff
@@ -132,6 +132,44 @@ def test_vs_celer_grouplasso(n_groups, n_features, shuffle):
...
+    quad_group = compiled_clone(quad_group, to_float32=X.dtype == np.float32)
+    group_penalty = compiled_clone(group_penalty)
+    w = group_bcd_solver(
+        X, y, quad_group, group_penalty, fit_intercept=True, tol=1e-12)[0]
+
+    model = GroupLasso(groups=groups, alpha=alpha, weights=weights,
+                       fit_intercept=True, tol=1e-12)
+    model.fit(X, y)
+
+    np.testing.assert_allclose(model.coef_, w[:X.shape[1]], atol=1e-5)
```
mathurinm, line 169: I'm not 100% sure what this tests. GroupLasso.fit calls group_bcd_solver, so what is expected here ? If group_bcd solver computed the wrong intercept, the test would pass anyway

skglm/tests/test_group.py
```diff
@@ -173,4 +211,4 @@ def test_anderson_acceleration():
 
 
 if __name__ == '__main__':
-    pass
```
mathurinm, line 176: cleanup

skglm/tests/test_intercept.py
mathurinm: agreed, this needs to be addressed

skglm/tests/test_penalties.py
```diff
@@ -61,3 +61,7 @@ def test_subdiff_diff_block(block_penalty):
     ).fit(X, Y)
     # assert the stopping criterion is satisfied
     assert_array_less(est.stop_crit_, est.tol)
+
```
mathurinm, line 64: cleanup

PABannier (commented):
LGTM overall with minor comments on tests as already pointed out. As someone who just discovers the PR and not being very familiar with intercept fitting, I would add a bit of documentation on how it is done in skglm. I understand we're doing intercept steps during the optimization procedure, but this would deserve a few lines of docs IMO.

skglm/datafits/multi_task.py
```diff
@@ -91,3 +91,6 @@ def full_grad_sparse(self, X_data, X_indptr, X_indices, Y, XW):
                     XjTXW[t] += X_data[i] * XW[X_indices[i], t]
             grad[j, :] = (XjTXW - self.XtY[j, :]) / n_samples
         return grad
+
+    def intercept_update_step(self, Y, XW):
```
mathurinm, line 95: use mean to be consistent as @PABannier suggested

skglm/datafits/single_task.py
```diff
@@ -169,6 +172,9 @@ def gradient_scalar_sparse(self, X_data, X_indptr, X_indices, y, Xw, j):
             grad -= X_data[i] * y[idx_i] * sigmoid(- y[idx_i] * Xw[idx_i])
         return grad / len(Xw)
 
+    def intercept_update_step(self, y, Xw):
+        return np.mean((- y * sigmoid(- y * Xw))) / 4
```
mathurinm, line 176: one pair of parentheses can be removed

skglm/estimators.py
```diff
@@ -141,15 +143,24 @@ def _glm_fit(X, y, model, datafit, penalty):
     coefs, p_obj, kkt = solver(
         X_, y, datafit_jit, penalty_jit, w, Xw, max_iter=model.max_iter,
         max_epochs=model.max_epochs, p0=model.p0,
-        tol=model.tol,  # ws_strategy=model.ws_strategy,
+        tol=model.tol, fit_intercept=model.fit_intercept,
+        # ws_strategy=model.ws_strategy,
```
mathurinm, line 147: rm comment

skglm/solvers/cd_solver.py
mathurinm: you could do something like `ws_ = np.hstack([ws, -1]) if fit_intercept else ws` and pass `w[ws_]`


Changes such as the current one potentially introduce a performance regression, that would need to be tested and this takes time. So I'd try and make the above fix work, to not change existing behavior

skglm/tests/test_estimators.py
```diff
@@ -97,27 +97,38 @@ def test_check_estimator(estimator_name):
...
+        estimator_ours.fit_intercpet = True
     estimator_sk.fit(X, y)
     estimator_ours.fit(X, y)
     coef_sk = estimator_sk.coef_
     coef_ours = estimator_ours.coef_
     # assert that something was fitted:
+    if fit_intercept:
+        # add test intercept from fit_intercept
+        estimator_sk.fit_intercpet = False
```
mathurinm, line 117: typo intercpet

this design should be avoided, it is dangerous on many levels (global modification of variable, sensitive to typos such as here)

PABannier: agreed. I don't understand why we need this hack btw. I tried running the tests while commenting these lines and it fails. Why?

skglm/datafits/multi_task.py
Badr-MOUFAD: @mathurinm, mean with axis argument is unfortunately not supported in Numba.
It throws an error at JIT compilation time

Fixed that by using sum over n_samples

skglm/utils.py
```diff
@@ -393,6 +393,18 @@ def check_group_compatible(obj):
             )
 
 
+def alpha_max_group_lasso(X, y, n_groups, grp_indices, grp_ptr, weights):
```
Badr-MOUFAD, line 396: IMO no need to implement a function for alpha_max as so far, we use it only one time.

PABannier: so far yes, but we might use it more than once in the future. So I added it in `utils` as suggested by @mathurinm .

skglm/solvers/cd_solver.py
```diff
@@ -226,32 +221,41 @@ def cd_solver(
...
     all_feats = np.arange(n_features)
     stop_crit = np.inf  # initialize for case n_iter=0
-    w_acc, Xw_acc = np.zeros(n_features), np.zeros(n_samples)
+    w_acc, Xw_acc = np.zeros(n_features + fit_intercept), np.zeros(n_samples)
 
     is_sparse = sparse.issparse(X)
+
+    if len(w) != n_features + fit_intercept:
+        raise ValueError(
+            "The size of coefficients should be n_features + fit_intercept, \
```
mathurinm, line 230: beware, this renders as:
`ValueError: The size of coefficients should be n_features + fit_intercept,                 expected 61, got 60`

skglm/utils.py
mathurinm: yes it's legit. just make it private to avoid maintenance costs

PABannier (approved):
LGTM! For the documentation, i don't feel able to write it for the intercept fitting right now as I don't fully understand what's going on. @Badr-MOUFAD do you want to give it a try? a few lines would suffice.

skglm/datafits/single_task.py
```diff
@@ -169,6 +172,9 @@ def gradient_scalar_sparse(self, X_data, X_indptr, X_indices, y, Xw, j):
             grad -= X_data[i] * y[idx_i] * sigmoid(- y[idx_i] * Xw[idx_i])
         return grad / len(Xw)
 
+    def intercept_update_step(self, y, Xw):
```
Badr-MOUFAD, line 175: We resolved to not implement intercept for logistic regression. So, I think no need for this method.
We might mark it as to do for later PR

skglm/datafits/single_task.py
```diff
@@ -336,10 +342,21 @@ def full_grad_sparse(
...
+                residual = y[X_indices[i]] - Xw[X_indices[i]]
+                if np.abs(residual) < self.delta:
+                    grad_j += - X_data[i] * residual
                 else:
-                    grad_j += - X_data[i] * np.sign(tmp) * self.delta
+                    grad_j += - X_data[i] * np.sign(residual) * self.delta
             grad[j] = grad_j / n_samples
         return grad
+
+    def intercept_update_step(self, y, Xw):
```
Badr-MOUFAD, line 353: Same here as logreg

skglm/solvers/cd_solver.py
```diff
@@ -226,32 +221,50 @@ def cd_solver(
...
     all_feats = np.arange(n_features)
     stop_crit = np.inf  # initialize for case n_iter=0
-    w_acc, Xw_acc = np.zeros(n_features), np.zeros(n_samples)
+    w_acc, Xw_acc = np.zeros(n_features + fit_intercept), np.zeros(n_samples)
 
     is_sparse = sparse.issparse(X)
+
+    if len(w) != n_features + fit_intercept:
+        raise ValueError(
+            "The size of coefficients should be n_features + %i, expected %i, got %i" %
```
Badr-MOUFAD, line 230: You can use an fstring and leverage the fact that ``0 * "some string"`` gives and empty string

```suggestion
            f"The size of coefficients should be n_features {fit_intercept * '+ 1'}"
```

skglm/solvers/cd_solver.py
```diff
@@ -226,32 +221,50 @@ def cd_solver(
...
         if ws_strategy == "subdiff":
-            opt = penalty.subdiff_distance(w, grad, all_feats)
+            opt = penalty.subdiff_distance(w[:n_features], grad, all_feats)
         elif ws_strategy == "fixpoint":
-            opt = dist_fix_point(w, grad, datafit, penalty, all_feats)
-        stop_crit = np.max(opt)
+            opt = dist_fix_point(w[:n_features], grad, datafit, penalty, all_feats)
+
+        if fit_intercept:
+            intercept_opt = np.abs(datafit.intercept_update_step(y, Xw))
```
Badr-MOUFAD, line 249: That's nice indeed @mathurinm!

skglm/solvers/group_bcd_solver.py
```diff
@@ -62,9 +66,13 @@ def group_bcd_solver(X, y, datafit, penalty, w_init=None, Xw_init=None, p0=10,
     n_samples, n_features = X.shape
     n_groups = len(penalty.grp_ptr) - 1
 
-    w = np.zeros(n_features) if w_init is None else w_init
+    w = np.zeros(n_features + fit_intercept) if w_init is None else w_init
     Xw = np.zeros(n_samples) if w_init is None else Xw_init
-    # Xw = X @ w
+    if len(w) != n_features + fit_intercept:
+        raise ValueError(
+            "The size of coefficients should be n_features + fit_intercept, \
```
Badr-MOUFAD, line 73: If you agree, we can go with the solution proposed above in cd solver

skglm/solvers/group_bcd_solver.py
```diff
@@ -4,8 +4,9 @@
 from skglm.utils import AndersonAcceleration, check_group_compatible
 
 
-def group_bcd_solver(X, y, datafit, penalty, w_init=None, Xw_init=None, p0=10,
-                     max_iter=1000, max_epochs=100, tol=1e-4, verbose=False):
+def group_bcd_solver(
```
Badr-MOUFAD, line 7: Is there a reason why we don't check intercept optimality here?

skglm/solvers/multitask_bcd_solver.py
```diff
@@ -212,6 +219,11 @@ def multitask_bcd_solver(
     all_feats = np.arange(n_features)
     stop_crit = np.inf  # initialize for case n_iter=0
 
+    if W.shape[0] != n_features + fit_intercept:
+        raise ValueError(
+            "The size of init coefficients should be n_features + fit_intercept, "
```
Badr-MOUFAD, line 224: ditto, we can go with the solution suggested in cd solver

skglm/tests/test_group.py
Badr-MOUFAD: Agree!
we can test w[-1] against intercept_ when fit_intercept=True?

skglm/utils.py
```diff
@@ -393,6 +393,18 @@ def check_group_compatible(obj):
             )
 
 
+def _alpha_max_group_lasso(X, y, n_groups, grp_indices, grp_ptr, weights):
```
Badr-MOUFAD, line 396: This is used only once, we can remove it, WDYT?

Badr-MOUFAD (commented):
I did a second pass. I think we are almost done.

Some minor comments.

skglm/datafits/single_task.py
mathurinm: it does not hurt to keep it, otherwise we have to redo the work in the follow-up PR

skglm/datafits/single_task.py
mathurinm: this is in Huber datafit so it's currently already used (beware of the diff that gets truncated, this class is not SVC)

skglm/solvers/cd_solver.py
mathurinm: actually I liked n_features + fit_intercept: if I use fit_intercept=True, I may pass a w_init of size n_features, and then the error message does not make it clear that it's because there is an intercept that it should be n_features + 1.

skglm/solvers/cd_solver.py
mathurinm: @Badr-MOUFAD it's one thing that remains to be done IMO. It's already done in multitask_bcd_solver (`ws_ = np.append(ws, -1)`

skglm/solvers/group_bcd_solver.py
mathurinm: I forgot it :) Can you implement it?

skglm/tests/test_group.py
```diff
@@ -106,6 +107,36 @@ def test_vs_celer_grouplasso(n_groups, n_features, shuffle):
     grp_indices, grp_ptr, groups = _generate_random_grp(n_groups, n_features, shuffle)
     weights = abs(rnd.randn(n_groups))
 
+    alpha_max = _alpha_max_group_lasso(X, y, n_groups, grp_indices, grp_ptr, weights)
```
mathurinm, line 110: @PABannier n_groups can be recovered from `grp_ptr.shape[0] - 0` so we can get rid of this parameter to make the function simpler

skglm/utils.py
mathurinm: it's a non trivial function, it could be useful in the future, let's keep it

skglm/solvers/cd_solver.py
```diff
@@ -268,19 +285,27 @@ def cd_solver(
         for epoch in range(max_epochs):
```
Badr-MOUFAD, line 285: @mathurinm, I don't know whether it's worth it to check intercept optimality when solving subproblems. WDYT?

Badr-MOUFAD (approved):

PR #57, opened 2022-08-23
ENH add Square root Lasso

### LINKED ISSUE #53: square-root LASSO implementation
As suggested by @agramfort , I am opening an issue here ([original](https://github.com/scikit-learn/scikit-learn/issues/24196)) to make a feature request for the square-root LASSO. 

### Describe the workflow you want to enable

The unconstrained LASSO is currently implemented in `lasso_path`. However, the parameter for unconstrained LASSO is sensitive to the measurement noise. Instead, the square-root LASSO has objective norm(Ax-b) + alpha * norm(x, 1) and its parameter, alpha, is insensitive to the noise level. I would like an `sklearn` interface (with `fit`, `predict`, etc.) coupled with a `lasso_path`-like function, `srlasso_path`, which solves the square-root LASSO optimization program. 

### Describe your proposed solution

I propose an `sklearn`-friendly wrapper around the `cvxpy` or `cvxopt` interface. More ambitiously, I propose an implementation based on [1] or [2] that implements a fast method for solving square-root LASSO (either via (accelerated) PGD or ADMM).

[1] "On fast convergence of proximal algorithms for SQRT-LASSO optimization: Don’t worry about its nonsmooth loss function" by Li, Xinguo,  Jiang, Haoming, et al. (2020) PMLR. 
[2] "The flare package for high dimensional linear regression and precision matrix estimation in R" by Li, Xingguo, Zhao, Tuo, Yuan, Xiaoming and Liu, Han. (2015) JMLR.

### Describe alternatives you've considered, if relevant

Currently, `statsmodels` offers a square-root LASSO solver implemented via the Python package `cvxopt` - this approach is slow. Using `cvxpy` with another interior point solver like MOSEK offers only 2x speed-up (it is snails compared to `lasso_path`). 

Separately, if the problem is appropriately regularized to avoid over-fitting, some work has shown that it's possible to use accelerated proximal gradient descent to achieve convergence. However, the paper offers an algorithm with several critical implementation details unprovided (referring to [1] above).

### Additional context

Note that the characterization of srLASSO as a second-order conic programming (SOCP) problem is given in [3]

[3] Belloni, A., Chernozhukov, V. and Wang, L. Square-Root LASSO: Pivotal recovery of sparse signals via conic programming https://arxiv.org/pdf/1009.5689.pdf

Badr-MOUFAD:
This proposes a Prox Newton solver for the Square root Lasso,
and potentially closes #53 

So far, the solver:
- gives the same solution as statsmodels up to 5 decimals
- is 10x faster than statsmodels

you can reproduce these results using [example_sqrtlasso](https://github.com/Badr-MOUFAD/skglm/blob/sqrt-lasso/example_sqrtlasso.py)

However, the solver comes with one limitation: can't be fit on problems
where ``y = Xw`` as it violates the assumption of the datafit
differentiability.


mathurinm:
The PN solver seems to work quite well until `lambda` goes too small (probably below `lambda_min` -- value for which the residuals vanish and the sqrt lasso solution is the basis pursuit solution) 

image

It seems we can be up to 100 times faster than statsmodel.

We could create a specific `sqrt_lasso_path` that would interrupt the path computation as soon as it reaches residuals with norm less than `1e-8 * norm(y)`. Then creating a SqrtLasso class should be easy given @Klopfe's work in #54

Badr-MOUFAD:
> It seems we can be up to 100 times faster than ``statsmodels``

Not to mention the additional speedups we would benefit from along a path through warm starts.

I looked up ``statsmodels``. They don't seem to implement a path solver. Though they do have a [``start_params``](https://www.statsmodels.org/dev/generated/statsmodels.regression.linear_model.OLS.fit_regularized.html#:~:text=start_paramsarray_like) in their fit method.

skglm/experimental/sqrt_lasso.py
```diff
@@ -0,0 +1,74 @@
...
+        for i in range(n_alphas):
+            l1_penalty.alpha = alphas[i]
+            coef_init = coefs[:, i-1]
+
+            coef, p_objs_out, stop_crit = prox_newton(
+                X, y, sqrt_quadratic, l1_penalty,
+                w_init=coef_init, max_iter=self.max_iter,
+                max_pn_iter=self.max_pn_iter,
+                tol=self.tol, verbose=self.verbose)
+
```
mathurinm, line 69: here I'd check the residual norm, and if it's too small, I'd stop the path computation while displaying a ConvergenceWarning

skglm/experimental/sqrt_lasso.py
```diff
@@ -0,0 +1,74 @@
...
+        return np.full(n_samples, fill_value)
+
+
+class SqrtLasso(Lasso):
+
+    def __init__(self, alpha=1., max_iter=100, max_pn_iter=1000, p0=10,
+                 tol=1e-4, fit_intercept=True, verbose=0,):
+        super(Lasso, self).__init__(
+            alpha=alpha, tol=tol, max_iter=max_iter,
+            fit_intercept=fit_intercept)
```
mathurinm, line 43: I realize this will be an issue if an intercept is fitted, because (to double check) the centering trick will be performed by sklearn Lasso.fit, but it does not work for the SqrtLasso

skglm/experimental/sqrt_lasso.py
```diff
@@ -0,0 +1,74 @@
...
+
+    def path(self, X, y, alphas=None, **kwargs):
+        n_features = X.shape[1]
+        n_alphas = len(alphas)
+        alphas = np.sort(alphas)[::-1]
+
+        sqrt_quadratic = compiled_clone(SqrtQuadratic())
+        l1_penalty = compiled_clone(L1(1.))
+
+        coefs = np.zeros((n_features, n_alphas+1))
```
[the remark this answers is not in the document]
Badr-MOUFAD: No need for +1, I will change the design

skglm/experimental/sqrt_lasso.py
```diff
@@ -0,0 +1,95 @@
...
+                X, y, sqrt_quadratic, l1_penalty,
+                w_init=coef_init, max_iter=self.max_iter,
+                max_pn_iter=self.max_pn_iter,
+                tol=self.tol, verbose=self.verbose)
+
+            residual = sqrt_quadratic.value(y, coef, X @ coef)
+            if residual < self.tol * scaled_norm_y:
+                warnings.warn(
+                    f"Small residuals will prevent the solver from converging.\n"
+                    f"Considering taking alphas greater than {alphas[i]:.4e}",
```
mathurinm, line 87: Considering > consider

Also display the residual norm, and the threshold you compare it too

skglm/experimental/sqrt_lasso.py
```diff
@@ -0,0 +1,95 @@
...
+            coef_init = coefs[:, i].copy() if i else np.zeros(n_features)
+
+            coef, p_objs_out, stop_crit = prox_newton(
+                X, y, sqrt_quadratic, l1_penalty,
+                w_init=coef_init, max_iter=self.max_iter,
+                max_pn_iter=self.max_pn_iter,
+                tol=self.tol, verbose=self.verbose)
+
+            residual = sqrt_quadratic.value(y, coef, X @ coef)
+            if residual < self.tol * scaled_norm_y:
```
mathurinm, line 84: I'm not sure the threshold should depend on tol

Either use a fixed value instead of tol (1e-2 ?) either make it a parameter of the model

skglm/experimental/tests.py
```diff
@@ -0,0 +1,46 @@
+import numpy as np
```
mathurinm, line 1: for organization: put a test folder in experimental, and a test_sqrt.py file inside

sqrt_class.py
```diff
@@ -0,0 +1,18 @@
...
+
+n_samples, n_features = 100, 200
+rho = 0.01
+
+X, y, _ = make_correlated_data(n_samples, n_features, random_state=0)
+
+alpha_max = norm(X.T @ y, ord=np.inf) / (np.sqrt(n_samples) * norm(y))
+alpha = rho * alpha_max
+
+clf = SqrtLasso(alpha=alpha_max, verbose=2)
```
Badr-MOUFAD, line 16: @mathurinm, you are fitting with ``alpha=alpha_max`` the solution is 0, right?

P.S: pass alpha instead of alpha_max

skglm/experimental/sqrt_lasso.py
Badr-MOUFAD: I agree @mathurinm, we can remove the ``tol``

Aside from that, this warning would fail when fitting with one alpha.
Also when the alphas are too spaced along the path

mathurinm: yes. Ideally we would need a way to stop the computation in the solver itself

sqrt_class.py
mathurinm: thanks, fixed

skglm/experimental/sqrt_lasso.py
```diff
@@ -0,0 +1,117 @@
...
+            # no warm start for the first alpha
+            coef_init = coefs[i].copy() if i else np.zeros(n_features)
+
+            try:
+                coef, p_objs_out, stop_crit = prox_newton(
+                    X, y, sqrt_quadratic, l1_penalty,
+                    w_init=coef_init, max_iter=self.max_iter,
+                    max_pn_iter=self.max_pn_iter,
+                    tol=self.tol, verbose=self.verbose)
+            except:
```
mathurinm, line 101: usually we try to avoid catching all exceptions (the solver could raise another exception that is not the one we want to catch here). For example we don't want to catch an exception is prox_newton was called with a X which is not an array like 
 
To avoid this we could define a `SmallResidualException` class (inheriting from Exception) to raise a very specific exception in that case, a catch only this one.

skglm/experimental/sqrt_lasso.py
```diff
@@ -0,0 +1,117 @@
...
+                residual = sqrt_quadratic.value(y, coef, X @ coef)
+                warnings.warn(
+                    f"Small residuals will prevent the solver from converging.\n"
+                    f"Consider taking alphas greater than {alphas[i]:.4e}\n"
+                    f"value of residual: {residual:.4e}",
+                    ConvergenceWarning
+                )
+
+            coefs[i] = coef
+            stop_criteria[i] = stop_crit
```
mathurinm, line 114: beware, this is not defined in case the exception is raised

skglm/experimental/sqrt_lasso.py
```diff
@@ -0,0 +1,117 @@
...
+
+        Raises:
+        -------
+            Exception
+                if value of residuals is too small (less than ``1e-10``).
+        """
+        minus_residual = Xw - y
+        norm_residuals = norm(minus_residual)
+
+        if norm_residuals < 1e-10:
```
mathurinm, line 44: we should aim at making the threshold homogeneous in y (avoid values that are not relative to the data, when possible)

alo, in the snippet I pushed, it takes 700 PN iterations to reach this warning: it means the threshold is too low.

can you try using `1e-3 * norm(y)` ? Or even 1e-2 ?

skglm/experimental/sqrt_lasso.py
Badr-MOUFAD: It would have been nice if we could implement it that way but unfortunately, Numba doesn't support custom exceptions

Badr-MOUFAD: @mathurinm, see my proposal pls

skglm/experimental/sqrt_lasso.py
Badr-MOUFAD: that way, we need to compute norm y every time we call ``raw_grad``, WDYT?

mathurinm: I think it's still better than having a solver running for 700 iterations before realizing it is stuck
The cost is of the same order of computing the gradient, it's OK: go for the norm recomputation

skglm/experimental/sqrt_lasso.py
```diff
@@ -0,0 +1,193 @@
...
+            raise Exception("SmallResidualException")
+
+        return minus_residual / norm_residuals / np.sqrt(len(y))
```
mathurinm, line 47: a / (b * c) is usually preferred to a / b / c

skglm/experimental/sqrt_lasso.py
```diff
@@ -0,0 +1,193 @@
...
+                    raise
+
+                # save coef despite not converging
+                # coef_init holds a ref to coef
+                coef = coef_init
+                residual = sqrt_quadratic.value(y, coef, X @ coef)
+                warnings.warn(
+                    f"Small residuals will prevent the solver from converging.\n"
+                    f"Consider taking alphas greater than {alphas[i]:.4e}\n"
+                    f"value of residual: {residual:.4e}",
```
mathurinm, line 188: residual norm rather than value of residual (residuals are a vector) 

+ beware, you have the normalization by n_samples in the current residual. Just do `norm(y - X @ coef)`

skglm/experimental/sqrt_lasso.py
```diff
@@ -0,0 +1,193 @@
...
+            try:
+                coef, _, _ = prox_newton(
+                    X, y, sqrt_quadratic, l1_penalty,
+                    w_init=coef_init, max_iter=self.max_iter,
+                    max_pn_iter=self.max_pn_iter,
+                    tol=self.tol, verbose=self.verbose)
+            except Exception as e:
+                # make sure to catch residual error
+                # it's implemented this way as Numba doesn't support custom Exception
+                if not str(e).startswith("SmallResidualException"):
```
mathurinm, line 178: maybe a simpler fix is raising a ValueError ? your call, I'm also Ok-ish with this one

skglm/experimental/tests/test_sqrt_lasso.py
```diff
@@ -0,0 +1,63 @@
+import numpy as np
+from numpy.linalg import norm
+from statsmodels.regression import linear_model
```
mathurinm, line 3: Their will be an error in test collection if statsmodel is not installed, and it's not obvious to the user that statsmodel is a requirement (it's not specified in setup.py -- and we don't want to put it there). 
IMO, the import must go inside the test (with a noqa), inside a try block, and the test should xfail in case of ImportError

example_sqrtlasso.py
```diff
@@ -0,0 +1,100 @@
+import numpy as np
+from numpy.linalg import norm
+import matplotlib.pyplot as plt
+from statsmodels.regression import linear_model
+
+from skglm.penalties import L1
+from skglm.datafits.single_task import SqrtQuadratic
```
QB3, line 7: This line is currently failing for me, I guess this should be replaced by `from skglm.experimental.sqrt_lasso import SqrtLasso`.
Overall it seems that file is redundant with `_plot_sqrt_lasso.py` and could be deleted

skglm/experimental/sqrt_lasso.py
```diff
@@ -0,0 +1,230 @@
...
+
+        Raises
+        ------
+            Exception
+                if value of residuals is too small (less than ``1e-10``).
+        """
+        minus_residual = Xw - y
+        norm_residuals = norm(minus_residual)
+
+        if norm_residuals < 1e-2 * norm(y):
```
QB3, line 44: The value `1e-2` seems conservative, and contrasts with the docstring `if value of residuals is too small (less than ``1e-10``)`

skglm/experimental/sqrt_lasso.py
```diff
@@ -0,0 +1,230 @@
...
+            raise ValueError("SmallResidualException")
+
+        return minus_residual / (norm_residuals * np.sqrt(len(y)))
+
+    def raw_hessian(self, y, Xw):
+        """Upper bound of the hessian w.r.t eigenvalues."""
```
QB3, line 50: Nitpick: I would try to be even more explicit that the upper bound is in the matrix sense, and that this is a diagonal matrix

skglm/experimental/sqrt_lasso.py
```diff
@@ -0,0 +1,230 @@
...
+
+    objs = []
+
+    L = norm(X, ord=2)
+    # take primal and dual stepsizes equal
+    tau = 0.99 / L
+    sigma = 0.99 / L
+
+    for t in range(max_iter):
+        w = ST_vec(w - tau * X.T @ (2 * z - z_old), alpha * np.sqrt(n_samples) * tau)
```
QB3, line 222: This feels weird to see a `np.sqrt(n_samples)` term.
It feels very sqrt-Lasso-dependent, and feels that this will not generalize well to other datafits.

I do not know for this PR, but maybe we could try to isolate the generic Chambolle-Pock solver, and allow to change the default solver (Prox-Newton or Chambolle-Pock) in `SqrtLasso`.

sqrt_class.py
```diff
@@ -0,0 +1,22 @@
+import numpy as np
```
QB3, line 1: Should we remove this file? It seems redundant with `_plot_sqrt_lasso.py`

QB3 (commented):
Thanks you so much for adding the sqrt-Lasso and Chambolle-Pock !
I left a few nitpicks (we might chose to isolate Chambolle-Pock in another PR, depending on the calendar)

example_sqrtlasso.py
mathurinm: I'll delete it indeed

skglm/experimental/sqrt_lasso.py
mathurinm: yep it's conservative, but we see numerical issues otherwise. We could expose this tolerance parameter in the future

skglm/experimental/sqrt_lasso.py
mathurinm: I agree. I have raised this comment already.

sqrt_class.py
mathurinm: definitely !

mathurinm (approved):

mathurinm:
thanks @Badr-MOUFAD

PR #59, opened 2022-08-24
[READY] ENH - Add Gram Solver for single task Quadratic datafit

Badr-MOUFAD:
This aims to add Gram solver that uses gram matrix precomputation to solve problems with quadratic datafit where the number of samples is largely greater than the number of features.

It will proceed as follows:
- [x] implement gram solver for dense data
- [x] extend gram solver to sparse data
- [x] perform benchmarks to show gains compared to ``main`` solver

------------

- [Link to benchmark results](https://badr-moufad.github.io/share-benchmarks/gram_solver/benchmark_lasso_benchopt_run_2022-08-25_09h22m25.html)
- [Link to repo to reproduce](https://github.com/Badr-MOUFAD/benchmark_lasso/tree/gram-bench)


skglm/solvers/gram_cd.py
```diff
@@ -0,0 +1,81 @@
...
+        p_objs_out.append(p_obj)
+    return w, p_objs_out, stop_crit
+
+
+@njit
+def _gram_cd_iter(XtX, Xty, w, XtXw, penalty, opt, ws, n_updates):
+    # inplace update of w, XtXw, opt
+    # perform greedy cd updates
+    for _ in range(n_updates):
+        grad = XtXw - Xty
```
mathurinm, line 71: division by n_samples

OK, I just realised that XtX is X.T @ X / n_samples. That is clever to avoid divisions by n_samples but error prone (imagine someone adding just a few lines to this function in one year, they will likely think that XtX is X.T @ X). 

Can you modify the name ? XtX / n can be `gram` or `hess` for example

skglm/solvers/gram_cd.py
```diff
@@ -0,0 +1,89 @@
...
+    scaled_Xty = X.T @ y / n_samples
+    scaled_y_norm2 = np.linalg.norm(y) ** 2 / (2 * n_samples)
+    all_features = np.arange(n_features)
+    stop_crit = np.inf  # prevent ref before assign
+    p_objs_out = []
+
+    w = np.zeros(n_features)
+    scaled_gram_w = np.zeros(n_features)
+    opt = penalty.subdiff_distance(w, -scaled_Xty, all_features)  # initial: grad = -Xty
+    if use_acc:
```
mathurinm, line 37: when you implement both `cd_strategy="greedy"` and `cd_strategy="cyclic"`, disable use_acc if greedy: it does not work

skglm/solvers/gram_cd.py
```diff
@@ -0,0 +1,117 @@
...
+    if issparse(X):
+        scaled_gram = scaled_gram.toarray()
+
+    all_features = np.arange(n_features)
+    stop_crit = np.inf  # prevent ref before assign
+    p_objs_out = []
+
+    w = np.zeros(n_features) if w_init is None else w_init
+    scaled_gram_w = np.zeros(n_features) if w_init is None else scaled_gram @ w_init
+    opt = penalty.subdiff_distance(w, -scaled_Xty, all_features)  # initial: grad = -Xty
```
mathurinm, line 35: beware that this will not hold if a non zero `w_init` is passed

Badr-MOUFAD: That's a good catch indeed!

skglm/solvers/gram_cd.py
```diff
@@ -0,0 +1,117 @@
...
+                f"stopping crit: {stop_crit:.2e}"
+            )
+
+        if stop_crit <= tol:
+            if verbose:
+                print(f"Stopping criterion max violation: {stop_crit:.2e}")
+            break
+
+        # inplace update of w, XtXw
+        _gram_cd_epoch = _gram_cd_greedy if cd_type == 'greedy' else _gram_cd_cyclic
```
mathurinm, line 59: I think it's better to factorize code by having a single gram_cd function and just changing the selection of `j` inside

skglm/solvers/gram_cd.py
```diff
@@ -0,0 +1,117 @@
...
+        opt = _gram_cd_epoch(scaled_gram, scaled_Xty, w, scaled_gram_w,
+                             penalty, all_features)
```
mathurinm, line 61: I'd disable ws at first so no need for `all_features` here

skglm/solvers/gram_cd.py
mathurinm: or just `greedy_cd=True/False`)

skglm/solvers/gram_cd.py
```diff
@@ -0,0 +1,106 @@
...
+        p_objs_out.append(p_obj)
+    return w, np.array(p_objs_out), stop_crit
+
+
+@njit
+def _gram_cd_epoch(scaled_gram, scaled_Xty, w, scaled_gram_w, penalty, greedy_cd):
+    all_features = np.arange(len(w))
+    for j in all_features:
+        # compute grad
+        grad = scaled_gram_w - scaled_Xty
```
mathurinm, line 86: you don't need to perform this O(p) operation if you use the cyclic rule

skglm/solvers/gram_cd.py
```diff
@@ -0,0 +1,106 @@
...
+        # store p_obj
+        p_obj = 0.5 * w @ scaled_gram_w - scaled_Xty @ w + penalty.value(w)
+        p_objs_out.append(p_obj)
+    return w, np.array(p_objs_out), stop_crit
+
+
+@njit
+def _gram_cd_epoch(scaled_gram, scaled_Xty, w, scaled_gram_w, penalty, greedy_cd):
+    all_features = np.arange(len(w))
+    for j in all_features:
```
mathurinm, line 84:
```suggestion
    for cd_iter in all_features:
```

skglm/solvers/gram_cd.py
```diff
@@ -0,0 +1,106 @@
...
+        # compute grad
+        grad = scaled_gram_w - scaled_Xty
+
+        # select feature j
+        if greedy_cd:
+            opt = penalty.subdiff_distance(w, grad, all_features)
+            chosen_j = np.argmax(opt)
```
mathurinm, line 91:
```suggestion
            j = np.argmax(opt)
```

skglm/solvers/gram_cd.py
```diff
@@ -0,0 +1,106 @@
...
+        else:  # cyclic
+            chosen_j = j
```
mathurinm, line 93:
```suggestion
            j = cd_iter
```

skglm/solvers/gram_cd.py
```diff
@@ -0,0 +1,106 @@
...
+
+        # update w_j
+        old_w_j = w[chosen_j]
```
mathurinm, line 96:
```suggestion
        old_w_j = w[j]
```

mathurinm: etc

skglm/solvers/gram_cd.py
Badr-MOUFAD: @mathurinm, aren't the ``grad`` needed to make the update of w (c.f. L140)?

mathurinm: you do need one partial gradient `grad[chosen_j]` but the rest are not used

Badr-MOUFAD:
I performed a [benchmark ](https://badr-moufad.github.io/share-benchmarks/gram_solver/benchmark_lasso_benchopt_run_2022-08-25_09h22m25.html) and am convinced with the significant speedups the gram solver offers in cases ``n >> p``.

However, I see that we have a huge overhead in the case of sparse datasets, which is not the case for dense data.

<details>
<summary>Sparse case</summary>

sparse-case

</details>


<details>
<summary>Sparse case</summary>

dense-case

</details>

That needs to be investigated further, I am particularly suspecting the sparse matrix multiplication and the conversion to dense as these are what differ the sparse from the dense case.
Meanwhile, any thoughts/ideas, @mathurinm?

skglm/solvers/gram_cd.py
```diff
@@ -0,0 +1,81 @@
...
+    """Run a Gram solver by reformulation the problem as below.
+
+    Minimize::
+        w.T @ Q @ w / (2*n_samples) - b.T @ w / n_samples + penalty(w)
+
+    where::
+        Q = X.T @ X
+        b = X.T @ y
+    """
+    n_samples, n_features = X.shape
```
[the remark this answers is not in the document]
mathurinm: I have written a comment in the docstring

skglm/solvers/gram_cd.py
```diff
@@ -0,0 +1,155 @@
...
+                # omit constant term for comparison
+                p_obj_acc = (0.5 * w_acc @ (scaled_gram @ w_acc) - scaled_Xty @ w_acc +
+                             penalty.value(w_acc))
+                p_obj = 0.5 * w @ (scaled_gram @ w) - scaled_Xty @ w + penalty.value(w)
+                if p_obj_acc < p_obj:
+                    w[:] = w_acc
+                    grad[:] = grad_acc
+
+        # store p_obj
+        p_obj = 0.5 * w @ (scaled_gram @ w) - scaled_Xty @ w + penalty.value(w)
```
mathurinm, line 130: I believe the constant term is missing

Badr-MOUFAD: Indeed, nice catch!

Badr-MOUFAD:
following https://github.com/scikit-learn-contrib/skglm/pull/59#issuecomment-1227029950, we will tackle the big overhead that we have in the sparse case due to the computation of the gram matrix, which particularly impacts the performance of the solver, in a separate issue+PR

mathurinm (approved):
LGTM, merge when green @Badr-MOUFAD

PR #63, opened 2022-08-29
API transform solver functions into classes

mathurinm:


skglm/estimators.py
```diff
@@ -133,16 +133,26 @@ def _glm_fit(X, y, model, datafit, penalty):
                 "The size of the WeightedL1 penalty weights should be n_features, "
                 "expected %i, got %i." % (X_.shape[1], len(penalty.weights)))
 
+    # TODO Handle multi-task case
     if is_classif:
-        solver = cd_solver  # TODO to be be replaced by an instance of BaseSolver
+        if isinstance(datafit, Logistic):
```
mathurinm, line 138: if you do this you still need to pass all arguments to model.

Instead, add a `solver` argument to `_glm_fit`.

Users will instantiate their solver, their penalty and their datafit that they will pass to `GeneralizedLinearEstimator`.

mathurinm: The goal is to avoid potentially long if elif like this if we have 10 solvers.

skglm/estimators.py
```diff
@@ -133,16 +133,26 @@ def _glm_fit(X, y, model, datafit, penalty):
...
-    # do not need to match across solvers
+        solver = AcceleratedCD(
+                fit_intercept=model.fit_intercept, max_iter=model.max_iter,
+                max_epochs=model.max_epochs, p0=model.p0, tol=model.tol,
+                ws_strategy=model.ws_strategy, verbose=model.verbose)
     # TODO QUESTIONS
     # What about ws_strategy?
-    coefs, p_obj, kkt = solver(
+    coefs, p_obj, kkt = solver.solve(
         X_, y, datafit_jit, penalty_jit, w, Xw, max_iter=model.max_iter,
```
mathurinm, line 156: all paremeters other X, y, df, pen, w and Xw should be removed and accessed as `solver.XXX`

otherwise you need a huge signature here that has all arguments of all solvers

skglm/estimators.py
```diff
@@ -133,20 +133,9 @@ def _glm_fit(X, y, model, datafit, penalty):
...
 
-    if is_classif:
-        solver = cd_solver  # TODO to be be replaced by an instance of BaseSolver
-    else:
-        solver = cd_solver if y.ndim == 1 else multitask_bcd_solver
-    # TODO this must be replaced by an instance of BaseSolver being passed
-    # so that arguments are attributes of the `solver` object and arguments
-    # do not need to match across solvers
     # TODO QUESTIONS
     # What about ws_strategy?
```
mathurinm, line 137: this can be removed as well, ws_strategy will be passed to solver

skglm/estimators.py
```diff
@@ -423,7 +422,11 @@ def fit(self, X, y):
         self :
             Fitted estimator.
         """
-        return _glm_fit(X, y, self, Quadratic(), L1(self.alpha))
+        solver = AcceleratedCD(
```
mathurinm, line 425: We want to be more flexible than that: pass `solver` to Lasso and call self.solver here

mathurinm: on second thoughts we want to be able to instantiate estimator with providing the solver. 
A default choice could be solver=None of solver='cd' and instantiate a solver at fit time.
solver=instance of BaseSolver should also be supported

skglm/estimators.py
```diff
@@ -185,35 +172,13 @@ class GeneralizedLinearEstimator(LinearModel):
         Penalty. If None, `penalty` is initialized as a `L1` penalty.
         `penalty` is replaced by a JIT-compiled instance when calling fit.
 
+    solver : instance of BaseSolver, optional
+        Solver. If None, `solver` is initialized as a `AcceleratedCD` solver.
```
mathurinm, line 176:
```suggestion
        Solver. If None, `solver` is initialized as an `AcceleratedCD` solver.
```

skglm/estimators.py
```diff
@@ -229,21 +194,12 @@ class GeneralizedLinearEstimator(LinearModel):
...
-    def __init__(self, datafit=None, penalty=None, is_classif=False, max_iter=100,
-                 max_epochs=50_000, p0=10, tol=1e-4, fit_intercept=True,
-                 warm_start=False, ws_strategy="subdiff", verbose=0):
+    def __init__(self, datafit=None, penalty=None, solver=None, is_classif=False):
         super(GeneralizedLinearEstimator, self).__init__()
         self.is_classif = is_classif
-        self.tol = tol
-        self.max_iter = max_iter
-        self.fit_intercept = fit_intercept
-        self.warm_start = warm_start
```
mathurinm, line 240: warm_start would remain an attribute of GLE for me. WDYT? same for fit_intercept

skglm/estimators.py
```diff
@@ -285,7 +241,9 @@ def fit(self, X, y):
         """
         self.penalty = self.penalty if self.penalty else L1(1.)
         self.datafit = self.datafit if self.datafit else Quadratic()
-        return _glm_fit(X, y, self, self.datafit, self.penalty)
+        self.solver = self.solver if self.solver else AcceleratedCD()
+        return self.solver.solve(X, y, self.datafit, self.penalty)
```
mathurinm, line 245: you need glm_fit here to do all the checks, and if solver is None this will fail so you need a default instantiation like for penalty and datafit

skglm/estimators.py
```diff
@@ -348,30 +306,8 @@ class Lasso(LinearModel, RegressorMixin):
...
-    p0 : int
-        First working set size.
-
-    verbose : bool or int
-        Amount of verbosity.
-
-    tol : float, optional
-        Stopping criterion for the optimization.
-
-    fit_intercept : bool, optional (default=True)
```
mathurinm, line 366: I would keep fit_intercept and warm_start here (same for all estimators)

skglm/estimators.py
```diff
@@ -348,30 +306,8 @@ class Lasso(LinearModel, RegressorMixin):
...
-        Whether or not to fit an intercept.
-
-    warm_start : bool, optional (default=False)
-        When set to True, reuse the solution of the previous call to fit as
-        initialization, otherwise, just erase the previous solution.
-
-    ws_strategy : str
-        The score used to build the working set. Can be ``fixpoint`` or ``subdiff``.
+    solver : instance of BaseSolver, optional
+        Solver. If None, `solver` is initialized as a `AcceleratedCD` solver.
```
mathurinm, line 310: an

skglm/estimators.py
```diff
@@ -423,7 +350,7 @@ def fit(self, X, y):
         self :
             Fitted estimator.
         """
-        return _glm_fit(X, y, self, Quadratic(), L1(self.alpha))
+        return self.solver.solve(X, y, Quadratic(), L1(self.alpha))
```
mathurinm, line 353: need to keep glm_fit

skglm/estimators.py
```diff
@@ -423,7 +350,7 @@ def fit(self, X, y):
...
 
     def path(self, X, y, alphas, coef_init=None, return_n_iter=True, **params):
```
mathurinm, line 355: we want to keep this (for compatibility with sklearn) but I think the `cd_solver_path` import is gone

We should replace it by solver.path here

skglm/estimators.py
PABannier: It depends where we put the boundary between what is similar with sklearn and what is not. `tol` or `max_epochs` are also attributes in sklearn yet they are abstracted away in skglm. One could argue that they are not often tuned while `fit_intercept` might be. I'd adopt the principle of least surprise, so we can expose `fit_intercept`. For `warm_start` no opinion.

PABannier: ok we need `fit_intercept` and `warm_start` for `test_check_estimator` to pass actually. debate closed.

skglm/solvers/gram_cd.py
```diff
@@ -53,85 +43,85 @@ def gram_cd_solver(X, y, penalty, max_iter=100, w_init=None,
...
-        p_objs_out.append(p_obj)
-    return w, np.array(p_objs_out), stop_crit
+    def __init__(self, max_iter=100, use_acc=True, greedy_cd=True, tol=1e-4, verbose=0):
+        self.max_iter = max_iter
+        self.use_acc = use_acc
+        self.greedy_cd = greedy_cd
+        self.tol = tol
+        self.verbose = verbose
+
+    def solve(self, X, y, datafit, penalty, w_init=None, Xw_init=None):
```
mathurinm, line 55:
```suggestion
    def solve(self, X, y, model, datafit, penalty, w_init=None, Xw_init=None):
```

skglm/tests/test_estimators.py
```diff
@@ -241,6 +244,7 @@ def test_grid_search(estimator_name):
     estimator_sk = clone(dict_estimators_sk[estimator_name])
     estimator_ours = clone(dict_estimators_ours[estimator_name])
     estimator_sk.tol = 1e-10
+    # XXX: No need for `tol` anymore as it already is in solver
```
mathurinm, line 247: so should it be `estimator_ours.solver.tol = 1e-10` ?

skglm/estimators.py
```diff
@@ -16,16 +17,14 @@
 from sklearn.preprocessing import LabelEncoder
 from sklearn.multiclass import OneVsRestClassifier, check_classification_targets
 
-
 from skglm.utils import compiled_clone
-from skglm.solvers import cd_solver_path, multitask_bcd_solver_path
```
mathurinm, line 21: we need to keep path for lasso.path (compatibility with sklearn)

mathurinm: and same for multitask I guess

skglm/estimators.py
```diff
@@ -70,7 +69,8 @@ def _glm_fit(X, y, model, datafit, penalty):
         raise ValueError("X and y have inconsistent dimensions (%d != %d)"
                          % (n_samples, y.shape[0]))
 
-    if not model.warm_start or not hasattr(model, "coef_"):
+    # if not model.warm_start or not hasattr(model, "coef_"):
+    if not solver.warm_start or not hasattr(model, "coef_"):
```
mathurinm, line 73: we probably need an example in the doc of fitting a GLE with warm start, as it will be surprising for sklearn users that warm_start is an attribute of the solver rather than of the model

skglm/estimators.py
```diff
@@ -107,23 +107,25 @@ def _glm_fit(X, y, model, datafit, penalty):
...
-        if model.fit_intercept:
+        if fit_intercept:
             w = np.hstack([w, model.intercept_])
-        Xw = X_ @ w[:w.shape[0] - model.fit_intercept] + model.fit_intercept * w[-1]
+        Xw = X_ @ w[:w.shape[0] - fit_intercept] + fit_intercept * w[-1]
     else:
         # TODO this should be solver.get_init() do delegate the work
         if y.ndim == 1:
-            w = np.zeros(n_features + model.fit_intercept, dtype=X_.dtype)
+            w = np.zeros(n_features + solver.fit_intercept, dtype=X_.dtype)
```
mathurinm, line 124:
```suggestion
            w = np.zeros(n_features + fit_intercept, dtype=X_.dtype)
```

skglm/estimators.py
```diff
@@ -107,23 +107,25 @@ def _glm_fit(X, y, model, datafit, penalty):
...
             Xw = np.zeros(n_samples, dtype=X_.dtype)
         else:  # multitask
-            w = np.zeros((n_features + model.fit_intercept, y.shape[1]), dtype=X_.dtype)
+            w = np.zeros((n_features + solver.fit_intercept,
```
mathurinm, line 127:
```suggestion
            w = np.zeros((n_features + fit_intercept,
```

skglm/estimators.py
```diff
@@ -393,19 +342,15 @@ class Lasso(LinearModel, RegressorMixin):
...
+    def __init__(self, alpha=1., tol=1e-4, max_iter=50, max_epochs=50_000,
+                 fit_intercept=True, warm_start=False):
         super().__init__()
+        self.alpha = alpha
         self.tol = tol
         self.max_iter = max_iter
+        self.max_epochs = max_epochs
         self.fit_intercept = fit_intercept
         self.warm_start = warm_start
-        self.verbose = verbose
```
mathurinm, line 404: keep verbose, p0 and other attributes (otherwise they can't be set for this estimator)

skglm/estimators.py
```diff
@@ -423,7 +368,11 @@ def fit(self, X, y):
         self :
             Fitted estimator.
         """
-        return _glm_fit(X, y, self, Quadratic(), L1(self.alpha))
+        # TODO: Add Gram solver
+        _solver = AcceleratedCD(
```
mathurinm, line 372: keep verbose and p0 

also I think `solver` should be an OK name here

skglm/estimators.py
```diff
@@ -464,13 +413,10 @@ def path(self, X, y, alphas, coef_init=None, return_n_iter=True, **params):
...
         penalty = compiled_clone(L1(self.alpha))
         datafit = compiled_clone(Quadratic(), to_float32=X.dtype == np.float32)
-
-        return cd_solver_path(
-            X, y, datafit, penalty, alphas=alphas,
-            coef_init=coef_init, max_iter=self.max_iter,
-            return_n_iter=return_n_iter, max_epochs=self.max_epochs,
-            p0=self.p0, tol=self.tol, verbose=self.verbose,
-            ws_strategy=self.ws_strategy)
+        _solver = AcceleratedCD(
```
mathurinm, line 416: `solver` ? replace everywhere if OK

skglm/estimators.py
```diff
@@ -489,18 +435,6 @@ class WeightedLasso(LinearModel, RegressorMixin):
         Positive weights used in the L1 penalty part of the Lasso
         objective. If None, weights equal to 1 are used.
 
-    max_iter : int, optional
```
mathurinm, line 492: these need to be brought back in

skglm/estimators.py
```diff
@@ -821,8 +722,8 @@ class MCPRegression(LinearModel, RegressorMixin):
         When set to True, reuse the solution of the previous call to fit as
         initialization, otherwise, just erase the previous solution.
 
-    ws_strategy : str
-        The score used to build the working set. Can be ``fixpoint`` or ``subdiff``.
+    solver : instance of BaseSolver
```
mathurinm, line 725: solver not exposed

skglm/estimators.py
```diff
@@ -843,20 +744,16 @@ class MCPRegression(LinearModel, RegressorMixin):
...
+                 fit_intercept=True, warm_start=False):
         super().__init__()
+        self.alpha = alpha
+        self.gamma = gamma
         self.tol = tol
         self.max_iter = max_iter
+        self.max_epochs = max_epochs
         self.fit_intercept = fit_intercept
         self.warm_start = warm_start
-        self.verbose = verbose
```
mathurinm, line 854: add back, pass to sovler in fit

skglm/estimators.py
```diff
@@ -1056,12 +938,11 @@ def path(self, X, y, alphas, coef_init=None, return_n_iter=True, **params):
         """
         penalty = compiled_clone(L1(self.alpha))
         datafit = compiled_clone(Logistic(), to_float32=X.dtype == np.float32)
-
-        return cd_solver_path(
-            X, y, datafit, penalty, alphas=alphas,
-            coef_init=coef_init, max_iter=self.max_iter,
-            return_n_iter=return_n_iter, max_epochs=self.max_epochs,
-            p0=self.p0, tol=self.tol, verbose=self.verbose)
+        _solver = ProxNewton(
```
mathurinm, line 941: I think we can remove LogisticRegression.path. We'll see if somebody complains

skglm/solvers/__init__.py
```diff
@@ -1,2 +1,5 @@
-from .cd_solver import cd_solver_path  # noqa F401
-from .multitask_bcd_solver import multitask_bcd_solver_path  # noqa F401
+from .accelerated_cd import AcceleratedCD  # noqa F401
```
mathurinm, line 1: rather than noqa'ing every line we can define `__all__ = [list of all classes imported in this file]`. I think it's the recommended solution

skglm/tests/test_estimators.py
```diff
@@ -55,7 +56,7 @@
 dict_estimators_sk["wLasso"] = Lasso_sklearn(
     alpha=alpha, tol=tol)
 dict_estimators_ours["wLasso"] = WeightedLasso(
-    alpha=alpha, tol=tol, weights=np.ones(n_features))
```
mathurinm, line 58: revert

skglm/tests/test_estimators.py
```diff
@@ -149,7 +150,7 @@ def test_mtl_path():
             X, Y, l1_ratio=1, tol=1e-14, max_iter=5_000, alphas=alphas
     )[1][:, :X.shape[1]]
     coef_ours = MultiTaskLasso(fit_intercept=fit_intercept, tol=1e-14).path(
-        X, Y, alphas, max_iter=10)[1][:, :X.shape[1]]
```
mathurinm, line 152: revert

mathurinm (commented):
One last choice to discuss: is warm_start an attribute of solver or model ? The more I think about it, the more it's of `model` for me (as solvers already handle this through w and Xw being passed). WDYT @PABannier ?

examples/plot_sparse_recovery.py
```diff
@@ -69,11 +69,13 @@
...
 mse_ref = mean_squared_error(np.zeros_like(y_test), y_test)
 
+solver = AcceleratedCD(ws_strategy="fixpoint", fit_intercept=False)
+
 for idx, estimator in enumerate(penalties.keys()):
     print(f'Running {estimator}...')
-    estimator_path = cd_solver_path(
+    estimator_path = solver.path(
         X, y, compiled_clone(datafit), compiled_clone(penalties[estimator]),
-        alphas=alphas, ws_strategy="fixpoint")
```
mathurinm, line 76: I believe we should keep this line to make the diff as simple as possible

skglm/penalties/__init__.py
```diff
@@ -1,9 +1,14 @@
-from .base import BasePenalty  # noqa F401
```
mathurinm, line 1: sorry, I may have been unclear: there is no particular reason to change this here. It makes the diff longer than it should

The PR is not concerned with penalties and should avoid modifying penalties' init

skglm/solvers/__init__.py
```diff
@@ -1,2 +1,8 @@
-from .cd_solver import cd_solver_path  # noqa F401
-from .multitask_bcd_solver import multitask_bcd_solver_path  # noqa F401
+from .accelerated_cd import AcceleratedCD
```
mathurinm, line 1: here it's ok, let's keep it

examples/plot_sparse_recovery.py
PABannier: path does not accept `ws_strategy`, I pass it when I initialize the solver L72.

skglm/tests/test_estimators.py
```diff
@@ -241,6 +246,7 @@ def test_grid_search(estimator_name):
     estimator_sk = clone(dict_estimators_sk[estimator_name])
     estimator_ours = clone(dict_estimators_ours[estimator_name])
     estimator_sk.tol = 1e-10
+    # XXX: No need for `tol` anymore as it already is in solver
     estimator_ours.tol = 1e-10
     estimator_sk.max_iter = 10_000
```
QB3, line 251: > so should it be estimator_ours.solver.tol = 1e-10 ?

same for `max_iter` ?

try_solver.py
```diff
@@ -0,0 +1,17 @@
+import numpy as np
```
QB3, line 1: Do you have something in mind with this file? Do you want to remove it or convert it into an example?

QB3 (commented):
Outstanding work guys, thanks you so much!

skglm/estimators.py
QB3: +1 for the warm start example, do you want to do it in this PR or in a separated one?

examples/plot_sparse_recovery.py
mathurinm: my mistake, thanks!

skglm/estimators.py
mathurinm: next PR would be OK, this one is big and hanging for a while already

skglm/solvers/prox_newton.py
```diff
@@ -1,34 +1,16 @@
 import numpy as np
 from numba import njit
 from scipy.sparse import issparse
+from skglm.solvers.base import BaseSolver
 
 
 EPS_TOL = 0.3
```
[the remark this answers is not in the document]
mathurinm: They could be, but they are constants and it's ok not to expose them in the API

try_solver.py
mathurinm: it was just to debug, I have removed it

skglm/solvers/base.py
```diff
@@ -0,0 +1,41 @@
...
+            Target values.
+
+        datafit : instance of Datafit class
+            Datafitting term.
+
+        penalty : instance of Penalty class
+            Penalty used in the model.
+
+        w_init : array, shape (n_features,)
+            Coefficient vector.
```
Badr-MOUFAD, line 26:
```suggestion
            Initial coefficient vector. It is used by the solver as an initial guess of the solution.
```

skglm/solvers/base.py
```diff
@@ -0,0 +1,41 @@
...
+
+        Xw_init : array, shape (n_samples,)
+            Model fit.
```
Badr-MOUFAD, line 29: Could we be more precise here? What it stands for and why we require it.

skglm/solvers/gram_cd.py
```diff
@@ -53,85 +43,88 @@ def gram_cd_solver(X, y, penalty, max_iter=100, w_init=None,
...
+                 fit_intercept=True, warm_start=False, verbose=0):
+        self.max_iter = max_iter
+        self.use_acc = use_acc
+        self.greedy_cd = greedy_cd
+        self.tol = tol
+        self.fit_intercept = fit_intercept
+        self.warm_start = warm_start
+        self.verbose = verbose
+
+    def solve(self, X, y, datafit, penalty, w_init=None, Xw_init=None):
```
Badr-MOUFAD, line 58: We don't pass in ``Xw_init``.
IMO, we should hint at this.

skglm/solvers/multitask_bcd.py
```diff
@@ -0,0 +1,428 @@
...
+        self.max_epochs = max_epochs
+        self.p0 = p0
+        self.tol = tol
+        self.use_acc = use_acc
+        self.ws_strategy = ws_strategy
+        self.fit_intercept = fit_intercept
+        self.warm_start = warm_start
+        self.verbose = verbose
+
+    def path(self, X, Y, datafit, penalty, alphas, W_init=None, return_n_iter=False):
```
Badr-MOUFAD, line 25: cosmetic
switch ``path`` and ``solve`` to stay concise with ``anderson_cd``

skglm/solvers/prox_newton.py
```diff
@@ -69,88 +40,101 @@ def prox_newton(X, y, datafit, penalty, w_init=None, p0=10,
...
+        self.max_pn_iter = max_pn_iter
+        self.tol = tol
+        self.fit_intercept = fit_intercept
+        self.warm_start = warm_start
+        self.verbose = verbose
+
+    def solve(self, X, y, datafit, penalty, w_init=None, Xw_init=None):
+        n_samples, n_features = X.shape
+        w = np.zeros(n_features) if w_init is None else w_init
+        Xw = np.zeros(n_samples) if Xw_init is None else X @ w_init
```
Badr-MOUFAD, line 57:
```suggestion
        Xw = np.zeros(n_samples) if Xw_init is None else Xw_init
```

skglm/solvers/base.py
```diff
@@ -0,0 +1,41 @@
+from abc import abstractmethod
+
+
+class BaseSolver():
```
Badr-MOUFAD, line 4: shouldn't we add also a ``path`` method?

Badr-MOUFAD (commented):
Congrats team for the hard work! That's a big nicely done PR!

I did a first pass. Some minor remarks.

skglm/solvers/base.py
PABannier: No we decided not to include it as it does not make sense for non-convex estimators (the path is not unique).

skglm/solvers/anderson_cd.py
```diff
@@ -0,0 +1,345 @@
...
+        for t in range(self.max_iter):
+            if is_sparse:
+                grad = datafit.full_grad_sparse(
+                    X.data, X.indptr, X.indices, y, Xw)
+            else:
+                grad = construct_grad(X, y, w[:n_features], Xw, datafit, all_feats)
+
+            # The intercept is not taken into account in the optimality conditions since
+            # the derivative w.r.t. to the intercept may be very large. It is not likely
+            # to change significantly the optimality conditions.
```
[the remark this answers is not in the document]
QB3: I have to say I am also not completely sure to understand this comment. Does it means that you compute the distance to the subdifferential only for the features (not the intercept), and that the intercept will always be updated?
The way the comment the comment is written seem to say that we are doing some kind of approximation.

QB3:
Apart from the very few comments left, LVGTM!

skglm/solvers/anderson_cd.py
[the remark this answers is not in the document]
mathurinm: Agreed. Let's fix it in another PR, as @PABannier suggested we need to explain better what is done for intercept

skglm/solvers/base.py
mathurinm: makes sense, but the goal of this PR is to move existing code into classes. If we start changing this code, it will be harder to review.

skglm/solvers/base.py
mathurinm: see commend above: I agree on the principle, but I'm not in favor of doing it here

skglm/solvers/anderson_cd.py
```diff
@@ -0,0 +1,345 @@
...
+        self.fit_intercept = fit_intercept
+        self.warm_start = warm_start
+        self.verbose = verbose
+
+    def solve(self, X, y, datafit, penalty, w_init=None, Xw_init=None):
+        if self.ws_strategy not in ("subdiff", "fixpoint"):
+            raise ValueError(
+                'Unsupported value for self.ws_strategy:', self.ws_strategy)
+        n_samples, n_features = X.shape
+        w = np.zeros(n_features) if w_init is None else w_init
```
Badr-MOUFAD, line 65: There is a missing ``self.fit_intercept`` here, right? Though we pass the tests!

mathurinm: Can you add a test that makes this break and then fix the faulty code ?

skglm/solvers/gram_cd.py
PABannier: I stick to the definition in the abstract class, but indeed I don't use it in the function body. WDYT?

Badr-MOUFAD: I added an small comment as hint

skglm/solvers/anderson_cd.py
Badr-MOUFAD: It doesn't break when fitting an estimator since we use ``_glm_fit`` function that passes in a ``w_init``.
However, it breaks when using ``AndersonCD`` solver.
I fixed that. On the other hand, I don't think it deserves to be added as a unitest. WDYT @mathurinm?

Badr-MOUFAD (approved):

mathurinm:
Thanks all !
This can break existing code and we did not use a deprecation cycle, so let's keep an eye out especially for benchopt benchmarks

PR #72, opened 2022-09-02
[READY] More idiomatic imports in `__init__` files

PABannier:
This PR changes the imports in the `__init__` files of the Datafit and Penalty submodules.
As suggested by @mathurinm , instead of appending `noqa F401` at the end of every line, imported objects can be made available in `__all__`.

### DIFF
--- a/skglm/datafits/__init__.py
+++ b/skglm/datafits/__init__.py
@@ -1,7 +1,12 @@
-from .base import BaseDatafit, BaseMultitaskDatafit  # noqa F401
+from .base import BaseDatafit, BaseMultitaskDatafit
+from .single_task import Quadratic, QuadraticSVC, Logistic, Huber
+from .multi_task import QuadraticMultiTask
+from .group import QuadraticGroup
 
-from .single_task import Quadratic, QuadraticSVC, Logistic, Huber  # noqa F401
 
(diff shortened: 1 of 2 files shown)


Badr-MOUFAD (commented):
@PABannier, looks tempting! Yet, I think by doing so, we introduce some redundancy as we should always keep the ``__all__`` synchronized with the imports.

I am not really in favor of this change. IMO, So far the ``#noqa`` doesn't hurt us much.

PABannier:
@mathurinm WDYT? I opened the PR to make it consistent with the changes made in #63 for the `solver` submodule.

QB3:
It is true that we are currently introducing it in the `solver` PR https://github.com/mathurinm/skglm/blob/0cee34a3b8d7631406ba5158f30d6a66f16ab0a3/skglm/solvers/__init__.py#L9

> Yet, I think by doing so, we introduce some redundancy as we should always keep the __all__ synchronized with the imports.

I am not sure to understand what you mean here, what is the potential problem?

PABannier:
So if we all agree that the maintenance cost (to keep the imports and `__all__` synchronized) is minimal, can we merge the PR? I need 2 approvals.

Badr-MOUFAD (approved):

mathurinm:
Thanks PA!

PR #77, opened 2022-10-03
ENH - fit intercept in ``ProxNewton`` solver

### LINKED ISSUE #68: ENH fit intercept in prox newton solver

Badr-MOUFAD:
This adds unpenalized intercept support to ``ProxNewton`` solver and closes #68.
It proceeds as follows
- [x] implement fit intercept
- [x] add unitest
- [x] perform benchmarks

-------------------
[View benchmark results](https://badr-moufad.github.io/share-benchmarks/pn-intercept/benchmark_logreg_l1_benchopt_run_2022-10-04_08h33m33.html)
[Link to benchmark repository](https://github.com/Badr-MOUFAD/benchmark_logreg_l1/tree/bench-intercept)

### DIFF
--- a/skglm/solvers/prox_newton.py
+++ b/skglm/solvers/prox_newton.py
@@ -152,8 +152,8 @@ def _descent_direction(X, y, w_epoch, Xw_epoch, grad_ws, datafit,
     for idx, j in enumerate(ws):
         lipschitz[idx] = raw_hess @ X[:, j] ** 2
 
-    # for a less costly stopping criterion, we do no compute the exact gradient,
-    # but store each coordinate-wise gradient every time we upate one coordinate:
+    # for a less costly stopping criterion, we do not compute the exact gradient,
+    # but store each coordinate-wise gradient every time we update one coordinate:
     past_grads = np.zeros(len(ws))
     X_delta_w_ws = np.zeros(X.shape[0])
     w_ws = w_epoch[ws]
--- a/skglm/solvers/prox_newton.py
+++ b/skglm/solvers/prox_newton.py
@@ -90,7 +90,7 @@ class ProxNewton(BaseSolver):
 
             # optimality of intercept
             if fit_intercept:
-                intercept_opt = np.absolute(np.sum(datafit.raw_grad(y, Xw)))
+                intercept_opt = np.abs(np.sum(datafit.raw_grad(y, Xw)))
             else:
                 intercept_opt = 0.
 
--- a/skglm/solvers/prox_newton.py
+++ b/skglm/solvers/prox_newton.py
@@ -24,6 +24,9 @@ class ProxNewton(BaseSolver):
     tol : float, default 1e-4
         Tolerance for convergence.
 
+    fit_intercept : bool, default True
+        If ``True``, fits an unpenalized intercept.
+
     verbose : bool, default False
         Amount of verbosity. 0/False is silent.
 
@@ -63,6 +66,17 @@ class ProxNewton(BaseSolver):
         is_sparse = issparse(X)
         if is_sparse:
             X_bundles = (X.data, X.indptr, X.indices)
+
+        if len(w) != n_features + self.fit_intercept:
+            if self.fit_intercept:
+                val_error_message = (
+                    "w should be of size n_features + 1 when using fit_intercept=True: "
+                    f"expected {n_features + 1}, got {len(w)}.")
+            else:
+                val_error_message = (
+                    "w should be of size n_features: "
+                    f"expected {n_features}, got {len(w)}.")
+            raise ValueError(val_error_message)
 
         for t in range(self.max_iter):
             # compute scores
--- a/skglm/solvers/prox_newton.py
+++ b/skglm/solvers/prox_newton.py
@@ -160,7 +160,7 @@ class ProxNewton(BaseSolver):
         return w, np.asarray(p_objs_out), stop_crit
 
 
-# @njit
+@njit
 def _descent_direction(X, y, w_epoch, Xw_epoch, fit_intercept, grad_ws, datafit,
                        penalty, ws, tol):
     # Given:
@@ -295,7 +295,7 @@ def _descent_direction_s(X_data, X_indptr, X_indices, y, w_epoch,
     return w_ws - w_epoch[ws_intercept], X_delta_w_ws
 
 
-# @njit
+@njit
 def _backtrack_line_search(X, y, w, Xw, fit_intercept, datafit, penalty, delta_w_ws,
                            X_delta_w_ws, ws):
     # 1) find step in [0, 1] such that:
--- a/skglm/solvers/prox_newton.py
+++ b/skglm/solvers/prox_newton.py
@@ -76,7 +76,7 @@ class ProxNewton(BaseSolver):
 
             # optimality of intercept
             if fit_intercept:
-                intercept_opt = np.abs(datafit.intercept_update_step(y, Xw))
+                intercept_opt = np.absolute(np.sum(datafit.raw_grad(y, Xw)))
             else:
                 intercept_opt = 0.
 
@@ -108,7 +108,7 @@ class ProxNewton(BaseSolver):
                 # find descent direction
                 if is_sparse:
                     delta_w_ws, X_delta_w_ws = _descent_direction_s(
-                        *X_bundles, y, w, Xw, grad_ws, datafit,
+                        *X_bundles, y, w, Xw, fit_intercept, grad_ws, datafit,
                         penalty, ws, tol=EPS_TOL*tol_in)
                 else:
                     delta_w_ws, X_delta_w_ws = _descent_direction(
@@ -118,8 +118,8 @@ class ProxNewton(BaseSolver):
                 # backtracking line search with inplace update of w, Xw
                 if is_sparse:
                     grad_ws[:] = _backtrack_line_search_s(
-                        *X_bundles, y, w, Xw, datafit, penalty, delta_w_ws,
-                        X_delta_w_ws, ws)
+                        *X_bundles, y, w, Xw, fit_intercept, datafit, penalty,
+                        delta_w_ws, X_delta_w_ws, ws)
                 else:
                     grad_ws[:] = _backtrack_line_search(
                         X, y, w, Xw, fit_intercept, datafit, penalty,
@@ -146,7 +146,7 @@ class ProxNewton(BaseSolver):
         return w, np.asarray(p_objs_out), stop_crit
 
 
-@njit
+# @njit
 def _descent_direction(X, y, w_epoch, Xw_epoch, fit_intercept, grad_ws, datafit,
                        penalty, ws, tol):
     # Given:
@@ -162,14 +162,14 @@ def _descent_direction(X, y, w_epoch, Xw_epoch, fit_intercept, grad_ws, datafit,
         lipschitz[idx] = raw_hess @ X[:, j] ** 2
 
     # for a less costly stopping criterion, we do not compute the exact gradient,
-    # but store each coordinate-wise gradient every time we update one coordinate:
+    # but store each coordinate-wise gradient every time we update one coordinate
     past_grads = np.zeros(len(ws))
     X_delta_w_ws = np.zeros(X.shape[0])
     ws_intercept = np.append(ws, -1) if fit_intercept else ws
     w_ws = w_epoch[ws_intercept]
 
     if fit_intercept:
-        lipschitz_intercept = np.mean(raw_hess)
+        lipschitz_intercept = np.sum(raw_hess)
         grad_intercept = np.sum(datafit.raw_grad(y, Xw_epoch))
 
     for cd_iter in range(MAX_CD_ITER):
@@ -189,7 +189,7 @@ def _descent_direction(X, y, w_epoch, Xw_epoch, fit_intercept, grad_ws, datafit,
                 X_delta_w_ws += (w_ws[idx] - old_w_idx) * X[:, j]
 
         if fit_intercept:
-            past_grads_intercept = grad_intercept + np.sum(raw_hess * X_delta_w_ws)
+            past_grads_intercept = grad_intercept + raw_hess @ X_delta_w_ws
             old_intercept = w_ws[-1]
             w_ws[-1] -= past_grads_intercept / lipschitz_intercept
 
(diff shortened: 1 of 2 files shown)


skglm/solvers/prox_newton.py
```diff
@@ -42,7 +42,7 @@ class ProxNewton(BaseSolver):
     """
 
     def __init__(self, p0=10, max_iter=20, max_pn_iter=1000, tol=1e-4,
-                 fit_intercept=True, warm_start=False, verbose=0):
+                 fit_intercept=False, warm_start=False, verbose=0):
```
mathurinm, line 45: changing default has a silent impact on users, we try to avoid it as much as possible (though our userbase may not be so large, let's take good habits from the start)

skglm/tests/test_prox_newton.py
```diff
@@ -20,13 +20,13 @@ def test_alpha_max(X_density):
```
mathurinm, line 20: alpha_max changes when fitting intercept

skglm/tests/test_prox_newton.py
```diff
@@ -37,14 +37,17 @@ def test_pn_vs_sklearn(rho, X_density):
     alpha = rho * alpha_max
 
     sk_log_reg = LogisticRegression(penalty='l1', C=1/(n_samples * alpha),
-                                    fit_intercept=False, tol=1e-9, solver='liblinear')
+                                    fit_intercept=fit_intercept,
```
mathurinm, line 40: liblinear penalizes intercept IIRC. Does it work fine here?

skglm/tests/test_prox_newton.py
```diff
@@ -37,14 +37,17 @@ def test_pn_vs_sklearn(rho, X_density):
...
 
     log_datafit = compiled_clone(Logistic())
     l1_penalty = compiled_clone(L1(alpha))
-    w = ProxNewton(tol=1e-9).solve(X, y, log_datafit, l1_penalty)[0]
+    prox_solver = ProxNewton(fit_intercept=fit_intercept, tol=1e-9)
+    w = prox_solver.solve(X, y, log_datafit, l1_penalty)[0]
 
-    np.testing.assert_allclose(w, sk_log_reg.coef_.flatten(), rtol=1e-6, atol=1e-6)
+    np.testing.assert_allclose(w[:n_features], sk_log_reg.coef_.flatten(),
+                               rtol=1e-6, atol=1e-6)
```
mathurinm, line 50: if fit intercept we can also check that the intercepts are non zero and equal

skglm/solvers/prox_newton.py
Badr-MOUFAD: Yes, I totally agree with you. I will revert it.
(I made such a change to pass tests)

mathurinm: perfect :)

skglm/tests/test_prox_newton.py
Badr-MOUFAD: It is ``alpha_max`` for the penalized coefficients. Hence adding an unpenalized intercept doesn't change it.

If I understand what you mean: the condition to have a ``0`` intercept translates to having a centered ``y``.

skglm/tests/test_prox_newton.py
Badr-MOUFAD: No it doesn't work here. The test against ``sklearn`` fails.
I was actually scratching my head to figure out why. Thanks @mathurinm for reminding me that liblinear fits an unpenalized intercept.

Any ideas against whom should we uni test?

skglm/tests/test_prox_newton.py
mathurinm: That would be the case for the Lasso, but for Logreg things are different. 
My point is, you compute the alpha_max without intercept, then fit with an intercept, so without more computations, it's not clear what should happen (should we expect 0 as a solution or not?)

skglm/tests/test_prox_newton.py
mathurinm: as a first try i'd test `intercept_scaling=1000`  in sklearn logreg it it doesn't break convergence (lower value else)

QB3: > liblinear fits an unpenalized intercept

Did you mean 'liblinear fits a penalized intercept' ? https://github.com/scikit-learn/scikit-learn/issues/10626

QB3: > Any ideas against whom should we uni test?

Could we compare to our own cd solvers with intercept ?

mathurinm: yes penalized, my bad
just try a high intercept scaling first to see if you get a value within reasonable tolerance of our unpenalized intercept

Badr-MOUFAD: I checked against [``scikit-learn:saga``](https://scikit-learn.org/stable/modules/linear_model.html#logistic-regression:~:text=The%20following%20table%20summarizes%20the%20penalties%20supported%20by%20each%20solver%3A) as it fits unpenalized intercept

skglm/tests/test_prox_newton.py
Badr-MOUFAD: I 100% agree with you! Yet, in terms of ``alpha_max``, Lasso resemble slightly LogReg as $\nabla F(0) = \frac{y}{2 n}$

$$ -\mathbf{X}^\top \nabla F(0) \in \lambda \partial \lVert \cdot \rVert_1 \quad \mathbb{1}_n^\top \nabla F(0) = 0$$

$$   \lVert \frac{\mathbf{X}^\top y}{2 n}  \rVert_\infty \leq \lambda, \quad \bar{y} = 0$$

I have added a comment in the unittest to clarify that.

Badr-MOUFAD:
here are the [benchmark results](https://badr-moufad.github.io/share-benchmarks/pn-intercept/benchmark_logreg_l1_benchopt_run_2022-10-04_08h33m33.html) and the [link to repo](https://github.com/Badr-MOUFAD/benchmark_logreg_l1/tree/bench-intercept) to reproduce.

I have one concern: I cannot explain why we don't have the same initial objective value with Blitz in the case of ``alpha=alpha_max`` (``reg=1``)

any thoughts?

mathurinm:
I'd test the output (intercept and coef_) outside of benchopt for these two solvers on the problematic data. Check that both get 0 (but I don't think they should, cf our discussion on alpha_max with intercept)

skglm/tests/test_prox_newton.py
Badr-MOUFAD: I have re-checked. There are some subtleties in the intercept case.
I resolved to exclude ``fit_intercept=True`` from the ``alpha_max`` unit test.

What do you think @mathurinm?

Badr-MOUFAD:
> I'd test the output (intercept and coef_) outside of benchopt for these two solvers on the problematic data. Check that both get 0 (but I don't think they should, cf our discussion on alpha_max with intercept)

The reason behind this is: Blitz performs an intercept update before entering the ``iter`` loop.
cf. https://github.com/tbjohns/BlitzL1/blob/8c1d2f6251307021ff4c8e2027df2d92b10d7f63/src/solver.cpp#L442-L443

So nothing surprising in such behavior.

skglm/tests/test_prox_newton.py
Badr-MOUFAD: I don't think it's a pathological case as ``alpha_max`` changes when fitting an unpenalized intercept.

I can show that if we set ``alpha_max = norm(X.T @ y, ord=np.inf) / n_samples`` (remove 2 from denominator),
we would have ``w[:n_features] == 0`` in both cases (``fit_intercept`` ``True`` or ``False``), 

However, I don't think it's a clever idea to refactor this unittest to handle ``fit_intercept=True`` and write a long math justification to explain it. Indeed, we have a second unittest that checks thoroughly the case ``fit_intercept=True``

mathurinm: I agree, it is not worth the complication. Tests should make our life easier not complicate the codebase :)

skglm/tests/test_prox_newton.py
```diff
@@ -9,8 +9,8 @@
 from skglm.utils import make_correlated_data, compiled_clone
 
 
-@pytest.mark.parametrize('X_density', [1, 0.5])
-def test_alpha_max(X_density):
+@pytest.mark.parametrize('X_density, fit_intercept', product([1, 0.5], [False]))
+def test_alpha_max(X_density, fit_intercept):
```
mathurinm, line 13: remove fit_intercept from this test

skglm/solvers/prox_newton.py
```diff
@@ -63,20 +67,37 @@ def solve(self, X, y, datafit, penalty, w_init=None, Xw_init=None):
...
             else:
-                grad = _construct_grad(X, y, w, Xw, datafit, all_features)
+                grad = _construct_grad(X, y, w[:n_features], Xw, datafit, all_features)
 
-            opt = penalty.subdiff_distance(w, grad, all_features)
+            opt = penalty.subdiff_distance(w[:n_features], grad, all_features)
+
+            # optimality of intercept
+            if fit_intercept:
+                intercept_opt = np.absolute(np.sum(datafit.raw_grad(y, Xw)))
```
PABannier, line 93:
```suggestion
                intercept_opt = np.abs(np.sum(datafit.raw_grad(y, Xw)))
```

PABannier (approved):
LGTM! Really good work @Badr-MOUFAD thanks!

skglm/solvers/prox_newton.py
```diff
@@ -24,6 +24,9 @@ class ProxNewton(BaseSolver):
     tol : float, default 1e-4
         Tolerance for convergence.
 
+    fit_intercept : bool, default True
```
mathurinm, line 27: My experience with sklearn is that it's quite confusing to fit an intercept by default. Should it be set to False here ?

skglm/solvers/prox_newton.py
```diff
@@ -63,20 +67,37 @@ def solve(self, X, y, datafit, penalty, w_init=None, Xw_init=None):
...
             else:
-                grad = _construct_grad(X, y, w, Xw, datafit, all_features)
+                grad = _construct_grad(X, y, w[:n_features], Xw, datafit, all_features)
 
-            opt = penalty.subdiff_distance(w, grad, all_features)
+            opt = penalty.subdiff_distance(w[:n_features], grad, all_features)
+
+            # optimality of intercept
+            if fit_intercept:
+                intercept_opt = np.abs(np.sum(datafit.raw_grad(y, Xw)))
```
mathurinm, line 93: To be sure I understand correctly : summing is like taking dot product with np.ones(n_features), so that's the magnitude of the gradient with respect to a constant feature here ?

Can you add a comment ("gradient wrt intercept" or "gradient with respect to constant feature of 1s") ?

skglm/solvers/prox_newton.py
```diff
@@ -174,22 +202,35 @@ def _descent_direction(X, y, w_epoch, Xw_epoch, grad_ws, datafit,
...
+
+            if stop_crit <= tol:
                 break
 
     # descent direction
-    return w_ws - w_epoch[ws], X_delta_w_ws
+    return w_ws - w_epoch[ws_intercept], X_delta_w_ws
 
 
 # sparse version of _compute_descent_direction
```
mathurinm, line 230:
```suggestion
# sparse version of _descent_direction
```

skglm/tests/test_prox_newton.py
```diff
@@ -20,14 +20,15 @@ def test_alpha_max(X_density):
...
+    w = ProxNewton(fit_intercept=False).solve(X, y, log_datafit, l1_penalty)[0]
 
     np.testing.assert_equal(w, 0)
 
 
-@pytest.mark.parametrize("rho, X_density", product([1e-1, 1e-2], [1., 0.5]))
-def test_pn_vs_sklearn(rho, X_density):
-    n_samples, n_features = 11, 19
+@pytest.mark.parametrize("rho, X_density, fit_intercept",
+                         product([1e-1, 1e-2], [1., 0.5], [True, False]))
```
mathurinm, line 29: @Badr-MOUFAD is there a particular thing to test rho=1 ? For me if it works for rho=0.1 it works for rho=1. We don't test anything particular for rho=1.

If not, it can be removed (the test will be simpler and have the same coverage)

skglm/tests/test_prox_newton.py
```diff
@@ -37,14 +38,19 @@ def test_pn_vs_sklearn(rho, X_density):
...
     l1_penalty = compiled_clone(L1(alpha))
-    w = ProxNewton(tol=1e-9).solve(X, y, log_datafit, l1_penalty)[0]
+    prox_solver = ProxNewton(fit_intercept=fit_intercept, tol=1e-9)
+    w = prox_solver.solve(X, y, log_datafit, l1_penalty)[0]
 
-    np.testing.assert_allclose(w, sk_log_reg.coef_.flatten(), rtol=1e-6, atol=1e-6)
+    np.testing.assert_allclose(w[:n_features], sk_log_reg.coef_.flatten(),
+                               rtol=1e-4, atol=1e-3)
+    if fit_intercept:
+        np.testing.assert_allclose(w[-1], sk_log_reg.intercept_, rtol=1e-4, atol=1e-3)
```
mathurinm, line 53: @Badr-MOUFAD use atol and rtol only when needed. The tests seem to pass even when they take their default values here. If not, consider lowering the tol.

atol=1e-3 may be big in some settings, so better avoid it.

skglm/solvers/prox_newton.py
Badr-MOUFAD: agree!

skglm/tests/test_prox_newton.py
Badr-MOUFAD: I have no strong opinion.
It's just a sanity check that we've used to implement.

mathurinm: Then I have a strong opinion :) 
It's a sanity check when we use benchopt, to check that the objective doesn't move. Here, we don't check that, so this parameter can be removed to make the test simpler. Just use a hardcoded rho=0.1

skglm/tests/test_prox_newton.py
mathurinm: spoke too fast, some tests seem to break. Can you check if lowering the tolerance of solvers allows us to remove the rtol and atol ? Or just to use atol=1e-9 ?

mathurinm (approved):

mathurinm:
Thanks @Badr-MOUFAD !

PR #78, opened 2022-10-03
ENH Add Poisson datafit

### LINKED ISSUE #62: ENH add Poisson datafit

PABannier:
Closes #62


PABannier:
Note to self: Poisson datafit is not Lipschitz, need for a line search

Badr-MOUFAD (commented):
You can instead use PN solver as it doesn't require datafits to be Lipschitz and performs a backtracking line search to find a suitable step.
We should implement two functions ``raw_gradient`` and ``raw_hessian`` as in the ``Logisitc`` datafit

https://github.com/scikit-learn-contrib/skglm/blob/19a5a6f52200cf3eca7eb6f5e3f342d50ead39c3/skglm/datafits/single_task.py#L132-L139

Did I get what you meant @PABannier?

mathurinm:
Casting `y` to positive values, I get convergence for all designs I tried (even 5000 x 3000). I think @PABannier you encountered issues because for negative ys, the loss is no longer convex.

It remains to add an example, test against `sklearn.linear_model.PoissonRegressor` and we are good to go on my side

test_poisson.py
```diff
@@ -0,0 +1,21 @@
+import numpy as np
```
Badr-MOUFAD, line 1: ``statsmodels`` provides $\ell_1$ regulized  [Poisson regressor](https://www.statsmodels.org/dev/generated/statsmodels.discrete.discrete_model.Poisson.fit_regularized.html#statsmodels.discrete.discrete_model.Poisson.fit_regularized).
We better unittest against them.

mathurinm:
Good idea ! But we don't want statsmodel as a dependency for the main code, so it will need to be a test dependency only 

I thought sklearn supported regularization, but it's only L2 apparently. 

To discuss: it we match sklearn on Poisson + L2, is it enough or do we need to add a statsmodel test requirement to test Poisson + L1 ? @QB3 @Klopfe @PABannier thoughts on this tradeoff ?

PABannier:
I'd opt for comparing l1-regularized models. I've added `statsmodels` as a test dependency.

PABannier:
For the example, do you think about something in particular @mathurinm ?

.github/workflows/main.yml
```diff
@@ -12,17 +12,18 @@ jobs:
...
-    - name: Install dependencies
-      run: |
-        python -m pip install --upgrade pip
-        python -m pip install git+https://github.com/mathurinm/celer.git
-        pip install pytest
-        pip install numpydoc
-        pip install .
-    - name: Test with pytest
-      run: pytest -v skglm/
+      - uses: actions/checkout@v2
```
mathurinm, line 15: @PABannier can you revert this big chunk of changes ? Normally there's just one line that changed

skglm/tests/test_datafits.py
```diff
@@ -3,12 +3,13 @@
 
 from sklearn.linear_model import HuberRegressor
 from numpy.testing import assert_allclose, assert_array_less
+from statsmodels.discrete.discrete_model import Poisson as PoissonRegressor
```
mathurinm, line 6: See discussion here: https://github.com/scikit-learn-contrib/skglm/pull/57/files#r990803934

we need a more advanced scheme to avoid imposing statsmodel for every user, yet to make it clear that it's a test dependency

mathurinm (changes requested):
PR looks great, once again a nice show of force from our design and solvers: it costs 20 lines to add a whole new model !!!

We only need to be a bit more careful with the statsmodel dependecy and that will be a green light for me

mathurinm:
> For the example, do you think about something in particular @mathurinm ?

It can be anything really, but to quote the great man "if it's not in the doc, it does not exist". Maybe check out what sklearn has ?

skglm/tests/test_datafits.py
PABannier: i've changed it

skglm/tests/test_datafits.py
```diff
@@ -56,5 +56,37 @@ def test_log_datafit():
     np.testing.assert_almost_equal(-grad * (y + n_samples * grad), hess)
 
 
+def test_poisson():
+    try:
+        from statsmodels.discrete.discrete_model import Poisson as PoissonRegressor  # noqa
+    except ImportError:
+        pytest.xfail()
```
mathurinm, line 63: Great! 
Nitpick: can you add an error message?

mathurinm (approved):

mathurinm:
Thanks @PABannier

PR #87, opened 2022-10-11
ENH Add IterativeRewieghtedL1

PABannier:
Implements a (POC) Reweighted Lasso in the spirit of https://web.stanford.edu/~boyd/papers/pdf/rwl1.pdf


skglm/experimental/reweighted.py
```diff
@@ -0,0 +1,44 @@
+import numpy as np
+from numpy.linalg import norm
+from skglm.estimators import Lasso
+
+
+class ReweightedLasso(Lasso):
+    def __init__(self, alpha=1., n_reweight=5, reweight_penalty=None, *args, **kwargs):
+        super().__init__(alpha=alpha, *args, **kwargs)
+        self.n_reweight = n_reweight
```
mathurinm, line 9: should we go for n_reweights ?

skglm/experimental/reweighted.py
```diff
@@ -0,0 +1,44 @@
...
+        self.reweight_penalty = reweight_penalty or self._reweight_penalty
```
mathurinm, line 10: in sklearn estimator's __init__ we're only allowed to do self.foo = foo, nothing more. All other stuff must happen at fit time

skglm/experimental/reweighted.py
```diff
@@ -0,0 +1,44 @@
...
+        self.loss_history_ = []
+        n_samples, n_features = X.shape
+        weights = np.ones(n_features)
+        # XXX: dot product X @ w is slow in high-dimension, to be improved
+        objective = (lambda w: np.sum((y - X @ w) ** 2) / (2 * n_samples)
+                     + self.alpha * np.sqrt(norm(w)))
+
+        for l in range(self.n_reweight):
+            # trick: rescaling the weights (XXX: sparse X would become dense?)
+            scaled_X = X / weights
```
mathurinm, line 23: use weightedL1 penalty instead :)

skglm/experimental/reweighted.py
```diff
@@ -0,0 +1,44 @@
...
+
+            if self.verbose:
+                print("#" * 10)
+                print(f"[REWEIGHT] iteration {l} :: loss {loss}")
+                print("#" * 10)
+
+        self.coef_ = scaled_coef
+
+    @staticmethod
+    def _reweight_penalty(coef):
```
mathurinm, line 41: can this be an attrbiute of the method? So that the reweighting scheme can be arbitrary

toy_reweight.py
```diff
@@ -0,0 +1,22 @@
...
+n_samples, n_features = 200, 500
+X, y, _ = make_correlated_data(n_samples=n_samples, n_features=n_features, random_state=24)
+alpha_max = norm(X.T @ y, ord=np.inf) / n_samples
+
+alpha = alpha_max / 10
+
+clf = ReweightedLasso(alpha=alpha, verbose=2, tol=1e-10)
+clf.fit(X, y)
+
+# reweighting can't increase the L2,0.5 objective, we check that
```
mathurinm, line 16: that's just the L05 objective, right?

PABannier: yes my bad (tests were extracted from Reweighted MTL)

skglm/experimental/reweighted.py
agramfort: It’s better to parametrize by the inverse of the weights to avoid dividing by zero. Besides once a weight is zero it’s a lot more effective to run the next iteration only on features with a non zero weight. Then you don’t need the eps in the weight computation

mathurinm: we have a WeightedL1 penalty so we don't need to divide X 
feature with infinite weights will be ignored by the working set rule

mathurinm: the code above does'nt work if you want to solve MCP with reweighting as its derivative vanishes for values greater than alpha * gamma. That's why we have introduced WeightedL1

skglm/experimental/reweighted.py
```diff
@@ -0,0 +1,107 @@
+import numpy as np
+from numpy.linalg import norm
+from skglm.estimators import Lasso
+
+
+def _reweight_by_coefficient_norm(coef):
```
mathurinm, line 6: What about calling it `L05_weights` ?
this way, we'd pave the way for a more fexible design where a penalty is passed, and penalty.weights() is called. 
WDYT?

this would allows us to solve MCP or L23 with reweighting with a simple change

skglm/experimental/reweighted.py
```diff
@@ -0,0 +1,107 @@
...
+    nrm = np.sqrt(norm(coef))
+    return 1 / (2 * nrm + np.finfo(float).eps)
+
+
+class ReweightedLasso(Lasso):
+    r"""Reweighted Lasso estimator.
```
mathurinm, line 12: Let's go for a general scheme, where the objective is not explicitely L05. We just solve a concave penalty by a MM/DC approach.

skglm/experimental/reweighted.py
```diff
@@ -0,0 +1,107 @@
...
+
+    Parameters
+    ----------
+    alpha : float, optional
+        Penalty strength.
+
+    n_reweights : int, optional
+        Number of reweighting iterations.
+
+    args: list
```
mathurinm, line 29: pass solver directly as we do in the v0.2 API

skglm/experimental/reweighted.py
```diff
@@ -0,0 +1,107 @@
...
+        Parameters for the inner Lasso estimator.
+        Note that `warm_start` is consistently set to `True` for performance reasons.
+
+    kwargs: dict
```
mathurinm, line 33: same. args and kwargs should be avoided as much as possible

skglm/experimental/reweighted.py
```diff
@@ -0,0 +1,107 @@
...
+        Objective history after every reweighting iteration
+
+    References
+    ----------
+    .. [1] Candès et al. (2007), Enhancing sparsity by reweighted l1 minimization
+           https://web.stanford.edu/~boyd/papers/pdf/rwl1.pdf
+    """
+
+    def __init__(self, alpha=1., n_reweights=5, *args, **kwargs):
+        # Warm start is crucial to ensure the solution of surrogate problem i
```
mathurinm, line 51: warm start is not related to the fact that we solve n_rewieghting problem IMO.

Instead warm start means coef_ is reused at each call to fit.

skglm/experimental/reweighted.py
```diff
@@ -0,0 +1,107 @@
...
+        Returns
+        -------
+        self :
+            Fitted estimator.
+        """
+        self.loss_history_ = []
+        n_samples, n_features = X.shape
+        weights = np.ones(n_features)
+        # XXX: dot product X @ w is slow in high-dimension, to be improved
+        objective = (lambda w: np.sum((y - X @ w) ** 2) / (2 * n_samples)
```
mathurinm, line 82: if it has a name it's not a lambda function :) 
you can do def objective(w): ...

mathurinm (changes requested):

skglm/experimental/reweighted.py
PABannier: Agreed!

skglm/experimental/reweighted.py
```diff
@@ -0,0 +1,104 @@
...
+def _L05_objective(coef):
+    return np.sqrt(np.abs(coef))
+
+
+def _L05_weights(coef):
+    return 1. / (2. * np.sqrt(np.abs(coef)) + np.finfo(float).eps)
+
+
+def _log_sum_objective(coef, eps=0.01):
+    return np.log(eps + np.abs(coef))
```
mathurinm, line 15: same remark, I think we can have infty as weight without any issue

skglm/experimental/reweighted.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+def _log_sum_weights(coef, eps=0.01):
+    return 1. / (eps + np.abs(coef))
```
mathurinm, line 19: Same remark, I think we handle np.inf gracefully in the weights

skglm/experimental/reweighted.py
```diff
@@ -0,0 +1,104 @@
...
+    .. [1] Candès et al. (2007), Enhancing sparsity by reweighted l1 minimization
+           https://web.stanford.edu/~boyd/papers/pdf/rwl1.pdf
+    """
+
+    def __init__(self, alpha, datafit=None, solver=None, n_reweights=5):
+        super().__init__(
+            datafit=datafit, penalty=WeightedL1(alpha, np.ones(1)), solver=solver)
+        self.n_reweights = n_reweights
+
+    def fit(self, X, y, pen_obj=_L05_objective, pen_weight=_L05_weights):
```
mathurinm, line 63: fit should only have X and y as arguments for sklearn compliance

skglm/experimental/reweighted.py
```diff
@@ -0,0 +1,104 @@
...
+
+def _log_sum_objective(coef, eps=0.01):
+    return np.log(eps + np.abs(coef))
+
+
+def _log_sum_weights(coef, eps=0.01):
+    return 1. / (eps + np.abs(coef))
+
+
+class ReweightedEstimator(GeneralizedLinearEstimator):
```
mathurinm, line 22: What do you think of IterativeReweightedL1 ? The same way we have Lasso

toy_reweight.py
```diff
@@ -0,0 +1,25 @@
+import numpy as np
```
mathurinm, line 1: can you come up with an interesting example ? For example comparing the solution with a direct prox cd approach and a reweighted one in terms of timing, minimum found, prediction error, support recovery on a simulation ?

skglm/experimental/_plot_reweight_lasso.py
```diff
@@ -0,0 +1,69 @@
...
+
+iterative_l05 = IterativeReweightedL1(
+    alpha, solver=AndersonCD(tol=tol, fit_intercept=False))
+
+# `subdiff` strategy for WS is uninformative for L0_5
+direct_l05 = GeneralizedLinearEstimator(
+    penalty=L0_5(alpha),
+    solver=AndersonCD(tol=tol, fit_intercept=False, ws_strategy="fixpoint"))
+
+# TODO: cache compilation
```
QB3, line 35: do you still want to do it?

skglm/experimental/reweighted.py
```diff
@@ -0,0 +1,107 @@
...
+        """
+        self.loss_history_ = []
+        n_features = X.shape[1]
+        self.penalty.weights = np.ones(n_features)
+
+        for iter_reweight in range(self.n_reweights):
+            super().fit(X, y)
+            self.penalty.weights = self.pen_weight(self.coef_)
+
+            # TODO: dot product X @ w is slow in high-dimension, to be improved
```
QB3, line 97: This is interesting, maybe we could add an `Xcoef_` attribute in the estimator (this is a longer shot)

skglm/experimental/_plot_reweight_lasso.py
```diff
@@ -0,0 +1,69 @@
...
+start = time.time()
+direct_l05.fit(X, y)
+print("Direct prox:", time.time() - start)
+
+print("#" * 20)
+
+# non-convex, not necessary same solution
+print("Coefficient norms")
+print("Reweighting:", norm(iterative_l05.coef_))
+print("Direct:", norm(direct_l05.coef_))
```
QB3, line 58: I would harmonize using `Direct prox` as before

skglm/experimental/_plot_reweight_lasso.py
```diff
@@ -0,0 +1,69 @@
...
+
+print("#" * 20)
+print("Objective value")
+print("Reweighting:", _obj(iterative_l05.coef_))
+print("Direct:", _obj(direct_l05.coef_))
```
QB3, line 63: I would harmonize using `Direct prox` as before

skglm/experimental/_plot_reweight_lasso.py
```diff
@@ -0,0 +1,69 @@
...
+
+print("#" * 20)
+
+print("Support recovery (F1 score)")
+print("Reweighting:", f1_score(w_true != 0, iterative_l05.coef_ != 0))
+print("Direct:", f1_score(w_true != 0, direct_l05.coef_ != 0))
```
QB3, line 69: I would harmonize using `Direct prox` as before

skglm/experimental/__init__.py
```diff
@@ -1,5 +1,9 @@
+from .reweighted import IterativeReweightedL1, _L23_objective, _L23_derivative
```
mathurinm, line 1: object prefixed with _ are "private" and should not be exposed. Can you find a workaround? Which piece of code needs them like this?

skglm/experimental/_plot_reweight_lasso.py
```diff
@@ -0,0 +1,69 @@
+import time
```
mathurinm, line 1: script name: plot_reweighted_L1 ?

skglm/experimental/reweighted.py
```diff
@@ -0,0 +1,115 @@
+import numpy as np
+from skglm.estimators import GeneralizedLinearEstimator
+from skglm.penalties import WeightedL1
+
+
+def _L05_objective(coef):
```
mathurinm, line 6: can you use the L05 penalty class?

skglm/experimental/reweighted.py
```diff
@@ -0,0 +1,115 @@
...
+    return np.sqrt(np.abs(coef))
+
+
+def _L05_derivative(coef):
```
mathurinm, line 10: can you implement this as penalty.derivative?

skglm/experimental/reweighted.py
```diff
@@ -0,0 +1,115 @@
...
+
+def _L23_objective(coef):
+    return np.abs(coef) ** (2/3)
+
+
+def _L23_derivative(coef):
+    return 2 / 3 / (np.abs(coef) ** (1/3) + np.finfo(float).eps)
+
+
+def _log_sum_objective(coef):
```
mathurinm, line 22: Can you create a log sum penalty class?

skglm/experimental/reweighted.py
```diff
@@ -0,0 +1,115 @@
...
+        Datafit. If None, `datafit` is initialized as a `Quadratic` datafit.
+        `datafit` is replaced by a JIT-compiled instance when calling fit.
+
+    solver : instance of BaseSolver, optional
+        Solver. If None, `solver` is initialized as an `AndersonCD` solver.
+
+    n_reweights : int, optional
+        Number of reweighting iterations.
+
+    pen_obj : Callable, optional
```
mathurinm, line 52: it would be simpler if it was just `penalty` that was passed

skglm/experimental/reweighted.py
```diff
@@ -0,0 +1,115 @@
...
+        Compute the concave objective.
+
+    pen_weight: Callable, optional
+        Compute the penalty weights from the coefficients.
+
+    Attributes
+    ----------
+    coef_ : array, shape (n_features,)
+        parameter vector (w in the cost function formula)
```
mathurinm, line 61: Parameter

skglm/experimental/reweighted.py
```diff
@@ -0,0 +1,115 @@
...
+            self.penalty.weights = self.pen_weight(self.coef_)
+
+            # TODO: dot product X @ w is slow in high-dimension, to be improved
+            loss = (self.datafit.value(y, self.coef_, X @ self.coef_)
+                    + self.penalty.alpha * np.sum(self.pen_obj(self.coef_)))
+            self.loss_history_.append(loss)
+
+            if self.solver.verbose:
+                print("#" * 10)
+                print(f"[REWEIGHT] iteration {iter_reweight} :: loss {loss}")
```
mathurinm, line 112: try to keep outputs in line with what's done in other solvers

mathurinm (changes requested):

skglm/experimental/reweighted.py
PABannier: I'd propose to do it in another PR. @Badr-MOUFAD , can you do it?

skglm/experimental/reweighted.py
```diff
@@ -0,0 +1,90 @@
...
+    .. [1] Candès et al. (2007), Enhancing sparsity by reweighted l1 minimization
+           https://web.stanford.edu/~boyd/papers/pdf/rwl1.pdf
+    """
+
+    def __init__(self, alpha, datafit=None, penalty=L0_5(1.), solver=None,
+                 n_reweights=5):
+        super().__init__(
+            datafit=datafit, penalty=WeightedL1(alpha, np.ones(1)), solver=solver)
+        self.n_reweights = n_reweights
+        self.nncvx_penalty = penalty
```
PABannier, line 55: Do you have a better name? Any idea on how to handle both penalties. I find this naming messy.

skglm/penalties/separable.py
```diff
@@ -429,6 +434,11 @@ def value(self, w):
         """Compute the value of the L2_3 norm at w."""
         return self.alpha * np.sum(np.abs(w) ** (2/3))
 
+    def derivative(self, w):
+        """Compute the element-wise derivative."""
+        # XXX: no alpha, not strcitly speaking derivative. WDYT?
```
PABannier, line 439: WDYT?

skglm/experimental/reweighted.py
mathurinm: use penalty for this one, and remove WeightedL1. then see below

mathurinm: I'd use penalty for the non convex one, then patch before and after fitting. Don't know if it's super clean... But otherwise can't you use solver.solver rather than super().fit ?

mathurinm:
> Coming a bit late, but from the "optimization" point of view, the following might be more natural, since we solve the same optimization problem, but with different solvers. Wdyt @mathurinm ?
> 
> ```
> direct_l05 = GeneralizedLinearEstimator(
>     penalty=L0_5(alpha),
>     solver=AndersonCD(tol=tol, fit_intercept=False, ws_strategy="fixpoint"))
> 
> iterative_l05 = GeneralizedLinearEstimator(
>     penalty=L0_5(alpha),
>     solver=Reweighted(n_reweighted, AndersonCD(tol=tol, fit_intercept=False, ws_strategy="fixpoint")))
> ```

That's a good alternative suggestion. Since everything is in `skglm.experimental`, I think we can go with the current design that has the benefit of being already written, but if you can send a PR with the alternate proposition later that'd be great for comparison.

skglm/experimental/_plot_reweight_l1.py
```diff
@@ -0,0 +1,53 @@
+import numpy as np
```
mathurinm, line 1: do we move this file to the doc so that it's built with other examples?

skglm/experimental/_plot_reweight_l1.py
```diff
@@ -0,0 +1,53 @@
...
+
+# `subdiff` strategy for WS is uninformative for L0_5
+direct_l05 = GeneralizedLinearEstimator(
+    penalty=L0_5(alpha),
+    solver=AndersonCD(tol=tol, fit_intercept=False, ws_strategy="fixpoint"))
+
+iterative_l05.fit(X, y)
+
+# reweighting can't increase the L0.5 objective
+assert iterative_l05.loss_history_[0] > iterative_l05.loss_history_[-1]
```
mathurinm, line 38: Do you think we can remove the assertion from the plot ?

skglm/experimental/_plot_reweight_l1.py
```diff
@@ -0,0 +1,53 @@
...
+direct_l05.fit(X, y)
+
+print("#" * 20)
+print("Objective value")
+print("Reweighting:", _obj(iterative_l05.coef_))
+print("Direct prox:", _obj(direct_l05.coef_))
+
+print("#" * 20)
+
+print("Support recovery (F1 score)")
```
mathurinm, line 51: I'm under the impression that we try to show too many things here. Just comparing objectives seems enough to me, without going into the details of statistical experiments. WDYT? Maybe just a comparison of timings and found objectives for say 3 alphas ?

skglm/experimental/reweighted.py
```diff
@@ -0,0 +1,99 @@
+import numpy as np
+from skglm.datafits import Quadratic
+from skglm.estimators import GeneralizedLinearEstimator
+from skglm.penalties import WeightedL1, L0_5
+from skglm.utils import compiled_clone
+
+
+# def _log_sum_objective(coef):
```
mathurinm, line 8: rm if not used to keep the `main` branch clean

mathurinm (commented):
Looks good, thanks. Let's only agree on what we want to display in the example, move the example to the full gallery, and that will be done!

skglm/experimental/_plot_reweight_l1.py
PABannier: Yes. I think we should have a simple test to check this assertion.

examples/plot_reweight_l1.py
```diff
@@ -0,0 +1,80 @@
...
+        "direct": (direct_l05, direct_time),
+    }
+    return results
+
+
+# caching Numba compilation
+fit_l05(1, alpha_max)
+
+# actual run
+for alpha in alphas:
```
mathurinm, line 61: just convert this into 2 bar plots and we're good to go on my side

skglm/experimental/reweighted.py
```diff
@@ -0,0 +1,90 @@
...
+from skglm.datafits import Quadratic
+from skglm.estimators import GeneralizedLinearEstimator
+from skglm.penalties import WeightedL1, L0_5
+from skglm.utils import compiled_clone
+
+
+class IterativeReweightedL1(GeneralizedLinearEstimator):
+    r"""Reweighted L1-norm estimator.
+
+    This estimator iteratively solves a non-convex problems by iteratively solving
```
Badr-MOUFAD, line 11:
```suggestion
    This estimator solves a non-convex problems by iteratively solving
```

Badr-MOUFAD (commented):
Thanks @PABannier for the hard work!

Some minor remarks then LGTM.

skglm/penalties/separable.py
```diff
@@ -380,6 +380,10 @@ def value(self, w):
         """Compute the value of L0_5 at w."""
         return self.alpha * np.sum(np.abs(w) ** 0.5)
 
+    def derivative(self, w):
+        """Compute the element-wise derivative."""
+        return 1. / (2. * np.sqrt(np.abs(w)) + 1e-12)
```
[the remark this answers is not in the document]
PABannier: yes. I usually use `np.finfo(float).eps` which is more consistent, but it is not JIT-compilable. Maybe a simpler solution would consist in passing `eps=1e-12` as derivative. WDYT?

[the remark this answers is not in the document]
Badr-MOUFAD: I don't think it solves the duplication problem.
I would proceed as follows:
```python
# small number added to a denominator to prevent it from being zero
EPS_DERIVATIVE = 1e-12
```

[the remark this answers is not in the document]
PABannier: I feel like it's a bit overkill to add a const variable in a large module like this. The PR should be finalized now. No strong opinions. Feel free to add it (cc @mathurinm )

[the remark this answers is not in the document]
mathurinm: Yes I agree, it feels overkill for now. One less variable to expose

mathurinm (approved):

mathurinm:
Thanks @PABannier @Badr-MOUFAD

PR #91, opened 2022-10-12
ENH Add FISTA solver

### LINKED ISSUE #89: FEAT add FISTA solver
could be useful for the SLOPE penalty

PABannier:
Closes #89 

A few points to discuss:

1) Currently the FISTA solver uses Gram updates (as per #4 ). Question: do we want to keep it this way? Implement without Gram update? Or have two options?

2) For non-coordinate-wise updates, we run into the issue of not having a `prox_vec` method in the `BasePenalty` class. If we want to support a larger class of penalties for FISTA (e.g.: L1, WeightedL1, SLOPE, ...), we need a `prox_vec` method.


mathurinm:
- Remove Gram to handle the generic case (+ gram is only suited to quadratics) 
- To keep the API simple, can you do a for loop over coordinates to compute the prox, calling prox_1D ? We'll lose a bit of time but the gradient computation should be the dominating cost

skglm/solvers/fista.py
```diff
@@ -0,0 +1,67 @@
+import numpy as np
+from numba import njit
+from skglm.solvers.base import BaseSolver
+
+
+@njit
+def _prox_vec(w, z, penalty, lipschitz):
```
mathurinm, line 7: I like it for now. I think it's painful for the user to define both prox and prox1d

skglm/solvers/fista.py
```diff
@@ -0,0 +1,67 @@
...
+    # XXX: TO DISCUSS: should add a vectorized prox update
+    n_features = w.shape[0]
+    for j in range(n_features):
+        w[j] = penalty.prox_1d(z[j], 1 / lipschitz, j)
+    return w
+
+
+class FISTA(BaseSolver):
+    r"""ISTA solver with Nesterov acceleration (FISTA)."""
```
mathurinm, line 16: FISTA solver by Beck and Teboulle

skglm/solvers/fista.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def __init__(self, max_iter=100, tol=1e-4, fit_intercept=False, warm_start=False,
+                 opt_freq=100, verbose=0):
```
mathurinm, line 19: It doesn't seem other solvers have `opt_freq` so there's no comparison, but why 100 here ?

skglm/solvers/fista.py
PABannier: +1

skglm/solvers/fista.py
```diff
@@ -0,0 +1,67 @@
...
+        t_new = 1
+
+        w = w_init.copy() if w_init is not None else np.zeros(n_features)
+        z = w_init.copy() if w_init is not None else np.zeros(n_features)
+        weights = weights if weights is not None else np.ones(n_features)
+
+        # FISTA with Gram update
+        G = X.T @ X
+        Xty = X.T @ y
+        lipschitz = np.linalg.norm(X, ord=2) ** 2 / n_samples
```
mathurinm, line 40: this will be an issue: for logreg it differs. For Poisson it does not exist

skglm/solvers/fista.py
PABannier: I used `opt_freq` to avoid computing the optimality conditions at every FISTA iteration. I know we don't implement this in other solvers but it feels like an outer iteration of WS + CD is not the same as FISTA. For 100, purely "au doigt mouillé".

skglm/solvers/fista.py
PABannier: Can we solve Poisson with FISTA? I thought we needed a line search.

skglm/solvers/fista.py
mathurinm: maybe 10 instead of 100 ? cost is the same as 1 single iteration, 10 % seems reasonable

skglm/solvers/fista.py
```diff
@@ -0,0 +1,93 @@
+import numpy as np
+from scipy.sparse import issparse
+from numba import njit
+from skglm.solvers.base import BaseSolver
+from skglm.solvers.common import construct_grad, construct_grad_sparse
+
+
+@njit
```
Badr-MOUFAD, line 8: cosmetics, can you put this function below the class?

skglm/solvers/fista.py
```diff
@@ -0,0 +1,93 @@
...
+
+    Attributes
+    ----------
+    max_iter : int, default 100
+        Maximum number of iterations.
+
+    tol : float, default 1e-4
+        Tolerance for convergence.
+
+    opt_freq : int, default 10
```
Badr-MOUFAD, line 29: Is it useful to expose the param to control the frequency of optimality check?

skglm/solvers/fista.py
```diff
@@ -0,0 +1,93 @@
...
+                    print(
+                        f"Iteration {n_iter+1}: {p_obj:.10f}, "
+                        f"stopping crit: {stop_crit:.2e}"
+                    )
+
+                if stop_crit < self.tol:
+                    if self.verbose:
+                        print(f"Stopping criterion max violation: {stop_crit:.2e}")
+                    break
+        return w
```
Badr-MOUFAD, line 93: can you return ``obj_out``, ``stop_crit`` to align with the existing solvers?

skglm/solvers/fista.py
```diff
@@ -0,0 +1,93 @@
...
+            w_old = w.copy()
+            if issparse(X):
+                grad = construct_grad_sparse(
+                    X.data, X.indptr, X.indices, y, z, X @ z, datafit, all_features)
+            else:
+                grad = construct_grad(X, y, z, X @ z, datafit, all_features)
+            z -= grad / lipschitz
+            w = _prox_vec(w, z, penalty, lipschitz)
+            Xw = X @ w
+            z = w + (t_old - 1.) / t_new * (w - w_old)
```
Badr-MOUFAD, line 76: I suggest using ``w_extrapolated`` instead of ``z``
WDYT?

skglm/solvers/fista.py
```diff
@@ -0,0 +1,93 @@
...
+    def __init__(self, max_iter=100, tol=1e-4, opt_freq=10, verbose=0):
+        self.max_iter = max_iter
+        self.tol = tol
+        self.opt_freq = opt_freq
+        self.verbose = verbose
+
+    def solve(self, X, y, datafit, penalty, w_init=None, Xw_init=None):
+        n_samples, n_features = X.shape
+        all_features = np.arange(n_features)
+        t_new = 1
```
Badr-MOUFAD, line 52:
```suggestion
        t_new = 1.
```

skglm/solvers/fista.py
```diff
@@ -0,0 +1,93 @@
...
+
+@njit
+def _prox_vec(w, z, penalty, lipschitz):
+    n_features = w.shape[0]
+    for j in range(n_features):
+        w[j] = penalty.prox_1d(z[j], 1 / lipschitz, j)
+    return w
+
+
+class FISTA(BaseSolver):
```
Badr-MOUFAD, line 16: What about fitting an intercept?

Badr-MOUFAD (commented):
Thanks for the hard work @PABannier!

Below, some minor remarks.

Besides, I have one concern: I don't think it's a good idea to add support for FISTA to all the datafits. We can limit ourselves to one of them just for testing purposes. 
Indeed, ``AndersonCD`` and ``ProxNewton`` are much faster for separable problems. We better keep FISTA for particular cases (e.g SLOPE #92)

WDYT?

PABannier:
@Badr-MOUFAD totally agree. FISTA would be for a subset of penalties where PN or AndersonCD are not available.

skglm/datafits/single_task.py
```diff
@@ -140,13 +153,16 @@ def raw_hessian(self, y, Xw):
...
+        self.global_lipschitz = norm(X, ord=2) ** 2 / (len(y) * 4)
 
     def initialize_sparse(self, X_data, X_indptr, X_indices, y):
         n_features = len(X_indptr) - 1
         self.lipschitz = np.zeros(n_features, dtype=X_data.dtype)
+        self.global_lipschitz = 0.
         for j in range(n_features):
             Xj = X_data[X_indptr[j]:X_indptr[j+1]]
             self.lipschitz[j] = (Xj ** 2).sum() / (len(y) * 4)
+            self.global_lipschitz += (Xj ** 2).sum() / (len(y) * 4)
```
mathurinm, line 165: that will yield a very crude bound, potentially with a loss or the order of `n_features`. 

Use a few iterations of the power method instead to approximate the lipschitz constant of the sparse matrix (there's also the Lanczos iteration but it's more complicated, let's implement the easy one first)

skglm/datafits/single_task.py
```diff
@@ -264,8 +289,16 @@ class Huber(BaseDatafit):
 
     Attributes
     ----------
+    delta : float
+        Shape hyperparameter.
```
mathurinm, line 293: Can we be more detailed mathematically here ? I don't know what shape means in this context. Is is a scale?

skglm/solvers/fista.py
mathurinm: or in skglm.utils ?

skglm/solvers/fista.py
mathurinm: z is quite standard in FISTA notation

skglm/solvers/fista.py
PABannier: I removed it.

skglm/solvers/fista.py
PABannier: `w_extrapolated` is confusing in our codebase. Someone could think we use AA under the hood.

skglm/solvers/fista.py
PABannier: I haven't really understood how to do it. I know @Klopfe is writing a piece of documentation about it, which I'll read eagerly. For now, I leave it as it and I open #97 for a subsequent PR.

skglm/datafits/single_task.py
PABannier: I've changed it to `Threshold` as it separates the domain where the function is quadratic and where it becomes linear (https://cvxr.rbind.io/cvxr_examples/cvxr_huber-regression/). Open to any suggestions

skglm/datafits/single_task.py
```diff
@@ -62,6 +69,7 @@ def initialize_sparse(
 
             self.lipschitz[j] = nrm2 / len(y)
             self.Xty[j] = xty
+            self.global_lipschitz += nrm2 / len(y)
```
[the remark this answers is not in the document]
PABannier: I didn't know that... Why? Is it explained somewhere? A reference?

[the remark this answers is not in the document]
Badr-MOUFAD: > I didn't know that... Why? Is it explained somewhere? A reference?

- For matrix inequality: [wikipediea ref](https://en.wikipedia.org/wiki/Matrix_norm#cite_note-13 ) section equivalence of norms or cf. [book](https://convexoptimization.com/TOOLS/MatrixComp.pdf) page 56.
- If you apply ``norm(X, ord=2)`` on a sparse matrix, you will get an error.

That being said, have a look on @mathurinm suggestion in https://github.com/scikit-learn-contrib/skglm/pull/91#discussion_r996412138

skglm/utils.py
```diff
@@ -457,3 +457,73 @@ def extrapolate(self, w, Xw):
         C = inv_UTU_ones / np.sum(inv_UTU_ones)
         # floating point errors may cause w and Xw to disagree
         return self.arr_w_[:, 1:] @ C, self.arr_Xw_[:, 1:] @ C, True
+
+
+@njit
+def prox_vec(w, z, penalty, lipschitz):
```
mathurinm, line 463: maybe make this private to have less doc?

mathurinm (changes requested):

skglm/solvers/fista.py
```diff
@@ -0,0 +1,82 @@
...
+from scipy.sparse import issparse
+from skglm.solvers.base import BaseSolver
+from skglm.solvers.common import construct_grad, construct_grad_sparse
+from skglm.utils import _prox_vec
+
+
+class FISTA(BaseSolver):
+    r"""ISTA solver with Nesterov acceleration (FISTA).
+
+    This solver implements accelerated proximal gradient descent for linear problems.
```
mathurinm, line 11: What does linear mean here ? If it's that the datafit is quadratic, it'd be better to state it explicitely

skglm/solvers/fista.py
```diff
@@ -0,0 +1,82 @@
...
+            w_old = w.copy()
+            if issparse(X):
+                grad = construct_grad_sparse(
+                    X.data, X.indptr, X.indices, y, z, X @ z, datafit, all_features)
+            else:
+                grad = construct_grad(X, y, z, X @ z, datafit, all_features)
+            z -= grad / lipschitz
+            w = _prox_vec(w, z, penalty, lipschitz)
+            Xw = X @ w
+            z = w + (t_old - 1.) / t_new * (w - w_old)
```
mathurinm, line 65: maybe use z[:]

skglm/tests/test_fista.py
```diff
@@ -0,0 +1,67 @@
...
+
+    solver_cd = AndersonCD(tol=tol, fit_intercept=False)
+    w_cd = solver_cd.solve(X, _y, datafit, penalty)[0]
+
+    np.testing.assert_allclose(w_fista, w_cd, atol=1e-7)
+
+
+def test_spectral_norm2():
+    n_samples, n_features = 50, 60
+    A_sparse = scipy.sparse.random(n_samples, n_features, density=0.5, format='csc')
```
mathurinm, line 57: Is this test deterministic ?

skglm/tests/test_fista.py
```diff
@@ -0,0 +1,67 @@
...
+
+    A_bundles = (A_sparse.data, A_sparse.indptr, A_sparse.indices)
+    lipschitz_our = spectral_norm2(*A_bundles, n_samples=len(y))
+    lipschitz_np = norm(A_sparse.toarray(), ord=2) ** 2
```
mathurinm, line 61: FYI it should be faster to use scipy.sparse.linalg.svds(A, k=1)

skglm/utils.py
```diff
@@ -457,3 +457,114 @@ def extrapolate(self, w, Xw):
         C = inv_UTU_ones / np.sum(inv_UTU_ones)
         # floating point errors may cause w and Xw to disagree
         return self.arr_w_[:, 1:] @ C, self.arr_Xw_[:, 1:] @ C, True
+
+
+@njit
+def _prox_vec(w, z, penalty, lipschitz):
+    # evaluate the vectorized proximal operator for the FISTA solver
```
mathurinm, line 464: it's not restricted to the fista solver

skglm/utils.py
```diff
@@ -457,3 +457,114 @@ def extrapolate(self, w, Xw):
         C = inv_UTU_ones / np.sum(inv_UTU_ones)
         # floating point errors may cause w and Xw to disagree
         return self.arr_w_[:, 1:] @ C, self.arr_Xw_[:, 1:] @ C, True
+
+
+@njit
+def _prox_vec(w, z, penalty, lipschitz):
```
mathurinm, line 463: rename lipschitz to step and use step=1/lipschitz. Otherwise this will become wrong if we decide to use linesearch at some point

skglm/utils.py
```diff
@@ -457,3 +457,114 @@ def extrapolate(self, w, Xw):
...
+    # evaluate the vectorized proximal operator for the FISTA solver
+    # lipschitz stands for global lipschitz constant
+    n_features = w.shape[0]
+    for j in range(n_features):
+        w[j] = penalty.prox_1d(z[j], 1 / lipschitz, j)
+    return w
+
+
+@njit
+def spectral_norm2(X_data, X_indptr, X_indices, n_samples,
```
mathurinm, line 473: the 2 is ambiguous here. 
rename the function to spectral_norm, return the sqrt of what's being computed, and square it when needed. It'll avoid surprises

skglm/utils.py
```diff
@@ -457,3 +457,114 @@ def extrapolate(self, w, Xw):
...
+                   max_iter=20, tol=1e-6):
+    """Compute the squared spectral norm of sparse matrix ``X``.
+
+    Find the largest eigenvalue of ``X @ X.T`` using the power method.
```
mathurinm, line 477: rename to "largest singular value of X" when you have used sqrt

skglm/tests/test_fista.py
Badr-MOUFAD: True. It's much faster than NumPy but not enough compared to us.

skglm/solvers/fista.py
Badr-MOUFAD: @mathurinm is there a difference here?
``w + (t_old - 1.) / t_new * (w - w_old)`` returns a new instance either ways.

skglm/solvers/fista.py
Badr-MOUFAD: I think "ISTA solver with Nesterov acceleration (FISTA)." is enough.
We better remove this statement.
WDYT?

mathurinm: I agree :+1:

skglm/solvers/fista.py
mathurinm: I don't know if that makes a real difference indeed. Let's keep as is then

Badr-MOUFAD:
I am unable to find the root cause of the problem at the uniitest.

Here is a small script to reproduce

```python
import numpy as np
from  scipy.sparse import csc_matrix
from skglm.estimators import LinearSVC

n_samples, n_features = 20, 30
X = np.random.randn(n_samples, n_features)
X_sparse = csc_matrix(X * np.random.binomial(1, 0.1, X.shape))
y = np.sign(np.random.randn(n_samples))

LinearSVC(C=1., tol=1e-9).fit(X_sparse, y)
```

**Output** (it depends)
```bash
Segmentation fault (core dumped)
```
or
```bash
corrupted size vs. prev_size
Aborted (core dumped)
```
or
```bash
python3: malloc.c:3852: _int_malloc: Assertion `chunk_main_arena (fwd)' failed.
Aborted (core dumped)
```

@mathurinm, @PABannier, any thoughts?

PABannier:
A segfault is usually thrown by Numba when it can't access something it should (e.g. missing initialization of datafit). Have you tried setting breakpoints at various places of the code to see which line is causing the issue?

Badr-MOUFAD:
Yes absolutely, I tried that. it breaks down in the initialization of datafit. Yet, I can't figure out why.
What is surprising is that it works for some X sizes,

PABannier:
Weird, that works for me on this branch. Can you try reinstalling `numba` and `skglm`?

Badr-MOUFAD:
I found the bug.
In the case of SVC, the design matrix is ``yXT`` which has particularly ``n_rows=n_features`` instead of ``n_samples``.
This caused an index out of range in ``spectral_norm``.

skglm/datafits/single_task.py
```diff
@@ -44,14 +50,18 @@ def params_to_dict(self):
...
             self.lipschitz[j] = (X[:, j] ** 2).sum() / len(y)
 
-    def initialize_sparse(
-            self, X_data, X_indptr, X_indices, y):
+    def initialize_sparse(self, X_data, X_indptr, X_indices, y):
         n_features = len(X_indptr) - 1
         self.Xty = np.zeros(n_features, dtype=X_data.dtype)
+
+        self.global_lipschitz = spectral_norm(X_data, X_indptr, X_indices, len(y)) ** 2
+        self.global_lipschitz /= len(y)
```
PABannier, line 63: As suggested previously, I would add a few comments on why we need the `spectral_norm` function. Like
```
# L2-norm of X not available in the sparse case, `spectral_norm` computes an upper bound on the spectral radius.
```

or something along those lines

mathurinm: well no, this time it's the exact spectral norm (beware, not equal to spectral radius. spectral radius is the largest eigenvalue of X, it can be 0 without X being 0)

for me it's explicit enough as is

mathurinm:
Wow, good catch @Badr-MOUFAD 

For a more robust design replace n_samples by n_rows_X and pass X.shape[0] explicitly

mathurinm (approved):

mathurinm:
Thanks @PABannier and @Badr-MOUFAD

PR #92, opened 2022-10-14
ENH Add SLOPE penalty

PABannier:
POC of the SLOPE penalty with FISTA solver. We should first merge #91 before this one.


poc_slope.py
```diff
@@ -0,0 +1,98 @@
...
+        spec = (
+            ('alphas', float64[:]),
+        )
+        return spec
+    
+    def params_to_dict(self):
+        return dict(alphas=self.alphas)
+
+    def value(self, w):
+        sorted_indices = np.argsort(w)[::-1]  # descending order
```
mathurinm, line 37: should be an absolute value here

poc_slope.py
```diff
@@ -0,0 +1,98 @@
...
+        return np.sum(w * self.alphas[sorted_indices])
```
mathurinm, line 38: should be an absolute value here too.

Maybe the one liner `np.sum(np.sort(np.abs(w)) * alphas[::-1])` is simpler and less error prone?

poc_slope.py
```diff
@@ -0,0 +1,98 @@
...
+    def prox_vec(self, x, stepsize):
+        def _prox(_x, _alphas):
+            sign_x = np.sign(_x)
+            _x = np.abs(_x)
+            sorted_indices = np.argsort(_x)[::-1]
+            prox = fast_prox_SL1(_x[sorted_indices], _alphas)
+            prox[sorted_indices] = prox
+            return prox * sign_x
+        return _prox(x, self.alphas * stepsize)
+
```
mathurinm, line 49: maybe add a coordinatewise prox that throws an error if called, to make the design robust ?

poc_slope.py
```diff
@@ -0,0 +1,98 @@
...
+
+@njit
+def fast_prox_SL1(y, alphas):
```
mathurinm, line 52: y is used in the sklearn terminology in our codebase (observation vector), we can use another letter

poc_slope.py
```diff
@@ -0,0 +1,98 @@
...
+    # w, alphas: nonnegative and nonincreasing sequences
```
mathurinm, line 53: w is not used here

poc_slope.py
```diff
@@ -0,0 +1,98 @@
...
+    n_features = y.shape[0]
+    x = np.empty(n_features)
+
+    k = 0
+    idx_i = np.empty((n_features,), dtype=np.int64)
+    idx_j = np.empty((n_features,), dtype=np.int64)
+    s = np.empty((n_features,), dtype=np.float64)
+    w = np.empty((n_features,), dtype=np.float64)
+
+    for i in range(n_features):
```
mathurinm, line 63: could you add a link to a source for this algorithm ?

poc_slope.py
```diff
@@ -0,0 +1,98 @@
...
+            k -= 1
+            idx_j[k] = i
+            s[k] += s[k+1]
+            w[k] = s[k] / (i - idx_i[k] + 1)
+
+        k += 1
+
+    for j in range(k):
+        d = w[j]
+        d = 0 if d < 0 else d
```
mathurinm, line 79: is this max(d, 0) ?

PABannier: it is. the whole function `prox` is inefficient at the moment, it is a raw copy-paste of an implementation in C.

mathurinm: I'm not sure it's inefficient, if jitted. I've seen a similar implem by @jolars

PABannier:
Note to self:
Work remaining: implemented vector fixed-point violation for stopping criterion and fix vector prox computation for FISTA.

skglm/penalties/__init__.py
```diff
@@ -1,6 +1,6 @@
 from .base import BasePenalty
 from .separable import (
-    L1_plus_L2, L0_5, L1, L2_3, MCPenalty, SCAD, WeightedL1, IndicatorBox
+    L1_plus_L2, L0_5, L1, L2_3, MCPenalty, SCAD, WeightedL1, IndicatorBox, SLOPE
```
mathurinm, line 3: deserves discussion: SLOPE is not separable. Where do we put it ?

skglm/penalties/separable.py
```diff
@@ -466,3 +467,45 @@ def is_penalized(self, n_features):
...
+
+    def params_to_dict(self):
+        return dict(alphas=self.alphas)
+
+    def value(self, w):
+        """Compute the value of SLOPE at w."""
+        return np.sum(np.sort(np.abs(w)) * self.alphas[::-1])
+
+    def prox_1d(self, value, stepsize, j):
+        raise NotImplementedError(
```
mathurinm, line 500: What do you think of ValueError ? Because  it's not that we did not implement it, it's that  it does not exist

skglm/penalties/__init__.py
PABannier: What does it imply for the user? If we want to be rigorous, maybe we should put it in `experimental`.

skglm/tests/test_penalties.py
```diff
@@ -68,5 +68,17 @@ def test_subdiff_diff_block(block_penalty):
     assert_array_less(est.stop_crit_, est.solver.tol)
 
 
+def test_slope():
+    # check that when alphas = [alpha, ..., alpha], SLOPE and L1 solutions are equal
+    alphas = np.repeat(alpha, n_features)
```
mathurinm, line 73: minor: I think np.full is standard here

mathurinm (commented):
minor changes, then LGTM!

skglm/penalties/__init__.py
Klopfe: Whether in the experimental folder or in a non-separable.py file in the penalties folder.

mathurinm:
What do you think of adding https://github.com/jolars/pyslope as a test dependecy for SLOPE and test that we find the same solution ?

Klopfe:
> What do you think of adding https://github.com/jolars/pyslope as a test dependecy for SLOPE and test that we find the same solution ?

I think it's a good idea. Then we can have a "true" test

skglm/penalties/__init__.py
Badr-MOUFAD: I agree with @Klopfe, we better add a new module ``non_seperable.py``. We've recently added FISTA solver which we intend to use for non-separable problems.

skglm/tests/test_penalties.py
```diff
@@ -68,5 +68,17 @@ def test_subdiff_diff_block(block_penalty):
     assert_array_less(est.stop_crit_, est.solver.tol)
 
 
+def test_slope():
+    # check that when alphas = [alpha, ..., alpha], SLOPE and L1 solutions are equal
+    alphas = np.full(n_features, alpha)
+    tol = 1e-10
+    est = GeneralizedLinearEstimator(
+        penalty=SLOPE(alphas),
+        solver=FISTA(max_iter=1000, tol=tol, opt_strategy="fixpoint"),
```
Badr-MOUFAD, line 77: test fails when ``opt_strategy="subdiff"`` since ``subdiff_distance`` is not implemented for SLOPE and returns ``None``,
However, we don't raise a meaningful exception fo
```shell
    if stop_crit < self.tol:
TypeError: '<' not supported between instances of 'NoneType' and 'float'
```
I have two alternatives to handle this case:
- define a ``subdiff_distance`` method in SLOPE and raise within it an exception
- add ``NotImplemented`` exception in ``BasePenalty``

PABannier: I would go with the first option. It's not so much that we haven't implemented it, but that it does not exist. So raise a ValueError.

skglm/solvers/fista.py
```diff
@@ -60,11 +63,21 @@ def solve(self, X, y, datafit, penalty, w_init=None, Xw_init=None):
...
             Xw = X @ w
             z = w + (t_old - 1.) / t_new * (w - w_old)
 
-            opt = penalty.subdiff_distance(w, grad, all_features)
+            if self.opt_strategy == "subdiff":
+                opt = penalty.subdiff_distance(w, grad, all_features)
+            elif self.opt_strategy == "fixpoint":
+                opt = np.abs(w - penalty.prox_vec(w - grad / lipschitz, 1 / lipschitz))
+            else:
+                raise ValueError("Unknown error optimality strategy. Expected " +
```
Badr-MOUFAD, line 78:
```suggestion
                raise ValueError(
                    "Unknown error optimality strategy. Expected "
                    f"`subdiff` or `fixpoint`. Got {self.opt_strategy}")
```

skglm/tests/test_penalties.py
```diff
@@ -4,12 +4,15 @@
 from numpy.linalg import norm
 from numpy.testing import assert_array_less
 
+from slope.solvers import pgd_slope
```
Badr-MOUFAD, line 7: Is there a reason for not importing these locally in ``test_slope`` and using ``.xfail`` if the dependence is not installed?

Cf. 
https://github.com/scikit-learn-contrib/skglm/blob/92c2a456b360a923f4cb961cea2a97c417fc13c2/skglm/experimental/tests/test_sqrt_lasso.py#L19-L23

Badr-MOUFAD (commented):
Some minor changes, then LGTM!

skglm/tests/test_penalties.py
PABannier: no. i'll change it, thx :)

mathurinm:
merge when green

Badr-MOUFAD (approved):
Thanks @PABannier, GG :muscle:!

PR #93, opened 2022-10-15
FIX Remove Numba warning when fitting multi-task estimators

PABannier:
Currently, when fitting a multi-task estimator, Numba is raising this warning (visible when doing benchmarks for instance):

```
/Users/pierre-antoine/Documents/skglm/skglm/solvers/multitask_bcd.py:294: NumbaPerformanceWarning: '@' is faster on contiguous arrays, called on (array(float64, 2d, A), array(float64, 2d, C))
  grad[idx, :] = datafit.gradient_j(X, Y, W, XW, j)
/Users/pierre-antoine/Documents/skglm/skglm/solvers/multitask_bcd.py:294: NumbaPerformanceWarning: '@' is faster on contiguous arrays, called on (array(float64, 2d, A), array(float64, 2d, C))
  grad[idx, :] = datafit.gradient_j(X, Y, W, XW, j)
```

The warning was caused by the following indexation `W[j:j+1, :]`, which can be simply changed to `W[j, :]`. No improvements (nor degradation) in speed have been observed.

(Something discovered by @QB3 in #39 )

### DIFF
--- a/skglm/datafits/multi_task.py
+++ b/skglm/datafits/multi_task.py
@@ -68,7 +68,7 @@ class QuadraticMultiTask(BaseMultitaskDatafit):
     def gradient_j(self, X, Y, W, XW, j):
         """Gradient with respect to j-th coordinate of W."""
         n_samples = X.shape[0]
-        return (X[:, j:j+1].T @ XW - self.XtY[j, :]) / n_samples
+        return (X[:, j].T @ XW - self.XtY[j, :]) / n_samples
 
     def gradient_j_sparse(self, X_data, X_indptr, X_indices, Y, XW, j):
         """Gradient with respect to j-th coordinate of W when X is sparse."""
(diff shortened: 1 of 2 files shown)


Badr-MOUFAD (approved):
Thanks @PABannier and @QB3, that was a nice catch indeed!

skglm/datafits/multi_task.py
```diff
@@ -68,7 +68,7 @@ def value(self, Y, W, XW):
     def gradient_j(self, X, Y, W, XW, j):
         """Gradient with respect to j-th coordinate of W."""
         n_samples = X.shape[0]
-        return (X[:, j:j+1].T @ XW - self.XtY[j, :]) / n_samples
+        return (X[:, j].T @ XW - self.XtY[j, :]) / n_samples
```
mathurinm, line 71: then, I guess there is no need to transpose the array since it's 1D ?

mathurinm (approved):

mathurinm:
Merge when green @PABannier, thanks for the catch. It feels like the warning should not be raised by numba here, maybe it's worth opening an issue with a minimal example on their repo?

PABannier:
@mathurinm actually the issue is already referenced:
- https://github.com/numba/numba/issues/8131
- https://github.com/numba/numba/issues/7840

PR #94, opened 2022-10-15
ENH - Add ``LogisitcGroup`` datafit

Badr-MOUFAD:
This adds a ``LogisticGroup`` datafit.

- [x] Implement ``LogisitcGroup`` datafit
- [x] unittest using group logistic regression

Soon, a ``GroupProxNewton`` solver :rocket:

### DIFF
--- a/skglm/datafits/group.py
+++ b/skglm/datafits/group.py
@@ -71,3 +71,78 @@ class QuadraticGroup(BaseDatafit):
 
     def intercept_update_step(self, y, Xw):
         return np.mean(Xw - y)
+
+
+class LogisticGroup(BaseDatafit):
+    r"""Logistic datafit used with group penalties.
+
+    The datafit reads::
+
+    (1 / n_samples) * \sum_i log(1 + exp(-y_i * Xw_i))
+
+    Attributes
+    ----------
+    grp_indices : array, shape (n_features,)
+        The group indices stacked contiguously
+        ([grp1_indices, grp2_indices, ...]).
+
+    grp_ptr : array, shape (n_groups + 1,)
+        The group pointers such that two consecutive elements delimit
+        the indices of a group in ``grp_indices``.
+
+    lipschitz : array, shape (n_groups,)
+        The lipschitz constants for each group.
+    """
+
+    def __init__(self, grp_ptr, grp_indices):
+        self.grp_ptr, self.grp_indices = grp_ptr, grp_indices
+
+    def get_spec(self):
+        spec = (
+            ('grp_ptr', int32[:]),
+            ('grp_indices', int32[:]),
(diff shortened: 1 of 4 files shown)


skglm/datafits/group.py
```diff
@@ -71,3 +71,78 @@ def gradient_scalar(self, X, y, w, Xw, j):
 
     def intercept_update_step(self, y, Xw):
         return np.mean(Xw - y)
+
+
+class LogisticGroup(BaseDatafit):
```
mathurinm, line 76: Could it inherit from Logistic to avoid duplicating code ?

skglm/tests/test_group.py
```diff
@@ -160,6 +161,66 @@ def test_intercept_grouplasso():
...
+    rnd = np.random.RandomState(1123)
+    X, y, _ = make_correlated_data(n_samples, n_features, random_state=rnd)
+    y = np.sign(y)
+
+    grp_indices, grp_ptr = grp_converter(1, n_features)
+    weights = np.ones(n_features)
+    alpha_max = norm(X.T @ y, ord=np.inf) / (2 * n_samples)
+    alpha = rho * alpha_max / 10.
+
+    log_group = LogisticGroup(grp_ptr=grp_ptr, grp_indices=grp_indices)
```
mathurinm, line 176: log means np.log for me. `logistic` removes this ambiguity

skglm/tests/test_group.py
```diff
@@ -160,6 +161,66 @@ def test_intercept_grouplasso():
...
+
+@pytest.mark.parametrize("n_groups, rho", [[15, 1e-1], [25, 1e-2]])
+def test_group_logreg(n_groups, rho):
+    n_samples, n_features, shuffle = 30, 60, True
+    random_state = 123
+
+    X, y, _ = make_correlated_data(n_samples, n_features, random_state=random_state)
+    y = np.sign(y)
+
+    np.random.seed(random_state)
```
mathurinm, line 200: usually this is frowned upo (it resets the seed globally, that's a heavy side effect) 

use rng = sklearn.utils.check_random_state(random_state), then rng.randn

skglm/datafits/group.py
Badr-MOUFAD: Agree!

**Side note**: There is a small pitfall in this case as we are using jitclasses: we can't call the constructor of the ``Logistic`` (I get a numba compilation error because of the ``super`` keyword).

skglm/tests/test_group.py
```diff
@@ -1,3 +1,4 @@
+from turtle import numinput
```
mathurinm, line 1: vscode <3

mathurinm (approved):

PR #99, opened 2022-10-18
Make l23 penalty work for Ali Rahimi

QB3:
Try to make l23 penalty work for this example https://github.com/a-rahimi/sparse-regression/blob/main/sparse%20regression.ipynb, following this thread https://twitter.com/alirahimi0/status/1581487188109594624


mathurinm:
Thanks for the investigation Quentin, we have moved it to a proper benchopt benchmark here:  https://github.com/mathurinm/benchmark_sparse

PR #113, opened 2022-10-27
FEAT Add Gamma regressor

### LINKED ISSUE #81: ENH add support for GammaRegressor
https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.GammaRegressor.html#sklearn.linear_model.GammaRegressor

given how easy it was to add Poisson datafit in #78 this may be a quick win

### LINKED ISSUE #114: BUG Raise ValueError when `y < 0` for Poisson datafit
Currently, there is no warning and the solver diverges. We should raise a value error at fit time.

PABannier:
Closes #81 
Closes #114

A few remarks:

- Solver is converging (please check the maths), but we don't have the same solution as `sklearn` for `alpha=0`. 
- `statsmodels` does not support a Gamma model so the only test I see would be to compare `sklearn` and `skglm` solutions for `alpha = 0`. WDYT?
- I've opened an issue #114  to warn the user when an incorrect value is passed to `y` for Poisson (`y>0`). Design question: where do we want to raise the ValueError? I'd go for the `value` method in datafit, to avoid having extra logic in `glm_fit`. WDYT?


skglm/datafits/single_task.py
```diff
@@ -467,3 +467,52 @@ def gradient_scalar_sparse(self, X_data, X_indptr, X_indices, y, Xw, j):
...
     def intercept_update_self(self, y, Xw):
         pass
+
+
+class Gamma(BaseDatafit):
+    r"""Gamma datafit.
+
+    The datafit reads::
+
+    (1 / n_samples) * \sum_i (Xw_i + y_i * exp(-Xw_i))
```
Badr-MOUFAD, line 477: For the maths, I looked up GammaRegressor in ``scikit-learn`` code base,
(cf. [GammaRegressor](https://github.com/scikit-learn/scikit-learn/blob/f3f51f9b611bf873bd5836748647221480071a87/sklearn/linear_model/_glm/glm.py#L679), and [HalfGammaLoss](https://github.com/scikit-learn/scikit-learn/blob/9c4f023528ca5be585c0900b224f917a7435db74/sklearn/_loss/loss.py#L671-L689))
```shell
raw_prediction_i = Xw_i
loss(x_i) = log(exp(raw_prediction_i)/y_true_i) + y_true/exp(raw_prediction_i) - 1
```
Hence, by removing the constant, the formula of ``Gamma`` datafit is correct.

toy_gamma.py
```diff
@@ -0,0 +1,29 @@
+import numpy as np
```
Badr-MOUFAD, line 1: ``statsmodels`` has a L1 gamma regressor

- [fit GLM with gamma family](https://www.statsmodels.org/stable/glm.html)
- check [``fit_regularized``](https://www.statsmodels.org/dev/generated/statsmodels.genmod.generalized_linear_model.GLM.fit_regularized.html), provide a ``L1_wt`` param

skglm/datafits/single_task.py
```diff
@@ -467,3 +467,52 @@ def gradient_scalar_sparse(self, X_data, X_indptr, X_indices, y, Xw, j):
...
+    def initialize_sparse(self, X_data, X_indptr, X_indices, y):
+        pass
+
+    def raw_grad(self, y, Xw):
+        """Compute gradient of datafit w.r.t. ``Xw``."""
+        return (Xw - y * np.exp(-Xw)) / len(y)
+
+    def raw_hessian(self, y, Xw):
+        """Compute Hessian of datafit w.r.t. ``Xw``."""
+        return (1 + y * np.exp(-Xw)) / len(y)
```
mathurinm, line 506: Shouldn't there be a minus in front of the exp ?

skglm/datafits/single_task.py
```diff
@@ -467,3 +467,52 @@ def gradient_scalar_sparse(self, X_data, X_indptr, X_indices, y, Xw, j):
...
+
+    def initialize(self, X, y):
+        pass
+
+    def initialize_sparse(self, X_data, X_indptr, X_indices, y):
+        pass
+
+    def raw_grad(self, y, Xw):
+        """Compute gradient of datafit w.r.t. ``Xw``."""
+        return (Xw - y * np.exp(-Xw)) / len(y)
```
mathurinm, line 502: So it should be 1 instead of Xw here no ?

skglm/datafits/single_task.py
mathurinm: Ok the minus is ok but the 1 may need to be 0, because in the objective it's Xw, so in gradient becomes 1, and in Hessian 0, no ?

skglm/datafits/single_task.py
mathurinm: @PABannier @Badr-MOUFAD do you think we should have the constant in the loss value ? It doesn't cost much and it harmonizes with sklearn

mathurinm (changes requested):
Thanks @PABannier !
Can you:
- add an entry to the doc 
- determine if we include the constant in the objective function (I would)
- see if we can come up with an example highlighting this and Poisson
- raise the error when calling `initialize` with a negative value in y ? It looks like the good place to do it

skglm/datafits/single_task.py
Badr-MOUFAD: I agree @mathurinm.
It intervenes in ``.value`` and we get rid of it in the formulas of grad and Hess.

skglm/datafits/single_task.py
```diff
@@ -467,3 +467,58 @@ def gradient_scalar_sparse(self, X_data, X_indptr, X_indices, y, Xw, j):
...
+    def raw_grad(self, y, Xw):
+        """Compute gradient of datafit w.r.t. ``Xw``."""
+        return (1 - y * np.exp(-Xw)) / len(y)
+
+    def raw_hessian(self, y, Xw):
+        """Compute Hessian of datafit w.r.t. ``Xw``."""
+        return (y * np.exp(-Xw)) / len(y)
+
+    def value(self, y, w, Xw):
+        return (np.sum(Xw + y * np.exp(-Xw)) - 1) / len(y)
```
Badr-MOUFAD, line 515: A ``log(y_i)`` is missing, right? 
cf. https://github.com/scikit-learn-contrib/skglm/pull/113/files#r1007810345

```suggestion
        return (np.sum(Xw + y * np.exp(-Xw)) - 1 - np.log(y)) / len(y)
```

skglm/datafits/single_task.py
Badr-MOUFAD: We modified the ``.value`` but forgot about updating that in the class docs

skglm/datafits/single_task.py
```diff
@@ -467,3 +467,58 @@ def gradient_scalar_sparse(self, X_data, X_indptr, X_indices, y, Xw, j):
...
+        pass
+
+    def get_spec(self):
+        pass
+
+    def params_to_dict(self):
+        return dict()
+
+    def initialize(self, X, y):
+        if (y <= 0).sum() > 0:
```
Badr-MOUFAD, line 495: we better ``(y <= 0).any()``,
WDY?

plot_gamma_poisson_reg.py
```diff
@@ -0,0 +1,15 @@
+"""
```
Badr-MOUFAD, line 1: This should go in the ``examples`` folder, right?

PABannier: yes, it's WIP

PABannier:
I find the example from [sklearn really well done and clear](https://scikit-learn.org/stable/auto_examples/linear_model/plot_tweedie_regression_insurance_claims.html#sphx-glr-auto-examples-linear-model-plot-tweedie-regression-insurance-claims-py). I wonder if we could not find a dataset with a lot of samples and features to illustrate the usefulness of skglm implementation of Gamma regressor. WDYT?

examples/plot_gamma_poisson_reg.py
```diff
@@ -0,0 +1,15 @@
+"""
```
Badr-MOUFAD, line 1: >I find the example from [sklearn really well done and clear](https://scikit-learn.org/stable/auto_examples/linear_model/plot_tweedie_regression_insurance_claims.html#sphx-glr-auto-examples-linear-model-plot-tweedie-regression-insurance-claims-py). I wonder if we could not find a dataset with a lot of samples and features to illustrate the usefulness of skglm implementation of Gamma regressor. WDYT?

I don't recall an example of large dataset

Badr-MOUFAD (commented):
Thanks @PABannier for the PR!

LGTM,
we shall just agree on ``examples/plot_gamma_poisson_reg.py``

mathurinm (approved):
In order to keep the momentum going I am in favor of delaying the finding of a convincing example, ie remove `plot_gamma_poisson_reg.py` and merge

PR #120, opened 2022-11-06
ENH Add `L_11` matrix norm

PABannier:
Quick PR to add the matrix L1 norm, I need it for sparse dictionary learning.

Note that in scikit-learn, the Lasso accept a target matrix and solves a multi-task objective with a L_11 penalty (see https://github.com/scikit-learn/scikit-learn/pull/24843) . As per the principle of least surprise, we should add it in the `Lasso` fit logic of skglm.

### DIFF
--- a/skglm/penalties/block_separable.py
+++ b/skglm/penalties/block_separable.py
@@ -5,7 +5,7 @@ from numba import float64, int32
 
 from skglm.penalties.base import BasePenalty
 from skglm.utils import (
-    BST, prox_block_2_05, prox_SCAD, value_SCAD, prox_MCP, value_MCP)
+    BST, ST_vec, prox_block_2_05, prox_SCAD, value_SCAD, prox_MCP, value_MCP)
 
 
 class L2_1(BasePenalty):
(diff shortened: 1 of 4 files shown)


mathurinm:
I agree with the principle of least surprise, but our experience is also that trying to follow sklearn increases the maintenance cost on our side, so I'm not sure there's an easy decision here.

skglm/tests/test_penalties.py
```diff
@@ -101,5 +103,15 @@ def test_slope():
     np.testing.assert_allclose(ours.coef_, pyslope_out["beta"], rtol=1e-5)
 
 
+def test_l11():
+    ours = GeneralizedLinearEstimator(
+        datafit=QuadraticMultiTask(),
+        penalty=L1_1(alpha),
+        solver=MultiTaskBCD(tol=tol, ws_strategy="fixpoint", max_iter=10),
+    ).fit(X, Y)
+    sk = Lasso(alpha, tol=tol, fit_intercept=False).fit(X, Y)
```
mathurinm, line 112: Don't you need to fit task per task?

PABannier: my bad, i wanted to compare against `Lasso`  from `sklearn`

skglm/penalties/block_separable.py
```diff
@@ -330,3 +330,43 @@ def generalized_support(self, w):
...
+    def subdiff_distance(self, W, grad, ws):
+        """Compute distance of negative gradient to the subdifferential at W."""
+        subdiff_dist = np.zeros_like(ws, dtype=grad.dtype)
+        for idx, j in enumerate(ws):
+            if not np.any(W[j, :]):
+                # distance of - grad_j to alpha * the unit l\infty ball
+                norm_grad_j = np.max(np.abs(grad[idx, :]))
+                subdiff_dist[idx] = max(0, norm_grad_j - self.alpha)
+            else:
+                subdiff_dist[idx] = norm(grad[idx, :] + self.alpha * np.sign(W[j, :]))
```
PABannier, line 367: I've done the computations thrice, but I'm not able to find my mistake. I'd be happy to have feedbacks ;) @mathurinm

PABannier: I'm confused then. I thought :math`||A||_{1, 1} = \sum_{i=1}^n ||A_i||_1 = \sum_{i=1}^n \sum_{j=1}^p |A_{i,j}|`. As per the  first equality, I thought we could cast it in `skglm` BCD framework. So maybe, I should just loop over `subdiff_dist` and compute the 1-d subdiff_dist for every feature.

mathurinm: Yes, but our working sets are defined row-wise (we perform BCD over the rows). Here since no structure is imposed, it's likely that there will often be at least one non zero coef per row, so working sets will consider the full matrix.

I am -1 on merging this because it would increase the size of the codebase. I believe this problem can be addressed by fitting several Lassos instead.

mathurinm: @QB3 thoughts?

PABannier:
Closing since it is not compatible with our working set algorithm.

PR #126, opened 2022-11-20
FEAT Add positivity penalty

PABannier:


skglm/penalties/separable.py
```diff
@@ -466,3 +466,48 @@ def is_penalized(self, n_features):
     def generalized_support(self, w):
         """Return a mask with non-zero coefficients."""
         return w != 0
+
+
+class NNLS(BasePenalty):
+    """Non-negative least square penalty."""
+
+    def __init__(self):
```
Badr-MOUFAD, line 474: to remove?

skglm/penalties/separable.py
```diff
@@ -466,3 +466,48 @@ def is_penalized(self, n_features):
...
+        pass
+
+    def get_spec(self):
+        return ()
+
+    def params_to_dict(self):
+        return dict()
+
+    def value(self, w):
```
Badr-MOUFAD, line 483: I would rather put in docs ``Compute the value of the NNLS penalty at w.`` (like the other penalties)
also, I think  "positive weights" might be a bit confusing  

is there a reason for not putting the explicit implementation of the penalty val?
```python
if (w  < 0).any():
    return np.inf
else:
    return 0
```

Badr-MOUFAD (commented):
GG @PABannier for the PR!

some minor remarks.

skglm/penalties/separable.py
```diff
@@ -466,3 +466,48 @@ def is_penalized(self, n_features):
     def generalized_support(self, w):
         """Return a mask with non-zero coefficients."""
         return w != 0
+
+
+class NNLS(BasePenalty):
```
mathurinm, line 471: LS here means least squares, but we have a penalty, independent of datafit.

Should the name be "PositiveConstraint" ?

skglm/penalties/separable.py
PABannier: can't be removed otherwise numba throws an error during JIT compilation

skglm/penalties/separable.py
```diff
@@ -466,3 +466,48 @@ def is_penalized(self, n_features):
     def generalized_support(self, w):
         """Return a mask with non-zero coefficients."""
         return w != 0
+
+
+class PositiveConstraint(BasePenalty):
+    """Non-negative least square penalty."""
```
Badr-MOUFAD, line 472: Since we changed the name, we better put ``Positive constraint penalty.`` ?

mathurinm:
```suggestion
    """Positivity constraint penalty."""
```

mathurinm (approved):

mathurinm:
Thanks @PABannier @Badr-MOUFAD, merge when green !
Next step: add it to benchopt/benchmark_nnls

PR #137, opened 2022-12-10
ENH - check ``datafit + penalty`` compatibility with solver

### LINKED ISSUE #90: ENH Inform user when `ws_strategy = subdiff` is not available
It is not obvious that `ws_strategy=subdiff` won't work for some penalties (e.g. L0.5, L2/3, ...). Throwing an error and issuing a warning in this situation would prevent users from spending time figuring out the issue.

### LINKED ISSUE #101: ENH Create a solver dispatcher
Currently, users are not explicitly warned when an unsupported solver is used with some penalties or datafits. We should write some form of mechanism (a dispatcher?) which raises an exception when the coupling (datafit, penalty, solver) is unsupported.

E.g.: for #100 currently we should raise an exception, that is explicit for the user

### LINKED ISSUE #109: ENH warn/raiseError in WS solvers if penalty is L05 or L23 for ws_strategy='subdiff'
This will cause the solver not to move since 0 is always a critical point

PABannier:
A quick proof-of-concept of a function that checks if the combination `(solver, datafit, penalty)` is supported. Currently we have some edge cases where one can pass `ProxNewton` solver with `L0_5` penalty without any error being raised.

Pros of this design: the validation rules are centralized and validating a 3-uple is a one-liner in `glm_fit`.
Cons: we have to update the rules as we enhance the capabilities of the solver.

All in all, I think it is very valuable to have more verbose errors when fitting estimators (e.g. Ali Rahimi initially passed a combination Quadratic, L2_3, ProxNewton which cannot be optimized at the moment of writing).

Closes #101
Closes #90 
Closes #109


PABannier:
@mathurinm Yes I think it's cleaner, currently refining the POC.

mathurinm:
This would be a nice addition if we can ship it in the 0.3 release @Badr-MOUFAD , given that we added a few datafits, penalties and solvers !

Badr-MOUFAD:
> @Badr-MOUFAD the issue popped up in https://github.com/scikit-learn-contrib/skglm/issues/188, do you have time to take this over ? A simple check, at the beginning of each solver, that the datafit and penalty are supported (eg AndersonCD does not support Gamma datafit)

Sure, I will resume this PR.

mathurinm:
Requires #191 to be implemented fit to allow for better checks

QB3 (commented):
What a massive piece of work! Congrats @PABannier @Badr-MOUFAD !

My complain would be that I did not fully understood the need of calling multiple functions in the `_validate` function in `solvers/base.py`.

skglm/penalties/non_separable.py
```diff
@@ -48,13 +48,3 @@ def prox_vec(self, x, stepsize):
         prox[sorted_indices] = prox_SLOPE(abs_x[sorted_indices], alphas * stepsize)
 
         return np.sign(x) * prox
-
-    def prox_1d(self, value, stepsize, j):
-        raise ValueError(
```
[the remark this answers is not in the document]
Badr-MOUFAD: That is a fair question,
It was a small typo (`return` ---> `raise`) that slipped through after adding SLOPE in PR https://github.com/scikit-learn-contrib/skglm/pull/92

Now this case is handled in the validation, thus we no longer need to explicitly raise errors.

skglm/solvers/base.py
```diff
@@ -39,3 +57,45 @@ def solve(self, X, y, datafit, penalty, w_init, Xw_init):
...
+        --------
+        >>> ...
+        >>> coefs, obj_out, stop_crit = solver(X, y, datafit, penalty)
+        """
+        self._validate(X, y, datafit, penalty)
+        return self.solve(X, y, datafit, penalty, w_init, Xw_init)
+
+    def _validate(self, X, y, datafit, penalty):
+        # execute both: attributes checks and `custom_compatibility_check`
+        self.custom_compatibility_check(X, y, datafit, penalty)
```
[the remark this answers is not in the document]
Badr-MOUFAD: > Does one check generic compatibility and the other check custom compatibility?

Yes indeed. Generic checks are checks for datafit/penalty implement the required attributes in ``_datafit_required_attr``/``_penalty_required_attr``.
On the other hand, ``custom_compatibility_check`` does other checks, for instance in GramCD it checks that datafit is instance of ``Quadratic``

skglm/solvers/fista.py
```diff
@@ -111,3 +108,18 @@ def solve(self, X, y, datafit, penalty, w_init=None, Xw_init=None):
...
+    def custom_compatibility_check(self, X, y, datafit, penalty):
+        # check datafit support sparse data
+        check_obj_solver_attr(
+            datafit, solver=self,
+            required_attr=self._datafit_required_attr,
+            support_sparse=issparse(X)
+        )
+
+        # optimality check
+        if self.opt_strategy == "subdiff" and not hasattr(penalty, "subdiff_distance"):
```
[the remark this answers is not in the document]
QB3: I see. Anyway, I think the cleanest would be to factorize the working sets themselves. But this is unclear we want to do it, and even if we want to do it, this would be for another PR.

skglm/solvers/base.py
```diff
@@ -1,8 +1,26 @@
...
 
 
-class BaseSolver():
-    """Base class for solvers."""
+class BaseSolver(ABC):
+    """Base class for solvers.
+
+    Attributes
+    ----------
+    _datafit_required_attr : list
```
mathurinm, line 10: I think we can make these public @BadrMoufad ? Maybe I'm missing a specific reason

skglm/solvers/base.py
```diff
@@ -39,3 +57,45 @@ def solve(self, X, y, datafit, penalty, w_init, Xw_init):
...
+
+        This method includes extra checks to perform
+        aside from checking attributes compatibility.
+
+        Parameters
+        ----------
+        X : array, shape (n_samples, n_features)
+            Training data.
+
+        y : array, shape (n_samples,)
```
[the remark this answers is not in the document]
mathurinm: I agree with @Badr-MOUFAD here, even if not needed at the moment it's not too hard to see cases where this would happen, and an API change will be painful

skglm/utils/validation.py
```diff
@@ -23,3 +28,68 @@ def check_group_compatible(obj):
...
+    Raises
+    ------
+        AttributeError
+            if any of the attribute in ``required_attr`` is missing
+            from ``obj`` attributes.
+    """
+    missing_attrs = []
+    suffix = SPARSE_SUFFIX if support_sparse else ""
+
+    # if `attr` is a list check that at least one of them
```
mathurinm, line 59: why "at least one of them" ?

skglm/experimental/pdcd_ws.py
```diff
@@ -88,10 +93,10 @@ def __init__(self, max_iter=1000, max_epochs=1000, dual_init=None,
         self.verbose = verbose
 
     def solve(self, X, y, datafit_, penalty_, w_init=None, Xw_init=None):
-        if issparse(X):
-            raise ValueError("Sparse matrices are not yet support in PDCD_WS solver.")
+        # jit compile classes
+        datafit = compiled_clone(datafit_)
```
mathurinm, line 97: I think this is done only for this solver, is there a particular reason?

skglm/solvers/base.py
mathurinm: also typo missing "that must BE" here and below

skglm/solvers/base.py
```diff
@@ -1,8 +1,26 @@
...
+    ----------
+    _datafit_required_attr : list
+        List of attributes that must implemented in Datafit.
+
+    _penalty_required_attr : list
+        List of attributes that must implemented in Penalty.
+
+    Notes
+    -----
+    For required attributes, if an attribute is given as a list of attributes
```
mathurinm, line 18: I had trouble understanding this, an example help (in which case do we want to check that one of several attributes is present?)

skglm/solvers/base.py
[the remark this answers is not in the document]
mathurinm: one of confusions for me is from the name, `check_obj_solver_attr` could be just `check_attr`; we pass `solver` just to get its name, though useful I don't think it is worth it.

skglm/solvers/fista.py
```diff
@@ -27,6 +28,9 @@ class FISTA(BaseSolver):
            https://epubs.siam.org/doi/10.1137/080716542
     """
 
+    _datafit_required_attr = ("get_global_lipschitz", ("gradient", "gradient_scalar"))
+    _penalty_required_attr = (("prox_1d", "prox_vec"),)
```
mathurinm, line 32: why does FISTA need prox_1D, it is not used in the code below

skglm/solvers/gram_cd.py
```diff
@@ -132,6 +137,12 @@ def solve(self, X, y, datafit, penalty, w_init=None, Xw_init=None):
             p_objs_out.append(p_obj)
         return w, np.array(p_objs_out), stop_crit
 
+    def custom_compatibility_check(self, X, y, datafit):
+        if not isinstance(datafit, Quadratic):
+            raise AttributeError(
+                f"`GramCD` supports only `Quadratic` datafit, got {datafit}"
```
mathurinm, line 143: this is very clean

skglm/solvers/multitask_bcd.py
```diff
@@ -229,6 +233,21 @@ def path(self, X, Y, datafit, penalty, alphas, W_init=None, return_n_iter=False)
...
+            datafit, solver=self,
+            required_attr=self._datafit_required_attr,
+            support_sparse=sparse.issparse(X)
+        )
+
+        # ws strategy
+        if self.ws_strategy == "subdiff" and not hasattr(penalty, "subdiff_distance"):
+            raise AttributeError(
+                "Penalty must implement `subdiff_distance` "
+                "to use self.ws_strategy='subdiff'."
```
mathurinm, line 248: remove `self.`

skglm/solvers/prox_newton.py
```diff
@@ -193,6 +196,14 @@ def solve(self, X, y, datafit, penalty, w_init=None, Xw_init=None):
             )
         return w, np.asarray(p_objs_out), stop_crit
 
+    def custom_compatibility_check(self, X, y, datafit, penalty):
+        # ws strategy
+        if self.ws_strategy == "subdiff" and not hasattr(penalty, "subdiff_distance"):
+            raise AttributeError(
+                "Penalty must implement `subdiff_distance` "
+                "to use self.ws_strategy='subdiff'."
```
mathurinm, line 204: remove `self.`

skglm/utils/validation.py
```diff
@@ -23,3 +28,68 @@ def check_group_compatible(obj):
...
+        required_attr = [_join_attrs_with_or(attrs, suffix) for attrs in required_attr]
+
+        # get name obj and solver
+        name_matcher = re.compile(r"\.(\w+)'>")
+
+        obj_name = name_matcher.search(str(obj.__class__)).group(1)
+        solver_name = name_matcher.search(str(solver.__class__)).group(1)
+
+        raise AttributeError(
+            f"{obj_name} is not compatible with {solver_name}"
```
mathurinm, line 80: I find this hard to read, and I don't get why we prefix the sentence with a `.` is not `support_sparse`.
Two separate ifs are less fancy but may be more readable.

mathurinm:
Trying to revive this to release `v0.4`. Some comments for discussion @QB3 @Badr-MOUFAD 

- I feel strongly against having both `solver.solve()` and `solver()`. "There should be one-- **and preferably only one** --obvious way to do it". 
The way I see it, all `solve` methods should be renamed `_solve`, and `BaseSolver` should implement its `solve` that does the checks then call `_solve`. What do you think? This is non-breaking from the API point of view, just requires people implementing their own solver to adapt (I have no such example in mind ; we can implement a test that no Solver implements solve by checking that A.solve == BaseSolver.solve) 
`BaseSolver.solve` can have a `run_checks` method which is True by default, and disabled if one wants to be faster (any idea on the impact of the checks @Badr-MOUFAD ?)
- I feel like checking for the sparse support is heavy at the moment and could be done in another PR, but that may just be me :)  
- I have mixed feelings for the name `check_obj_solver_attr`; see my comment, I'd go for `check_attr` and not pass `solver`, which is used only for its name.
- `custom_compatibility_check` could be `custom_checks` (?)

skglm/solvers/base.py
Badr-MOUFAD: > I think we can make these public @BadrMOUFAD ? Maybe I'm missing a specific reason

I think these two attributes should be read only.

While there is a way to make attributes read only, namely using [the property decorator](https://docs.python.org/3/library/functions.html#property), I believe it adds two much complexity to the code and hence doesn’t serve our goal to make components implementation user-friendly.

I opted for the “start with underscore” naming convention to make variables private to signal to the user that these are attributes to not mess up with.

skglm/solvers/base.py
Badr-MOUFAD: I 100% agree with you @mathurinm, I should have accompanied the docs with an example

In Fista solver, this

```python
    _datafit_required_attr = ("get_global_lipschitz", ("gradient", "gradient_scalar"))
    _penalty_required_attr = (("prox_1d", "prox_vec"),)
```

is interpreted as
- datafit is required to have ``get_global_lipschitz`` and (``gradient`` or ``gradient_scalar``)
- penalty is required to have  ``prox_1d`` or ``prox_vec``

This is the way I implemented ``check_obj_solver_attr`` function: whenever attributes are wrapped in parenthesis, it is interpreted as the “or” operator and a comma is interpreted the “and” operator.

skglm/solvers/fista.py
Badr-MOUFAD: It is used in ``_prox_vec`` which take as argument the penalty (cf. https://github.com/PABannier/skglm/blob/e687bc2ecaacfe920b9aaad3e33e1f0cbdbac683/skglm/solvers/fista.py#L83)

The algorithm works if penalty has either of ``prox_1d`` or  ``prox_vec``. 
(for reference https://github.com/scikit-learn-contrib/skglm/pull/137#discussion_r1613225915)

skglm/experimental/pdcd_ws.py
Badr-MOUFAD: Yes you are right.

I removed it. 
One thing is that is might have some side effects (for user) as the compilation was done in ``_validate_init``.

Badr-MOUFAD:
- I’m +1 with having ``_solve`` method and adding ``run_checks`` argument to ``solve`` method.
I feel it give us more freedom to standardize the behavior and make less verbose the ``solve`` method.
I don’t think the check have a big overhead, though I didn’t check that in practice
- The support of sparse data is covered in ``check_obj_solver_attr`` function, so I don’t think it hurts us to cover it  in this PR
-  I have no strong opion about the names, I’m +1 with your proposed name @mathurinm

mathurinm:
I think the error messages help a lot. 
Suggestion: add datafit/penalty and sovler

from 
```
AttributeError: Poisson is not compatible with AndersonCD. It must implement `get_lipschitz` and `gradient_scalar`
Missing `get_lipschitz`.
```
to 
```
AttributeError: Datafit `Poisson` is not compatible with solver `AndersonCD`. It must implement `get_lipschitz` and `gradient_scalar`
Missing `get_lipschitz`.
```

skglm/experimental/tests/test_sqrt_lasso.py
```diff
@@ -69,7 +70,10 @@ def test_PDCD_WS(with_dual_init):
 
     dual_init = y / norm(y) if with_dual_init else None
 
-    w = PDCD_WS(dual_init=dual_init).solve(X, y, SqrtQuadratic(), L1(alpha))[0]
+    datafit = compiled_clone(SqrtQuadratic())
+    penalty = compiled_clone(L1(alpha))
+
+    w = PDCD_WS(dual_init=dual_init).solve(X, y, datafit, penalty)[0]
```
mathurinm, line 76: how was this working before if the datafit and penalties were not compiled ?

Badr-MOUFAD: They were compiled inside the solver.

skglm/solvers/base.py
```diff
@@ -1,8 +1,26 @@
...
+    ----------
+    _datafit_required_attr : list
+        List of attributes that must be implemented in Datafit.
+
+    _penalty_required_attr : list
+        List of attributes that must be implemented in Penalty.
+
+    Notes
+    -----
+    For required attributes, if an attribute is given as a list of attributes
```
mathurinm, line 18: Can you be a bit more explciit like : "if an element of the required attribute list is itself a list, it means that at least of theses attributes must be implemented
    e.g. `_datafit_required_attr = ("a", ("b", "c"))` means that the datafit must implement `a`, together with at least one of `b` and `c`"  ?

skglm/solvers/base.py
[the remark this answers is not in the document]
mathurinm: Maybe (as per @QB3 remark) harmonizing the names would help: 
```check_attributes` and `check_custom_comaptibility` ? not really sure...

File: skglm/solvers/fista.py (hunk 9 above)
Comment (mathurinm) [reply skglm/solvers/fista.py:32]: got it, thanks !

File: skglm/solvers/fista.py [hunk 16]
```diff
@@ -111,3 +108,18 @@ def solve(self, X, y, datafit, penalty, w_init=None, Xw_init=None):
...
+            datafit, solver=self,
+            required_attr=self._datafit_required_attr,
+            support_sparse=issparse(X)
+        )
+
+        # optimality check
+        if self.opt_strategy == "subdiff" and not hasattr(penalty, "subdiff_distance"):
+            raise AttributeError(
+                "Penalty must implement `subdiff_distance` "
+                "to use self.opt_strategy='subdiff'."
```
mathurinm, line 124:
```suggestion
                "to use `opt_strategy='subdiff'` in Fista solver."
```

skglm/solvers/gram_cd.py
mathurinm, line 143:
```suggestion
                f"`GramCD` supports only `Quadratic` datafit, got {datafit}."
```

skglm/solvers/base.py
Badr-MOUFAD: Yes indeed

skglm/solvers/gram_cd.py
```diff
@@ -132,6 +137,12 @@ def solve(self, X, y, datafit, penalty, w_init=None, Xw_init=None):
             p_objs_out.append(p_obj)
         return w, np.array(p_objs_out), stop_crit
 
+    def custom_compatibility_check(self, X, y, datafit):
```
mathurinm: I just fixed the number of argument (penalty was missing)

skglm/solvers/gram_cd.py
```diff
@@ -132,6 +137,12 @@ def solve(self, X, y, datafit, penalty, w_init=None, Xw_init=None):
             p_objs_out.append(p_obj)
         return w, np.array(p_objs_out), stop_crit
 
+    def custom_compatibility_check(self, X, y, datafit, penalty):
+        if not isinstance(datafit, Quadratic):
+            raise AttributeError(
+                f"`GramCD` supports only `Quadratic` datafit, got {datafit}."
```
mathurinm, line 143:
```suggestion
                f"`GramCD` supports only `Quadratic` datafit, got {datafit.__class__.__name__}."
```
otherrwise error message is: 
`AttributeError: `GramCD` supports only `Quadratic` datafit, got <numba.experimental.jitclass.boxing.Poisson object at 0x7f4bb1fa3f40>`

skglm/utils/validation.py
mathurinm: ok got it thanks to your expalnations @Badr-MOUFAD , maybe explain with a bit more words here too

skglm/solvers/prox_newton.py
```diff
@@ -193,6 +196,14 @@ def solve(self, X, y, datafit, penalty, w_init=None, Xw_init=None):
             )
         return w, np.asarray(p_objs_out), stop_crit
 
+    def custom_compatibility_check(self, X, y, datafit, penalty):
+        # ws strategy
+        if self.ws_strategy == "subdiff" and not hasattr(penalty, "subdiff_distance"):
+            raise AttributeError(
+                "Penalty must implement `subdiff_distance` "
+                "to use ws_strategy='subdiff'."
```
mathurinm, line 204: in proxNewton solver

mathurinm:
> * I’m +1 with having `_solve` method and adding `run_checks` argument to `solve` method.
>       I feel it give us more freedom to standardize the behavior and make less verbose the `solve` method.
>       I don’t think the check have a big overhead, though I didn’t check that in practice

Let's perform checks all the time for now, this will simplify our lives. WDYT?

>     * The support of sparse data is covered in `check_obj_solver_attr` function, so I don’t think it hurts us to cover it  in this PR
OK
> 
>     * I have no strong opion about the names, I’m +1 with your proposed name @mathurinm
@QB3 any opinion?

Badr-MOUFAD (approved):
LGTM, thanks for everyone 💪 

We should be careful in internalizing the jit-compilation of dataffits and penalties as jit-compilation undoes the datafit initialization.

I believe this PR brings several contributions to the API and touches many parts of the codebase. we better merge it and tackle the aforementioned issue in that in a separate PR.

Issue #144, opened 2023-02-11
DOC Objectives docstring do not use math mode


KishManani:
See: https://contrib.scikit-learn.org/skglm/generated/skglm.datafits.Poisson.html#skglm.datafits.Poisson

<img width="800" alt="image" src="https://user-images.githubusercontent.com/30973056/218250470-a147da51-f777-49fb-8ebe-e3670133c733.png">

mathurinm:
Hi @KishManani, thanks for letting us know. Do you think you'd be able to send a PR fixing this, for example by looking at how it's fine in other classes where the rendering is correct ?

PR #146, opened 2023-03-01
DOC use math mode in all docstrings

borispf:
This changes the rendering for the objective to math mode.

There were no other autogenerated doc pages with math mode, so I tried to change as little as possible about the formatting. One style question: should `\texttt{n_samples}` be changed to `\text{n_samples}` just `n`?

I can change the other objectives and other equations in the docs to math mode either in this PR or a following one.

---

[CircleCI artifact](https://output.circle-artifacts.com/output/job/14fe9193-ff80-4c60-86af-151a77475793/artifacts/0/dev/generated/skglm.datafits.Poisson.html)

Local preview:

Screenshot 2023-03-01 at 16 59 41

Compare with #144 

image


mathurinm:
Thanks for the PR @borispf !
FYI once the github actions are allowed to run (which needs to be done manually because you are a first time contributor), everyone can directly check the rendered doc as output of the `build doc artifact` job : https://output.circle-artifacts.com/output/job/14fe9193-ff80-4c60-86af-151a77475793/artifacts/0/dev/generated/skglm.datafits.Poisson.html#skglm.datafits.Poisson


I'm ok for the change, with two reservations:
- sklearn does not use math mode https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.Ridge.html
- when reading the docstring directly inside a terminal, the current syntax is more readable than fancy math mode 


Thoughts @QB3 @PABannier @Badr-MOUFAD ? Beauty of website doc beats interpreter doc ?


For the n_samples I am in favor of not using n (n_samples is more explicit, notation differ across fields) but it does not need to be in texttt

Badr-MOUFAD:
> Thoughts @QB3 @PABannier @Badr-MOUFAD ? Beauty of website doc beats interpreter doc ?

I am with using math mode (so with the beauty of the doc :)).
Though it might make the expression a bit obscure to read (code-wise), It's more clear for the user.

> For the n_samples I am in favor of not using n ...

We can use `n` to make the expression more concise and math conformist, but I have no strong opinion about that.

PABannier:
Thanks for the PR @borispf 

1. I think the goal of a doc is to be quickly readable and clear. The current design could confuse users/readers. I prefer the maths mode even if it does not strictly follows sklearn rules. 
2. I agree with @mathurinm : `n` is confusing. We should aim for clarity: `n_samples` is self-explanatory, `n` is not. I'd stick to `n_samples`.

mathurinm:
Thanks for the feedback. So using math mode it can be applied in all docstrings in this PR @borispf !

skglm/datafits/single_task.py
```diff
@@ -423,9 +423,9 @@ def intercept_update_step(self, y, Xw):
 class Poisson(BaseDatafit):
     r"""Poisson datafit.
 
-    The datafit reads::
+    The datafit reads:
 
-    (1 / n_samples) * \sum_i (exp(Xw_i) - y_i * Xw_i)
+    .. math:: \frac{1}{\texttt{n_samples}} \sum_i (\exp(Xw_i) - y_i Xw_i)
```
mathurinm, line 428: Xw_i migh be ambiguous. Using (Xw)_i would be preferrable.
While we're at it, using \sum_{i=1}^{n_\mathrm{n_samples}} would be more explicit too

borispf:
Thanks for the feedback @mathurinm @Badr-MOUFAD @PABannier! I was poking around MathJax and it supports [AsciiMath](http://asciimath.org) as an [input format](https://docs.mathjax.org/en/latest/input/asciimath.html). Sphinx can be configured to spit out the asciimath delimiters instead of the latex math ones, but it would affect all math in the documentation and break output formats other than html.

Thoughts?

---

Example of the poisson objective in asciimath:

    (1 / n_"samples") * \sum_(i=1)^(n_"samples") (exp((Xw)_i) - y_i * (Xw)_i)

Screenshot 2023-03-02 at 12 25 18

mathurinm:
Thanks! I find the syntax way more readable than latex (quotes instead of mathrm, no need for mathrm in log, no frac). 

Do we need output in other formats than HTML? If not I'd say go for it

The rendering seems to go out of the box, is that a glitch?

borispf:
> Do we need output in other formats than HTML? If not I'd say go for it

If you need other formats we can probably use [py-asciimath](https://pypi.org/project/py-asciimath/)
and a sphinx plugin to convert the asciimath to latex during compilation. I don’t have any experience with sphinx plugins but I can look into it

> The rendering seems to go out of the box, is that a glitch?

I think that’s just a glitch on the asciimath.org website. Other examples I’ve seen with asciimath + MathJax have  identical output to latex + MathJax.

mathurinm:
It's good to know that we have the option if we need latex later, but at the moment we don't so let's go with asciimath in all the docstrings !

mathurinm:
Thanks for the meticulous work @borispf! 

Discussion: I think we can get rid of the dots here, and replace the sum by a $||w||_1$.
image

In settings such as the enet, some dots are necessary (l1_ratio times alpha times ||w||_1) but I tend to prefer x signs in this case. WDYT?
image

borispf:
Thanks for fixing up the linter errors!

> Thanks for the meticulous work @borispf!
> 
> Discussion: I think we can get rid of the dots here, and replace the sum by a ||w||1. image
> 
> In settings such as the enet, some dots are necessary (l1_ratio times alpha times ||w||_1) but I tend to prefer x signs in this case. WDYT? image

My intention was to change as little as possible to begin with. But I also think a lot of the dots are unnecessary and where they help clarity the x signs are better. I'll go through and make those changes.

borispf:
Okay, I went through and removed `*` between single letter variables, and replaced them with `xx` between long variables. I also replaced any norm related sum with a `norm(.)_[1/2]` or `||.||_[1/2]`, and fixed up any slight inconsistencies I stumbled upon.

It's ready for another look @mathurinm

mathurinm:
very minor comment, there is an extra space when using norm (datafit) and not when using || (penalty): 
image
 
 
 It doesn't bother me too much but maybe introducing `norm` is not needed
 Apart form that I'm good !

borispf:
> very minor comment, there is an extra space when using norm (datafit) and not when using || (penalty): image

I agree that it looks a little weird. I've changed all the `norm(.)`s to `||.||`s since it didn't take much time and looks better 😄

Badr-MOUFAD (approved):
thanks @borispf for this amazing work :rocket: 

LGTM!

mathurinm:
Thanks @borispf, keep us posted if there are other issues that you'd like to work on and on which we could provide guidance

PR #154, opened 2023-04-24
MNT - remove attribute ``XTy`` from ``Quadratic`` datafit

Badr-MOUFAD:
This removes the attribute ``XTy`` from ``Quadratic`` datafit.
It does the same for Multitask version.

### DIFF
--- a/skglm/datafits/multi_task.py
+++ b/skglm/datafits/multi_task.py
@@ -27,7 +27,6 @@ class QuadraticMultiTask(BaseMultitaskDatafit):
 
     def get_spec(self):
         spec = (
-            ('XtY', float64[:, :]),
             ('lipschitz', float64[:]),
         )
         return spec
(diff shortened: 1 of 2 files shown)


PABannier (commented):
Ok for me, but why do we remove this? No speed ups in practice?

Badr-MOUFAD:
> From the theoretical point of view, this adds O(n) to the computation of the gradient in the dense case. Doesn't it degrade performance in practice ?

Right @mathurinm! It does so, though I haven't checked that through a benchmark.
I believe the standardization and modularity advantages would pay back.

> Ok for me, but why do we remove this? No speed ups in practice?

I set up a [milestone](https://github.com/scikit-learn-contrib/skglm/milestone/2) describing the motivations behind changes. I'm open to discussions.

Badr-MOUFAD:
Addressed accordingly in #192

PR #124, opened 2022-11-17
ENH Add Cox loss

PABannier:
This PR implements the loss of Cox proportional hazards model (both Berlow's and Efron's estimate for tied times).

- [ ] Implement Berlow's estimate
- [ ] Implement Efron's estimate
- [ ] Write example / test comparison against `lifelines` `CoxPHFitter` 
- [ ] Build a `CoxPHEstimator` class
- [ ] Write an example on a large RNAseq data

Closes #121

PR #157, opened 2023-05-17
ENH - implement Cox estimator

Badr-MOUFAD:
This adds a Cox estimator. It proceeds by

- [ ] implementing a Cox datafit using Breslow estimate
- [ ] Derive Cox estimator from it
- [ ] unittest/ benchmark against ``R`` survival package

## Problem setup

This problems falls in survival analysis setup. Given

- $\mathbf{X} \in \mathbb{R^{n \times p}}$ a matrix of $p$ predictors $X_j$ and $n$ samples $x_i$
- $y \in \mathbb{R}^n$ a vector recording the times of events occurrence
- $s \in \{ 0, 1 \}^n$ a binary vector where $1$ means *event occurred*

The $-\log$-likelihood to be minimized reads [1]

$$
l(\beta) = \sum_{i=1}^{n} -s_i \langle x_i, \beta \rangle + \textstyle\sum_{y_j \geq y_i} e^{\langle x_j, \beta \rangle}
$$

which represent the datafit to be considered.

## Datafit expression

Defining $b \in \mathbb{R}^n$ with 

$$b_i = \sum_{l=1}^{n} s_l \delta_{y_i \geq y_l}$$

the datafit can be rewritten as

$$
\begin{align*}
l(\beta) & = - \langle s, \mathbf{X} \beta \rangle + \langle b, e^{\mathbf{X} \beta} \rangle \\\\ & = F(\mathbf{X} \beta)
\end{align*}
$$

with $F(u) = - \langle s, u \rangle + \langle b, e^{u} \rangle$. We can establish

- $\nabla F(u) = -s + b \odot e^u$
- $\nabla^2 F(u) = \mathrm{diag}(b \odot e^u)$

## Cox Estimator

Based on the that and referring to ``skglm`` toolbox, Cox estimator can be fit using a ``ProxNewton`` solver.

----
**Note**

Revival of #124

**References**

[1] Lin, D. Y. "On the Breslow estimator." Lifetime data analysis 13 (2007): 471-480.


Badr-MOUFAD:
@PABannier, WDYT about the input conversion (`X`, `y`, `s`)?

Also, I would be grateful if you could check the datafit expression :pray:

PABannier:
The input conversion is sound and greatly simplifies the expression of the datafit. However I'm not sure about Breslow's estimate. Could you compare it with lifeline's or PySurvival? Make sure you don't have tied ties otherwise Efron's and Breslow's estimates are different.

Badr-MOUFAD:
When unit testing against ``lifeline``

1. I don't get the same solution as ``lifeline`` (Perhaps become of the optimization method ?)
2. The objective values mismatch (with `skglm` having the lowest)

I pushed a `debug_script` where I fit a Cox estimator on a dummy dataset with different occurrence times to avoid data ties.

@PABannier, @mathurinm, any thoughts?

mathurinm:
The script fails on my machine with:
```python
In [1]: %run debug_cox_lifeline.py
---------------------------------------------------------------------------
TypingError                               Traceback (most recent call last)
File ~/workspace/skglm/debug_cox_lifeline.py:33
     30 datafit = compiled_clone(Cox())
     31 penalty = compiled_clone(L1(alpha))
---> 33 datafit.initialize(X, (tm, s))
     35 w, _, _ = ProxNewton(fit_intercept=False).solve(
     36     X, (tm, s), datafit, penalty
     37 )
     39 # fit lifeline estimator

File ~/mambaforge/lib/python3.10/site-packages/numba/experimental/jitclass/boxing.py:61, in _generate_method.<locals>.wrapper(*args, **kwargs)
     59 @wraps(func)
     60 def wrapper(*args, **kwargs):
---> 61     return method(*args, **kwargs)

File ~/mambaforge/lib/python3.10/site-packages/numba/core/dispatcher.py:468, in _DispatcherBase._compile_for_args(self, *args, **kws)
    464         msg = (f"{str(e).rstrip()} \n\nThis error may have been caused "
    465                f"by the following argument(s):\n{args_str}\n")
    466         e.patch_message(msg)
--> 468     error_rewrite(e, 'typing')
    469 except errors.UnsupportedError as e:
    470     # Something unsupported is present in the user code, add help info
    471     error_rewrite(e, 'unsupported_error')

File ~/mambaforge/lib/python3.10/site-packages/numba/core/dispatcher.py:409, in _DispatcherBase._compile_for_args.<locals>.error_rewrite(e, issue_type)
    407     raise e
    408 else:
--> 409     raise e.with_traceback(None)

TypingError: Failed in nopython mode pipeline (step: nopython frontend)
- Resolution failure for literal arguments:
Failed in nopython mode pipeline (step: nopython frontend)
No implementation of function Function(<built-in function getitem>) found for signature:

 >>> getitem(array(float64, 1d, C), Tuple(slice<a:b>, none))

There are 22 candidate implementations:
  - Of which 20 did not match due to:
  Overload of function 'getitem': File: <numerous>: Line N/A.
    With argument(s): '(array(float64, 1d, C), Tuple(slice<a:b>, none))':
   No match.
  - Of which 2 did not match due to:
  Overload in function 'GetItemBuffer.generic': File: numba/core/typing/arraydecl.py: Line 166.
    With argument(s): '(array(float64, 1d, C), Tuple(slice<a:b>, none))':
   Rejected as the implementation raised a specific error:
     NumbaTypeError: unsupported array index type none in Tuple(slice<a:b>, none)
  raised from /home/mathurin/mambaforge/lib/python3.10/site-packages/numba/core/typing/arraydecl.py:72

During: typing of intrinsic-call at /home/mathurin/workspace/skglm/skglm/datafits/single_task.py (593)
During: typing of static-get-item at /home/mathurin/workspace/skglm/skglm/datafits/single_task.py (593)

File "skglm/datafits/single_task.py", line 593:
    def initialize(self, X, y):
        <source elided>
        tm, s = y
        self.B = (tm >= tm[:, None]).astype(X.dtype)
        ^

- Resolution failure for non-literal arguments:
None

During: resolving callee type: BoundFunction((<class 'numba.core.types.misc.ClassInstanceType'>, 'initialize') for instance.jitclass.Cox#7f376353e2c0<B:array(float64, 2d, C)>)
During: typing of call at <string> (3)
```

updating numba from 0.56.3 to 0.57 fixed the issue : if we can avoid requiring numba>=0.57 it's better, otherwise we should put it in the requirements

Badr-MOUFAD:
My bad, forgot to require ``numba`` 0.57.

Badr-MOUFAD:
Here is a [link to benchmark results](https://badr-moufad.github.io/share-benchmarks/cox_bench/benchmark_cox_benchopt_run_2023-05-24_16h39m24.html) on a dummy dataset.
Also, a link to the [benchopt repo](https://github.com/Badr-MOUFAD/benchmark_cox) to reproduce

skglm/utils/data.py
```diff
@@ -123,13 +124,13 @@ def make_correlated_data(
...
+def make_dummy_survival_data(n_samples, n_features, normalize=False,
+                             X_density=1., random_state=None):
     """Generate a random dataset for survival analysis.
 
-    The design matrix ``X`` is generated according to standard normal,
-    the vector of time is chosen uniformly from ``[0, 10 * n_samples]``
-    without replacement, and the vector of censorship is drawn from a Bernoulli
-    with parameter 0.5.
+    The design matrix ``X`` is generated according to standard normal, the vector of
+    time ``tm`` is chosen uniformly from ``[0, 10 * n_samples]`` without replacement,
```
PABannier, line 132: Minor: for `tm`, you can sample from a Weibull distribution, this will yield more realistic samples.

.github/workflows/main.yml
```diff
@@ -26,5 +26,6 @@ jobs:
         pip install .
         pip install statsmodels cvxopt
         pip install git+https://github.com/jolars/pyslope.git
+        pip install lifelines
```
mathurinm, line 29: add a comment: for Cox/
Easier for future maintenance to keep track of the need for each dependency 

thinking about it, in another PR these should go in the pyproject.toml in the test section

skglm/datafits/single_task.py
```diff
@@ -544,3 +544,105 @@ def gradient_scalar_sparse(self, X_data, X_indptr, X_indices, y, Xw, j):
...
+
+        out = -s + exp_Xw * (self.B.T @ (s / B_exp_Xw))
+        return out / n_samples
+
+    def raw_hessian(self, y, Xw):
+        """Compute a diagonal upper bound of the datafit's Hessian w.r.t. ``Xw``.
+
+        The diagonal upper bound reads
+
+            exp_Xw * (B.T @ s / B_exp_Xw) / n_samples
```
mathurinm, line 625: maybe write the full hessian and its diagonal upper bound, and indicate a source for that ? If it can't be found in an article, then maybe we should put our derivation somewhere in the doc?

skglm/tests/test_datafits.py
```diff
@@ -114,5 +115,34 @@ def test_gamma():
     np.testing.assert_allclose(clf.coef_, gamma_results.params, rtol=1e-6)
 
 
+def test_cox():
+    rng = np.random.RandomState(1265)
+    n_samples, n_features = 10, 30
```
mathurinm, line 120: This test basically tests the shapes of the returned argument but the maths could be wrong. could we use automatic diff in this test to check that we have the correct gradient and upper bound on Hessian?

skglm/tests/test_estimators.py
```diff
@@ -164,6 +168,89 @@ def test_mtl_path():
...
+    try:
+        from lifelines import CoxPHFitter
+    except ModuleNotFoundError:
+        pytest.xfail(
+            "Testing Cox Estimator requires `lifelines` packages\n"
+            "Run `pip install lifelines`"
+        )
+
+    reg = 1e-2
+    n_samples, n_features = 100, 30
```
mathurinm, line 181: specify here that lifelines fails if n_features > n_samples

skglm/tests/test_estimators.py
```diff
@@ -164,6 +168,89 @@ def test_mtl_path():
...
+    random_state = 1265
```
mathurinm, line 182: no need to declare variable if used only once IMO

skglm/tests/test_estimators.py
```diff
@@ -16,14 +16,18 @@
...
+from skglm.utils.data import make_correlated_data, make_dummy_survival_data
 from skglm.estimators import (
     GeneralizedLinearEstimator, Lasso, MultiTaskLasso, WeightedLasso, ElasticNet,
     MCPRegression, SparseLogisticRegression, LinearSVC)
-from skglm.datafits import Logistic, Quadratic, QuadraticSVC, QuadraticMultiTask
+from skglm.datafits import Logistic, Quadratic, QuadraticSVC, QuadraticMultiTask, Cox
 from skglm.penalties import L1, IndicatorBox, L1_plus_L2, MCPenalty, WeightedL1
 from skglm.solvers import AndersonCD
 
+import pandas as pd
```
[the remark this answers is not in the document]
Badr-MOUFAD: Didn't pay attention to it.

After investigation, it comes with ``lifelines``, also in ``celer --> seaborn --> pandas``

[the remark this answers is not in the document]
mathurinm: agree on the principle, but `python -c "import this" | grep Explicit` so it should go into the requirements

skglm/tests/test_estimators.py
Badr-MOUFAD: Agree!
Though, if you don't have a strong opinion on it, I would rather stick with it as it makes debugging easier.

skglm/tests/test_datafits.py
PABannier: +1 for autodiff. Not sure about `check_grad`, isn't it based on finite difference?

PABannier (approved):
LGTM! 

Open question: do we want to make it an estimator? Once Efron is implemented, I think it's worth implementing, especially if we aim at a wider adoption by the biostat community.

skglm/tests/test_datafits.py
Badr-MOUFAD: @mathurinm, @PABannier, any idea about how to unittest the upper bound on the Hessian?

The explicit expression of the Hessian is huge (considering also Efron)

skglm/datafits/single_task.py
mathurinm: agreed!

skglm/tests/test_datafits.py
```diff
@@ -114,5 +116,41 @@ def test_gamma():
...
+
+    # generate data
+    X = rng.randn(n_samples, n_features)
+    tm = rng.choice(n_samples*n_features, size=n_samples, replace=True).astype(float)
+    s = rng.choice(2, size=n_samples).astype(float)
+    y = (tm, s)
+
+    # generate dummy w, Xw
+    w = rng.randn(n_features)
+    Xw = rng.randn(n_samples)
```
mathurinm, line 131: even if dummy and currently harmless this is error prone if we need to modify the code later : use X @ w

skglm/datafits/single_task.py
Badr-MOUFAD: added issue #161  to tackle that in another PR

PR #158, opened 2023-05-27
MNT - ``ProxNewton`` solver

Badr-MOUFAD:
This performs the following minor changes

- warn when ``ProxNewton`` didn't converge
- ~Increase default ``max_iter``, 20 --> 50~
- refrain from using ``y`` shape

### DIFF
--- a/skglm/solvers/prox_newton.py
+++ b/skglm/solvers/prox_newton.py
@@ -19,7 +19,7 @@ class ProxNewton(BaseSolver):
     p0 : int, default 10
         Minimum number of features to be included in the working set.
 
-    max_iter : int, default 50
+    max_iter : int, default 20
         Maximum number of outer iterations.
 
     max_pn_iter : int, default 1000
@@ -48,7 +48,7 @@ class ProxNewton(BaseSolver):
         code: https://github.com/tbjohns/BlitzL1
     """
 
-    def __init__(self, p0=10, max_iter=50, max_pn_iter=1000, tol=1e-4,
+    def __init__(self, p0=10, max_iter=20, max_pn_iter=1000, tol=1e-4,
                  fit_intercept=True, warm_start=False, verbose=0):
         self.p0 = p0
         self.max_iter = max_iter
--- a/skglm/solvers/prox_newton.py
+++ b/skglm/solvers/prox_newton.py
@@ -1,7 +1,11 @@
+import warnings
+
 import numpy as np
 from numba import njit
 from scipy.sparse import issparse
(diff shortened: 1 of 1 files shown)


skglm/solvers/prox_newton.py
```diff
@@ -242,7 +253,7 @@ def _descent_direction_s(X_data, X_indptr, X_indices, y, w_epoch,
 
     # see _descent_direction() comment
     past_grads = np.zeros(len(ws))
-    X_delta_w_ws = np.zeros(len(y))
+    X_delta_w_ws = np.zeros(Xw_epoch.shape[0])
     ws_intercept = np.append(ws, -1) if fit_intercept else ws
```
Badr-MOUFAD, line 257: A small hack that we used to implement Cox estimator was to pass in ``y`` as tuple ``(tm, s)``.
Doing so would break this function.

skglm/solvers/prox_newton.py
```diff
@@ -242,7 +253,7 @@ def _descent_direction_s(X_data, X_indptr, X_indices, y, w_epoch,
 
     # see _descent_direction() comment
     past_grads = np.zeros(len(ws))
-    X_delta_w_ws = np.zeros(len(y))
+    X_delta_w_ws = np.zeros(Xw_epoch.shape[0])
```
Badr-MOUFAD, line 256: A small hack that we used to implement Cox estimator was to pass in ``y`` as tuple ``(tm, s)``.
Doing so would break this function.

skglm/solvers/prox_newton.py
```diff
@@ -15,7 +19,7 @@ class ProxNewton(BaseSolver):
     p0 : int, default 10
         Minimum number of features to be included in the working set.
 
-    max_iter : int, default 20
+    max_iter : int, default 50
```
mathurinm, line 22: Choosing default values is hard but I don't think 20 is a bad value, it prevents the solver from running too long and people are usually interested in a moderately accurate solution.
Are there reasons to change this ?

skglm/solvers/prox_newton.py
mathurinm: good catch

mathurinm:
+1 for the shape and the warning, not so sure for the change of default default for max_iter.

skglm/solvers/prox_newton.py
Badr-MOUFAD: I don't have a strong opinion on this change.
Adding a warning is indeed sufficient.

PABannier (approved):

PR #159, opened 2023-05-30
ENH - Implement Cox with Efron estimate 

Badr-MOUFAD:
A follow-up of #157,

Handling tied data using Efron estimate can be obtained by slightly modifying the Cox datafit as follows

$$
l(\beta) =  -\langle s, \mathbf{X}\beta \rangle + \langle s, \log(\mathbf{B}e^{\mathbf{X}\beta} - \mathbf{A}e^{\mathbf{X}\beta}) \rangle
$$

where $\mathbf{A}$ is a matrix chosen accordingly to account for the additional term in the Breslow $\log$.
Also, evaluating $\mathbf{A} v$ or  $\mathbf{A}^\top v$ is cheap and can be obtained in linear time.

---------
[Link to the maths behind](https://github.com/scikit-learn-contrib/skglm/files/11600001/main.pdf)


Badr-MOUFAD:
> Maybe you can add a small section at the end of the doc example with lifelines to show that we handle ties the same way lifelines does? no need for another speed benchmark, just showing that we have this functionality too

Yes, I agree with the idea!

examples/plot_survival_analysis.py
```diff
@@ -195,11 +195,74 @@
...
+
+# check data has tied observations
+print("Number of unique times", np.unique(tm))
+print("Number of samples", n_samples)
+
+# %%
+# It is straightforward to fit an :math:`\ell_1` Cox estimator with the Efron estimate.
+# We only need to pass in ``use_efron=True`` to the ``Cox`` datafit.
+
+# ensure using Efron estimate
```
mathurinm, line 228: this is potentially dangerous in other cases (changing the parameter sometimes does not change it in an estimator). For pedagogical purposes it may be better to just reinstantiate a Cox datafit, what do you think ?

examples/plot_survival_analysis.py
```diff
@@ -195,11 +195,74 @@
...
+# Let's start by generating data with tied observation. This can be achieved
+# by passing in a ``with_ties=True`` to ``make_dummy_survival_data`` function.
+tm, s, X = make_dummy_survival_data(
+    n_samples, n_features,
+    normalize=True,
+    with_ties=True,
+    random_state=0
+)
+
+# check data has tied observations
```
mathurinm, line 220:
```suggestion
# check that data has tied observations
```

examples/plot_survival_analysis.py
```diff
@@ -195,11 +195,74 @@
...
+print("Number of unique times", np.unique(tm))
```
mathurinm, line 221: If it can fit on a single line with "number of unique times : xx out of yy samples" it's nicer

Badr-MOUFAD: Looks much cleaner, thanks!

examples/plot_survival_analysis.py
Badr-MOUFAD: It would be problematic if we change params and forget to call ``initialize``

Agree, we better avoid that. Also considering that we use ``lru_cache``, re-defining the datafit won't incur an overhead.

Open question: maybe we should think about a safe mechanism to prevent changing attributes in datafits?

mathurinm: yes, the possibility to call solvers on datafits that have been initialized on other data is potentially harmful... and the need for "initialize" is not obvious to the average user.

Random thoughts:

I'm not sure the problem is solvable, for example in sklearn you can use a Lasso with warm_start=True, and if you fit a second time on an X with a different shape it'll break : 
image



In our situation I'm foreseeing cases with silent, hard to catch convergence issues because someone fits on another X with compatible shape but without recalling `initialize`.
If people use the `GeneralizedLinearEstimator` this won't happen because we call initialize inside.

we could gather all this in an issue for future discussion

Badr-MOUFAD: Open an issue https://github.com/scikit-learn-contrib/skglm/issues/164 to address that

mathurinm:
The edited example and benchmark in this PR may interest @ogrisel too :)

mathurinm (approved):

ogrisel:
Haha the benchmark plot! 😅

ogrisel:
BTW do you see a similar perf improvement for smooth penalties?

Badr-MOUFAD:
@ogrisel we have added support for L2 regularization too (with scipy LBFGS and our Prox Newton methods).
Check the bench figure

L2-Cox-bench-figure

Also, here is the [link to the complete benchmark](https://badr-moufad.github.io/share-benchmarks/cox_bench/benchmark_cox_benchopt_run_2023-06-16_16h29m31.html) and the [repo](https://github.com/benchopt/benchmark_cox) to reproduce it.

PR #170, opened 2023-06-19
FIX - Support of ``float32`` in ProxNewton solver 

Badr-MOUFAD:
This would allow us to solve problems involving ``float32`` data.

@PABannier


skglm/solvers/prox_newton.py
```diff
@@ -329,7 +333,8 @@ def _backtrack_line_search(X, y, w, Xw, fit_intercept, datafit, penalty, delta_w
         grad_ws = _construct_grad(X, y, w[:n_features], Xw, datafit, ws)
         # TODO: could be improved by passing in w[ws]
         stop_crit = penalty.value(w[:n_features]) - old_penalty_val
-        stop_crit += step * grad_ws @ delta_w_ws[:len(ws)]
+        dot = grad_ws @ delta_w_ws[:len(ws)]
```
Badr-MOUFAD, line 336: The two implementations are equivalent, yet Numba complains about the types of the two arrays, though they have the same `dtype`.

I submitted an issue here: https://github.com/numba/numba/issues/9025

mathurinm: nice pinning of the problem. can you add the above comment to the code so that the info and link get into the codebase ?

mathurinm (approved):

PR #171, opened 2023-06-20
ENH - Implement a Cox scikit-learn-like Estimator

Badr-MOUFAD:
follow up of #157, #165, #167 

This wraps the problem of solving a Cox problem in a scikit-learn-like estimator.

PR #173, opened 2023-06-20
ENH - add support for sparse dataset in ``LBFGS``

Badr-MOUFAD:
This enables ``LBFGS`` to solve problems involving sparse ``X``.

To shorten the scope/diff of #171.

### DIFF
--- a/skglm/datafits/single_task.py
+++ b/skglm/datafits/single_task.py
@@ -186,6 +186,16 @@ class Logistic(BaseDatafit):
     def gradient(self, X, y, Xw):
         return X.T @ self.raw_grad(y, Xw)
 
+    def gradient_sparse(self, X_data, X_indptr, X_indices, y, Xw):
+        n_features = X_indptr.shape[0] - 1
+        out = np.zeros(n_features, dtype=X_data.dtype)
+        raw_grad = self.raw_grad(y, Xw)
+
+        for j in range(n_features):
+            out[j] = _sparse_xj_dot(X_data, X_indptr, X_indices, j, raw_grad)
+
+        return out
+
     def full_grad_sparse(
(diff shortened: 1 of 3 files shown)


skglm/datafits/single_task.py
```diff
@@ -5,6 +5,7 @@
 
 from skglm.datafits.base import BaseDatafit
 from skglm.utils.sparse_ops import spectral_norm
+from skglm.solvers.prox_newton import _sparse_xj_dot
```
Badr-MOUFAD, line 8: Thinking back, we shall put this (and also other ones) in [``utils/sparse_ops``](https://github.com/scikit-learn-contrib/skglm/blob/main/skglm/utils/sparse_ops.py)

WDYT @mathurinm ?

mathurinm: definitely makes sense if they are used by other solvers indeed

mathurinm: and they are private so we can move them at will

Badr-MOUFAD: Issue #174  to pin it in the backlog

skglm/solvers/lbfgs.py
```diff
@@ -40,13 +41,20 @@ def objective_function(w):
 
             return datafit_value + penalty_value
 
-        def jacobian_function(w):
+        def d_jacobian_function(w):
```
mathurinm, line 44: small discussion:
- I think `_function` can be removed 
- the jacobian is a gradient here because the output dimension is 1

If we say jacobian to match scipy (which is perfectly fine), I think we can go with `jacobian` or `jac` (+ the s/d prefix)

mathurinm (approved):

mathurinm:
squash and merge when green @Badr-MOUFAD

PR #179, opened 2023-06-27
FIX - compiled and uncompiled ``SLOPE`` give different prox values

mathurinm:


skglm/penalties/non_separable.py
```diff
@@ -40,14 +40,16 @@ def value(self, w):
         return np.sum(np.sort(np.abs(w)) * self.alphas[::-1])
 
     def prox_vec(self, x, stepsize):
-        def _prox(_x, _alphas):
-            sign_x = np.sign(_x)
-            _x = np.abs(_x)
-            sorted_indices = np.argsort(_x)[::-1]
-            prox = prox_SLOPE(_x[sorted_indices], _alphas)
-            prox[sorted_indices] = prox
```
Badr-MOUFAD, line 48: @mathurinm, found the damn bug :bug:! 

it is precisely this line. In particular, refer to example in debug script, ``numpy`` interpret it as

```python
prox[2] = prox[0]
prox[0] = prox[1]
prox[1] = prox[2]
```
and hence overrides `prox[2]`. We endup thereafter with `prox[1] == prox[0]`

Badr-MOUFAD: I believe when compiling, we move to the realm of numba, where the intermediate variable are created automatically hence not having the error.

Badr-MOUFAD:
Thinking back, I would blame ``numpy`` for such a bug.
Indeed, this code swaps the two variables without messing things up

```python
a, b = 1, 2
a, b = b, a

print(a, b)
# OUT: 2, 1
```

Therefore, I would accept the same from `numpy` given how intimate it is with `Python`.

mathurinm:
It's worth escalating this to an issue on the nuova repo since one should expect same outputs between compiled and original code

Reordering of the array is done in place for efficiency, hence the surprising behavior. In your example efficiency is less of an issue since having an int intermediate variable is not costly

mathurinm:
> Thinking back, I would blame `numpy` for such a bug. Indeed, this code swaps the two variables without messing things up
> 
> ```python
> a, b = 1, 2
> a, b = b, a
> 
> print(a, b)
> # OUT: 2, 1
> ```
> 
> Therefore, I would accept the same from `numpy` given how intimate it is with `Python`.

I wouldn't rely too much on such comparisons. Numpy departs from "typical" Python behavior in several ways, for example 
```python
a = b
b += 1 
print(a)
```
will differ if `b` is a float or an array of floats



an easy fix in our case is `prox[indices] = prox.copy()`. LGTY?

Badr-MOUFAD:
> an easy fix in our case is prox[indices] = prox.copy(). LGTY?

btw, I forgot to mention that my last commit fixes the problem.
Do you agree with the suggested solution?

skglm/penalties/non_separable.py
```diff
@@ -40,14 +40,16 @@ def value(self, w):
...
-            return prox * sign_x
-        return _prox(x, self.alphas * stepsize)
+        _alphas = self.alphas * stepsize
+        _x = x
+        prox = np.zeros_like(x)
+
+        sign_x = np.sign(_x)
+        _x = np.abs(_x)
+        sorted_indices = np.argsort(_x)[::-1]
+        prox[sorted_indices] = prox_SLOPE(_x[sorted_indices], _alphas)
```
mathurinm, line 50:
```suggestion
        alphas = self.alphas * stepsize
        prox = np.zeros_like(x)

        sign_x = np.sign(x)
        x = np.abs(x)
        sorted_indices = np.argsort(x)[::-1]
        prox[sorted_indices] = prox_SLOPE(x[sorted_indices], alphas)
```

mathurinm: @Badr-MOUFAD LGTY?

Badr-MOUFAD: LGTM!
one small remark, I changed 
```python
_alphas = self.alphas * stepsize
```
to avoid confusion

mathurinm: better indeed, thanks!

Badr-MOUFAD (approved):

PR #180, opened 2023-06-27
ENH - Add ``global_lipschitz`` to ``Cox`` datafit

### LINKED ISSUE #177: Unable to fit estimator with Cox datafit and SLOPE penalty
`SLOPE` is a subclass of `BasePenalty`, for which `is_penalized()` is defined as an abstract method.
However, `SLOPE` does not implement `is_penalized()`.
This is not recognized during class instantiation because `BasePenalty` is not properly implemented as a subclass of `ABC` and therefore does not enforce the `@abstractmethod` functionality.

Badr-MOUFAD:
Following discussion https://github.com/scikit-learn-contrib/skglm/issues/177#issuecomment-1607757655, it happens that the raw Hessian of Cox datafit can be further upper-bounded by

$$
\nabla^2 F(u) \preceq  \frac{\lVert s \rVert_1}{n}  \mathbf{I}_n
$$

which therefore enables to derive the Lipschitz constant

$$
\frac{\lVert s \rVert_1}{n} \lVert \mathbf{X} \rVert^2_2
$$

This adds ``global_lipschitz`` to Cox datafit and thereby allows it to be combined with non-separable penalties, such SLOPE.

-----
fixes #177

also pining @tobias-zehnder, @qklopfenstein-owkin
The implemented [unittest here](https://github.com/Badr-MOUFAD/skglm/blob/cf77084f8a0b9cf2994e102ca37a10f0041b4bdf/skglm/tests/test_estimators.py#L330-L357) provides a handy starter example for fitting Cox + SLOPE

### DIFF
--- a/skglm/datafits/single_task.py
+++ b/skglm/datafits/single_task.py
@@ -695,13 +695,13 @@ class Cox(BaseDatafit):
         self.initialize(X_data, y)
 
     def init_global_lipschitz(self, X, y):
-        tm, s = y[:, 0], y[:, 1]  # noqa
+        s = y[:, 1]
 
         n_samples = X.shape[0]
         self.global_lipschitz = s.sum() * norm(X, ord=2) ** 2 / n_samples
 
     def init_global_lipschitz_sparse(self, X_data, X_indptr, X_indices, y):
-        tm, s = y[:, 0], y[:, 1]  # noqa
+        s = y[:, 1]
 
         n_samples = s.shape[0]
         norm_X = spectral_norm(X_data, X_indptr, X_indices, n_samples)
--- a/skglm/datafits/single_task.py
+++ b/skglm/datafits/single_task.py
@@ -599,6 +599,7 @@ class Cox(BaseDatafit):
             ('use_efron', bool_),
             ('T_indptr', int64[:]), ('T_indices', int64[:]),
             ('H_indptr', int64[:]), ('H_indices', int64[:]),
+            ('global_lipschitz', float64),
         )
 
     def params_to_dict(self):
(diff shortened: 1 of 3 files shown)


skglm/datafits/single_task.py
```diff
@@ -693,6 +694,20 @@ def initialize_sparse(self, X_data, X_indptr, X_indices, y):
         # small hack to avoid repetitive code: pass in X_data as only its dtype is used
         self.initialize(X_data, y)
 
+    def init_global_lipschitz(self, X, y):
+        tm, s = y[:, 0], y[:, 1]  # noqa
```
mathurinm, line 698: I think it's cleaner to not define tm and remove the noqa, just do s = y[:, 1]

skglm/datafits/single_task.py
```diff
@@ -693,6 +694,20 @@ def initialize_sparse(self, X_data, X_indptr, X_indices, y):
...
+
+        n_samples = X.shape[0]
+        self.global_lipschitz = s.sum() * norm(X, ord=2) ** 2 / n_samples
+
+    def init_global_lipschitz_sparse(self, X_data, X_indptr, X_indices, y):
+        tm, s = y[:, 0], y[:, 1]  # noqa
```
mathurinm, line 704: Same

skglm/tests/test_estimators.py
```diff
@@ -326,6 +326,37 @@ def test_Cox_sk_compatibility():
     check_estimator(CoxEstimator())
 
 
+@pytest.mark.parametrize("use_efron", [True, False])
+def test_cox_SLOPE(use_efron):
+    reg = 1e-2
```
mathurinm, line 331: Nice, can.you just add a comment that this only checks SLOPE in the case where all alphas are equal because there is no alternative slope cox solver

Btw can you test if scipy.optimize.minimize gives the same result up to small errors, for a sequence alphas which is not constant ? That'd be a better test IMO

Badr-MOUFAD: yes sure, 
any hint at the gradient of SLOPE in this case?

mathurinm: don't pass the gradient of slope, just let scipy estimate it with finite differences. it should work up to small numerical errors

qklopfenstein-owkin:
thank you so much @Badr-MOUFAD!

mathurinm (approved):

PR #184, opened 2023-07-13
Add weights and positive options to MCP

QB3:
The goal of this PR is to add weights and positive options to the MCP penalty.

### DIFF
--- a/skglm/__init__.py
+++ b/skglm/__init__.py
@@ -2,5 +2,5 @@ __version__ = '0.3.1dev'
 
 from skglm.estimators import (  # noqa F401
     Lasso, WeightedLasso, ElasticNet, MCPRegression, MultiTaskLasso, LinearSVC,
-    SparseLogisticRegression, GeneralizedLinearEstimator, CoxEstimator
+    SparseLogisticRegression, GeneralizedLinearEstimator, CoxEstimator, WeightedMCPRegression
 )
--- a/skglm/estimators.py
+++ b/skglm/estimators.py
@@ -810,6 +810,9 @@ class MCPRegression(LinearModel, RegressorMixin):
     fit_intercept : bool, optional (default=True)
         Whether or not to fit an intercept.
 
+    positive : bool, optional
+        When set to ``True``, forces the coefficient vector to be positive.
+
     warm_start : bool, optional (default=False)
         When set to ``True``, reuse the solution of the previous call to fit as
         initialization, otherwise, just erase the previous solution.
@@ -837,19 +840,20 @@ class MCPRegression(LinearModel, RegressorMixin):
     """
 
     def __init__(self, alpha=1., gamma=3, max_iter=50, max_epochs=50_000, p0=10,
-                 verbose=0, tol=1e-4, fit_intercept=True, warm_start=False,
-                 ws_strategy="subdiff"):
+                 verbose=0, tol=1e-4, positive=False, fit_intercept=True,
+                 warm_start=False, ws_strategy="subdiff"):
         super().__init__()
         self.alpha = alpha
         self.gamma = gamma
-        self.tol = tol
         self.max_iter = max_iter
         self.max_epochs = max_epochs
         self.p0 = p0
-        self.ws_strategy = ws_strategy
+        self.verbose = verbose
+        self.tol = tol
+        self.positive = positive
         self.fit_intercept = fit_intercept
         self.warm_start = warm_start
-        self.verbose = verbose
+        self.ws_strategy = ws_strategy
 
     def path(self, X, y, alphas, coef_init=None, return_n_iter=True, **params):
         """Compute MCPRegression path.
@@ -890,7 +894,8 @@ class MCPRegression(LinearModel, RegressorMixin):
             The number of iterations along the path. If return_n_iter is set to
             ``True``.
         """
-        penalty = compiled_clone(MCPenalty(self.alpha, self.gamma))
+        penalty = compiled_clone(
+            MCPenalty(self.alpha, self.gamma, self.positive))
         datafit = compiled_clone(Quadratic(), to_float32=X.dtype == np.float32)
         solver = AndersonCD(
             self.max_iter, self.max_epochs, self.p0, tol=self.tol,
@@ -918,7 +923,7 @@ class MCPRegression(LinearModel, RegressorMixin):
             self.max_iter, self.max_epochs, self.p0, tol=self.tol,
             ws_strategy=self.ws_strategy, fit_intercept=self.fit_intercept,
             warm_start=self.warm_start, verbose=self.verbose)
-        return _glm_fit(X, y, self, Quadratic(), MCPenalty(self.alpha, self.gamma),
+        return _glm_fit(X, y, self, Quadratic(), MCPenalty(self.alpha, self.gamma, self.positive),
                         solver)
 
 
--- a/skglm/estimators.py
+++ b/skglm/estimators.py
@@ -22,7 +22,7 @@ from skglm.utils.jit_compilation import compiled_clone
 from skglm.solvers import AndersonCD, MultiTaskBCD
 from skglm.datafits import Cox, Quadratic, Logistic, QuadraticSVC, QuadraticMultiTask
 from skglm.penalties import (L1, WeightedL1, L1_plus_L2, L2,
-                             MCPenalty, IndicatorBox, L2_1)
+                             MCPenalty, WeightedMCPenalty, IndicatorBox, L2_1)
 
 
 def _glm_fit(X, y, model, datafit, penalty, solver):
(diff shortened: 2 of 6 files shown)


skglm/__init__.py
```diff
@@ -2,5 +2,6 @@
 
 from skglm.estimators import (  # noqa F401
     Lasso, WeightedLasso, ElasticNet, MCPRegression, MultiTaskLasso, LinearSVC,
-    SparseLogisticRegression, GeneralizedLinearEstimator, CoxEstimator
+    SparseLogisticRegression, GeneralizedLinearEstimator, CoxEstimator,
+    WeightedMCPRegression
```
mathurinm, line 6: As far as I'm concerned I'd just expose `weights` in MCPRegression in order not to bloat the API

For WeightedLasso we went the other way to gain exposure, but here it's a more niche model so I reckon we can spare the estimator. WDYT?

skglm/estimators.py
```diff
@@ -22,7 +22,7 @@
 from skglm.solvers import AndersonCD, MultiTaskBCD
 from skglm.datafits import Cox, Quadratic, Logistic, QuadraticSVC, QuadraticMultiTask
 from skglm.penalties import (L1, WeightedL1, L1_plus_L2, L2,
-                             MCPenalty, IndicatorBox, L2_1)
+                             MCPenalty, WeightedMCPenalty, IndicatorBox, L2_1)
```
mathurinm, line 25: @Badr-MOUFAD @PABannier : do you think we should implement a new penalty, or simply support weights in the existing `MCPenalty` ?

skglm/estimators.py
```diff
@@ -837,19 +840,189 @@ class MCPRegression(LinearModel, RegressorMixin):
...
         self.p0 = p0
-        self.ws_strategy = ws_strategy
+        self.verbose = verbose
+        self.tol = tol
+        self.positive = positive
         self.fit_intercept = fit_intercept
         self.warm_start = warm_start
+        self.ws_strategy = ws_strategy
+
+    def path(self, X, y, alphas, coef_init=None, return_n_iter=True, **params):
```
mathurinm, line 858: do you need it enough that the code addition is justified ? i.e. calling fit repeatedly with warm start is not an option  ?

skglm/penalties/separable.py
```diff
@@ -222,42 +223,59 @@ class MCPenalty(BasePenalty):
...
-                # distance of -grad to alpha * [-1, 1]
-                subdiff_dist[idx] = max(0, np.abs(grad[idx]) - self.alpha)
-            elif np.abs(w[j]) < self.alpha * self.gamma:
-                # distance of -grad_j to (alpha * sign(w[j]) - w[j] / gamma)
-                subdiff_dist[idx] = np.abs(
-                    grad[idx] + self.alpha * np.sign(w[j]) - w[j] / self.gamma)
+            if self.positive:
+                if w[j] < 0:
+                    subdiff_dist[idx] = np.inf
+                elif w[j] == 0:
```
mathurinm, line 258: duplicated code, no ?
positive just needs to handle w[j] < 0 separately, in that case set subdiff dist to infty and break
```
if positive and w[j] < 0:
    subdiff_dist[idx] = np.inf
    break
and then you keep the old code below

Comment (QB3) [reply skglm/penalties/separable.py:258]: Do we agree that the case w[j]  = 0 also needs to be handled separately? As it is currently done for the l1 norm?

Positive l1 https://github.com/scikit-learn-contrib/skglm/blob/562f42b94a1bae58e2c16ad56a0c28f3a9317548/skglm/penalties/separable.py#L44
l1 https://github.com/scikit-learn-contrib/skglm/blob/562f42b94a1bae58e2c16ad56a0c28f3a9317548/skglm/penalties/separable.py#L51

File: skglm/estimators.py (hunk 2 above)
Comment (Badr-MOUFAD) [reply skglm/estimators.py:25]: I don't have a strong opinion but I would stick with the current design and have a new `WeightedMCPenalty`

Comment (QB3) [reply skglm/estimators.py:25]: I did a try with adding `weights` in the `MCPenalty`, and it starts to be ugly and a pain to add 'weights=np.ones(n_features)' everywhere.
Having only one `MCPRegression` with a `weights` option, and having two penalties `MCPenalty` and `WeightedMCPenalty` seems to be the cleanest option

File: skglm/__init__.py (hunk 1 above)
Comment (QB3) [reply skglm/__init__.py]: Done!

File: skglm/penalties/separable.py (hunk 4 above)
Comment (QB3) [reply skglm/penalties/separable.py]: I merged the cases `w[j] > 0` for `MCPenalty` and `WeightedMCPenalty`

File: skglm/estimators.py (hunk 3 above)
Comment (QB3) [reply skglm/estimators.py]: Since `WeightedMCPRegression` and `MCPRegression` were merged, I guess this comment is fixed

File: skglm/estimators.py (hunk 2 above)
Comment (PABannier) [reply skglm/estimators.py:25]: Agreed with @QB3 !

File: skglm/estimators.py [hunk 5]
```diff
@@ -890,7 +899,14 @@ def path(self, X, y, alphas, coef_init=None, return_n_iter=True, **params):
             The number of iterations along the path. If return_n_iter is set to
             ``True``.
         """
-        penalty = compiled_clone(MCPenalty(self.alpha, self.gamma))
+        weights = np.ones(X.shape[1]) if self.weights is None else self.weights
+        if X.shape[1] != len(weights):
+            raise ValueError("The number of weights must match the number of \
+                              features. Got %s, expected %s." % (
+                len(weights), X.shape[1]))
```
Badr-MOUFAD, line 906: we better use f-string
```suggestion
            raise ValueError(
                "The number of weights must match the number of features. "
                f"Got {len(weights)}, expected {X.shape[1]}."
            )
```

skglm/estimators.py
```diff
@@ -890,7 +899,14 @@ def path(self, X, y, alphas, coef_init=None, return_n_iter=True, **params):
...
+        penalty = compiled_clone(
+            WeightedMCPenalty(
+                self.alpha, self.gamma, weights, self.positive))
```
Badr-MOUFAD, line 909:
```suggestion
            WeightedMCPenalty(self.alpha, self.gamma, weights, self.positive)
        )
```

skglm/estimators.py
```diff
@@ -914,12 +930,16 @@ def fit(self, X, y):
         self :
             Fitted estimator.
         """
+        if self.weights is None:
```
Badr-MOUFAD, line 933: Maybe I'm missing some context, but is there a reason for not checking the shape of weights to match the number of features?

skglm/penalties/separable.py
```diff
@@ -222,42 +223,51 @@ class MCPenalty(BasePenalty):
...
 
     def value(self, w):
         return value_MCP(w, self.alpha, self.gamma)
 
     def prox_1d(self, value, stepsize, j):
         """Compute the proximal operator of MCP."""
-        return prox_MCP(value, stepsize, self.alpha, self.gamma)
+        return prox_MCP(value, stepsize, self.alpha, self.gamma, self.positive)
 
     def subdiff_distance(self, w, grad, ws):
```
Badr-MOUFAD, line 251: :ok_hand:

skglm/penalties/separable.py
```diff
@@ -273,6 +283,89 @@ def alpha_max(self, gradient0):
...
+
+    Notes
+    -----
+    With :math:`x >= 0`:
+
+    .. math::
+        "pen"(x) = {(alpha x - x^2 / (2 gamma), if x =< alpha gamma),
+                    (gamma alpha^2 / 2        , if x > alpha gamma):}
+    .. math::
+        "value" = sum_(j=1)^(n_"features") "pen"(abs(w_j))
```
Badr-MOUFAD, line 297: Weights are missing in the formula, right?
```suggestion
        "value" = sum_(j=1)^(n_"features") "weights"_j "pen"(abs(w_j))
```

skglm/penalties/separable.py
```diff
@@ -273,6 +283,89 @@ def alpha_max(self, gradient0):
...
+
+    def value(self, w):
+        return value_weighted_MCP(w, self.alpha, self.gamma, self.weights)
+
+    def prox_1d(self, value, stepsize, j):
+        """Compute the proximal operator of the weighted MCP."""
+        return prox_MCP(
+            value, stepsize, self.alpha, self.gamma, self.positive, self.weights[j])
+
+    def subdiff_distance(self, w, grad, ws):
```
Badr-MOUFAD, line 329: :ok_hand:

skglm/utils/prox_funcs.py
```diff
@@ -60,15 +60,26 @@ def value_MCP(w, alpha, gamma):
...
-def prox_MCP(value, stepsize, alpha, gamma):
-    """Compute the proximal operator of stepsize * MCP penalty."""
+def value_weighted_MCP(w, alpha, gamma, weights):
+    """Compute the value of the weighted MCP."""
+    s0 = np.abs(w) < gamma * alpha
+    value = np.zeros_like(w)
+    value[~s0] = weights[~s0] * gamma * alpha ** 2 / 2.
+    value[s0] = alpha * np.abs(w[s0]) - w[s0]**2 / (2 * gamma)
+    value[s0] *= weights[s0]
+    return np.sum(value)
```
Badr-MOUFAD, line 70: Since the weighted MCP has the following formula

$$
\sum_j w_j \ \mathrm{MCP}(\beta_j)
$$

Its implementation should resemble `value_MCP` with the exception of having a weighted sum in the end.

```suggestion
def value_weighted_MCP(w, alpha, gamma, weights):
    """Compute the value of the weighted MCP."""
    s0 = np.abs(w) < gamma * alpha
    value = np.full_like(w, gamma * alpha ** 2 / 2.)
    value[s0] = alpha * np.abs(w[s0]) - w[s0]**2 / (2 * gamma)
    return np.sum(weights * value)
```

skglm/utils/prox_funcs.py
```diff
@@ -60,15 +60,26 @@ def value_MCP(w, alpha, gamma):
...
+
+
+@njit
+def prox_MCP(value, stepsize, alpha, gamma, positive=False, weight=1.):
```
Badr-MOUFAD, line 74: I'm not sure the maths are correct
With positive MCP we add $\delta_{x \geq0 }$, hence we should end up with [^1]<sup>(Eq 2 and below)</sup>

$$
\mathrm{prox} = \Pi_{\mathbb{R}} \circ \mathrm{prox}_{\mathrm{MCP}}
$$

Which I don't think reflects the last condition.
Any thoughts @mathurinm?

[^1]: Pustelnik, N. and Condat, L., 2017. Proximity operator of a sum of functions; application to depth map estimation. IEEE Signal Processing Letters, 24(12), pp.1827-1831.

skglm/estimators.py
```diff
@@ -890,7 +899,14 @@ def path(self, X, y, alphas, coef_init=None, return_n_iter=True, **params):
             The number of iterations along the path. If return_n_iter is set to
             ``True``.
         """
-        penalty = compiled_clone(MCPenalty(self.alpha, self.gamma))
+        weights = np.ones(X.shape[1]) if self.weights is None else self.weights
```
Badr-MOUFAD, line 902: This way we would use `WeightedMCP` even though the `self.weights` isn't provided.
We better use the same logic as you implemented in `.fit` method.

Badr-MOUFAD (commented):
Thanks, @QB3 for the PR :rocket:!
Overall it looks good :ok_hand:

skglm/utils/prox_funcs.py
```diff
@@ -60,15 +60,26 @@ def value_MCP(w, alpha, gamma):
...
+    """Compute the proximal operator of stepsize * weight MCP penalty."""
     tau = alpha * stepsize
```
mathurinm, line 76: prox of weight * MCP(gamma, alpha) is not prox of MCP(gamma, alpha * weight) so I don't get why only tau is multiplied by weight. Shouldn't gamma be divided too?

skglm/utils/prox_funcs.py
mathurinm: Why is not reflected in the last condition ? proxMCP has the same sign as its input

Badr-MOUFAD: That's right, thanks for the clarification!

skglm/utils/prox_funcs.py
Badr-MOUFAD: The prox formula is correct as when ``tau`` and ``gamma`` are multiplied the ``weight`` cancel out.

However, I do agree with you, we better write
```python
tau = alpha * stepsize * weight
gamma = gamma / stepsize / weight
```
and remove the hardcoded `weight` in the remaining function body.

QB3: +1 in order to handle 0 weights and avoid numerical issues, I do not call MCP(gamma / weight, alpha * weight) . Instead, I replaced "by hand" `alpha` by `alpha * weight` and `gamma` by `gamma / weight`.
This way, the weight cancels in `g * tau` here https://github.com/QB3/skglm/blob/461065eb7a1b96874629ade362beb6350be09b7a/skglm/utils/prox_funcs.py#L80

skglm/estimators.py
QB3: done thx!

Badr-MOUFAD (commented):
LGTM!
Thanks @QB3 for the hard work!

Before merging, could you add this new feature to the what's new section of the documentation?

Badr-MOUFAD (approved):
Thanks @QB3 again for this nice job :ok_hand:

doc/changes/0.4.rst
```diff
@@ -0,0 +1,5 @@
+.. _changes_0_4:
+
+Version 0.4 (in progress)
```
Badr-MOUFAD, line 3: v0.4.0 instead of v0.3.1 as we are [adding a new feature](https://semver.org/)

Issue #191, opened 2023-10-24
API: refactor what `initialize` does


mathurinm:
Currently, it's impossible to check if a datafit computes Lipschitz constants without calling its `initialize` method, which is costly.

As discussed IRL with @Badr-MOUFAD, we can solve this by:
- doing only what needs to be done regardless of the solver in `datafit.initialize()` (eg B computation in Cox)
- having dedicated methods such as `Quadratic.compute_lipschitz`, which compute it (and store it ? Probably not needed), as we currently do for `Quadratic.init_global_lipschitz`. 

Each solver can then check easily that a datafit/penalty implements all required methods and then calls them to compute whatever quantities it requires to perform optimization.

PR #192, opened 2023-10-25
API - Expose a datafit method for solvers to get Lipschitz

Badr-MOUFAD:
## Context of the PR

As highlighted in #191,
This makes Lipschitz (resp. global Lipschitz) accessible via a method instead of an attribute.
``AndersonCD``, ``FISTA`` will require the datafit to define respectively ``get_lipschitz_*`` and ``get_global_lipschitz_*``

Now on ``initialize_*`` is a method to initialize internal attributes of the datafit (attributes not used by solver) such as ``Xty`` of the ``Quadratic`` datafit.

## Contributions of the PR

- expose a datafit method called by solvers to get Lipschitz constants

### Checks before merging PR

- ~[ ] added unittests~
- [x] edited the [what's new](../doc/changes/whats_new.rst)

---
The ``*``  is for ``sparse``

### DIFF
--- a/doc/changes/0.4.rst
+++ b/doc/changes/0.4.rst
@@ -3,4 +3,4 @@
 Version 0.4 (in progress)
 ---------------------------
 - Add support for weights and positive coefficients to :ref:`MCPRegression Estimator <skglm.MCPRegression>` (PR: :gh:`184`)
-- Expose a datafit method called by solvers to get Lipschitz. Now on, ``initialize_*`` is meant to initialize internal attributes of the datafit. (PR: :gh:`192`)
+- Move solver specific computations from ``Datafit.initialize()`` to separate ``Datafit`` methods to ease ``Solver`` - ``Datafit`` compatibility check (PR: :gh:`192`)
--- a/doc/changes/0.4.rst
+++ b/doc/changes/0.4.rst
@@ -3,3 +3,4 @@
 Version 0.4 (in progress)
 ---------------------------
 - Add support for weights and positive coefficients to :ref:`MCPRegression Estimator <skglm.MCPRegression>` (PR: :gh:`184`)
+- Expose a datafit method called by solvers to get Lipschitz. Now on, ``initialize_*`` is meant to initialize internal attributes of the datafit. (PR: :gh:`192`)
--- a/skglm/datafits/multi_task.py
+++ b/skglm/datafits/multi_task.py
@@ -28,36 +28,52 @@ class QuadraticMultiTask(BaseMultitaskDatafit):
     def get_spec(self):
         spec = (
             ('XtY', float64[:, :]),
-            ('lipschitz', float64[:]),
         )
         return spec
 
     def params_to_dict(self):
         return dict()
 
-    def initialize(self, X, Y):
-        """Compute optimization quantities before fitting on X and Y."""
-        self.XtY = X.T @ Y
+    def get_lipschitz(self, X, Y):
         n_samples, n_features = X.shape
-        self.lipschitz = np.zeros(n_features)
-        for j in range(n_features):
-            self.lipschitz[j] = norm(X[:, j]) ** 2 / n_samples
 
-    def initialize_sparse(self, X_data, X_indptr, X_indices, Y):
+        lipschitz = np.zeros(n_features)
+        for j in range(n_features):
+            lipschitz[j] = norm(X[:, j]) ** 2 / n_samples
+
+        return lipschitz
+
+    def get_lipschitz_sparse(self, X_data, X_indptr, X_indices, Y):
         """Pre-computations before fitting on X and Y, when X is sparse."""
         n_samples, n_tasks = Y.shape
         n_features = len(X_indptr) - 1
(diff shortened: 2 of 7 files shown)


skglm/datafits/multi_task.py
```diff
@@ -28,36 +24,52 @@ def __init__(self):
...
+        n_samples, n_features = X.shape
+
+        lipschitz = np.zeros(n_features)
+        for j in range(n_features):
+            lipschitz[j] = norm(X[:, j]) ** 2 / n_samples
+
+        return lipschitz
+
+    def get_lipschitz_sparse(self, X_data, X_indptr, X_indices, Y):
+        """Pre-computations before fitting on X and Y, when X is sparse."""
```
mathurinm, line 43: rm docstring

skglm/solvers/anderson_cd.py
```diff
@@ -277,6 +284,9 @@ def _cd_epoch(X, y, w, Xw, datafit, penalty, ws):
     Xw : array, shape (n_samples,)
         Model fit.
 
+    lc : array, shape (n_features,)
+        Lipschitz constants of the features.
```
mathurinm, line 288: "Coordinatewise gradient Lipschitz constants."

skglm/solvers/anderson_cd.py
```diff
@@ -322,6 +331,9 @@ def _cd_epoch_sparse(X_data, X_indptr, X_indices, y, w, Xw, datafit, penalty, ws
     Xw : array, shape (n_samples,)
         Model fit.
 
+    lc :  array, shape (n_features,)
+        Lipschitz constants of the features.
```
mathurinm, line 335: same

skglm/solvers/multitask_bcd.py
```diff
@@ -353,6 +363,10 @@ def _bcd_epoch(X, Y, W, XW, datafit, penalty, ws):
     XW : array, shape (n_samples, n_tasks)
         Model fit.
 
+    lc :  array, shape (n_features,)
+        Lipschitz constants of each bloc.
+
```
mathurinm, line 368: extra blank line

skglm/solvers/multitask_bcd.py
```diff
@@ -146,7 +150,9 @@ def solve(self, X, Y, datafit, penalty, W_init=None, XW_init=None):
                     if self.ws_strategy == "subdiff":
                         opt_ws = penalty.subdiff_distance(W, grad_ws, ws)
                     elif self.ws_strategy == "fixpoint":
-                        opt_ws = dist_fix_point(W, grad_ws, datafit, penalty, ws)
+                        opt_ws = dist_fix_point(
+                            W, grad_ws, lipschitz, datafit, penalty, ws
```
mathurinm, line 154: dist_fix_point is really dist_fixed_point_cd (it's algorithm dependent, otherwise it should not depend on the stepsize).

it felt more elegant when the user did not have to pass `lipschitz`, but then again now it's more explicit because you don't run into an error when calling `dist_fix_point` on a datafit which has not been initialized and does not have a `lipschitz` attribute, so I think the current design is safer.

What are your thoughts on renaming `dist_fix_point` (later) ?

skglm/solvers/multitask_bcd.py
Badr-MOUFAD: Nice catch thanks!

skglm/solvers/multitask_bcd.py
Badr-MOUFAD: I'm more in favor of tackling that in a separate PR

doc/changes/0.4.rst
```diff
@@ -3,3 +3,4 @@
 Version 0.4 (in progress)
 ---------------------------
 - Add support for weights and positive coefficients to :ref:`MCPRegression Estimator <skglm.MCPRegression>` (PR: :gh:`184`)
+- Expose a datafit method called by solvers to get Lipschitz. Now on, ``initialize_*`` is meant to initialize internal attributes of the datafit. (PR: :gh:`192`)
```
mathurinm, line 6: Something like "Move solver specific computations (eg Lipschitz constants) from Datafit.initialize() to separate Datafit methods (eg get_lipschitz_constant), to ease Solver/Datafit compatibility check" makes it less about the aprticular case of Lipschitz constants and more about the general principle

skglm/solvers/multitask_bcd.py
```diff
@@ -403,6 +415,9 @@ def _bcd_epoch_sparse(X_data, X_indptr, X_indices, Y, W, XW, datafit, penalty, w
     XW : array, shape (n_samples, n_tasks)
         Model fit.
 
+    lc :  array, shape (n_features,)
+        Lipschitz constants of each bloc.
```
mathurinm, line 419: same renaming

mathurinm (approved):
up to minor changes

one potential issue with not storing lipschitz constants anymore is that calling the solver multiple times on same X (eg when using a reweighted L1 approach) without recomputing the LC is no longer possible

PR #214, opened 2023-12-14
ENH - add support for intercept in `SqrtLasso`

### LINKED ISSUE #96: ENH - Add support for intercept in ``SqrtLasso``
Doable after merging #68

1. Expose a ``fit_intercept`` argument in ``SqrtLasso``
2. Handle ``fit_intercept`` in the ``path`` method
3. Set ``intercept_`` in the ``fit`` method

PascalCarrivain:
## Context of the PR

This PR adds support for intercept in `SqrtLasso` estimator.
Closes #96 

### Checks before merging PR

- [ ] added unittests
- [ ] edited the [what's new](../doc/changes/whats_new.rst)


skglm/experimental/sqrt_lasso.py
```diff
@@ -193,12 +200,18 @@ def path(self, X, y, alphas=None, eps=1e-3, n_alphas=10):
...
             # no warm start for the first alpha
-            coef_init = coefs[i].copy() if i else np.zeros(n_features)
+            coef_init = coefs[i].copy() if i else np.zeros(n_features
+                                                           + self.fit_intercept)
 
             try:
-                coef, _, _ = self.solver_.solve(
-                    X, y, sqrt_quadratic, l1_penalty,
-                    w_init=coef_init, Xw_init=X @ coef_init)
+                if self.fit_intercept:
```
Badr-MOUFAD, line 207: Here we are inverting the cases, therefore it break the CI

skglm/experimental/sqrt_lasso.py
```diff
@@ -209,7 +222,10 @@ def path(self, X, y, alphas=None, eps=1e-3, n_alphas=10):
                 # save coef despite not converging
                 # coef_init holds a ref to coef
                 coef = coef_init
-                res_norm = norm(y - X @ coef)
+                if self.fit_intercept:
+                    res_norm = norm(y - X @ coef[:-1] - coef[-1])
+                else:
+                    res_norm = norm(y - X @ coef)
```
Badr-MOUFAD, line 228: WDYT about this implementation to make it more precise?

```suggestion
                X_coef = X @ coef if not fit_intercept else X @ coef[:-1] + coef[-1]
                res_norm = norm(y - X_coef)
```

Badr-MOUFAD (commented):
Thanks @PascalCarrivain for the PR!
Here are some remarks

skglm/experimental/sqrt_lasso.py
mathurinm, line 207: There's duplicated here, I'd just put the definition of  Xw_init in the if else, but then coef, _, _ = ... can be factorized outside

skglm/experimental/tests/test_sqrt_lasso.py
```diff
@@ -16,7 +16,10 @@ def test_alpha_max():
 
     sqrt_lasso = SqrtLasso(alpha=alpha_max).fit(X, y)
 
-    np.testing.assert_equal(sqrt_lasso.coef_, 0)
+    if sqrt_lasso.fit_intercept:
+        np.testing.assert_equal(sqrt_lasso.coef_[:-1], 0)
+    else:
+        np.testing.assert_equal(sqrt_lasso.coef_, 0)
```
Badr-MOUFAD, line 22: WDYT about this refactoring?

```suggestion
    np.testing.assert_equal(sqrt_lasso.coef_[:n_features], 0)
```

Badr-MOUFAD (commented):
Many thanks Pascal for the hard work :muscle:

Just a small bit before merging

We didn't add tests for the intercept (might be vulnerable to bugs later)

I'm not aware of any package to check against it the new feature, any suggestions @mathurinm?

With that being said, we can leverage the equivalence between the problem with intercept and without as the datafit is a norm_2 hence setting its gradient w.r.t. the intercept to zero implies setting the residual to zero

$$
\\begin{equation}
\\min_{w, w_0} \\| y - (Xw - w_0 \mathbb{1})  \\| + \\lambda \\| w \\|_1  \\Longleftrightarrow
\\begin{cases}
\\min_w & \\| \\bar{y} - \\bar{X}w \\| + \\lambda \\| w \\|_1 \\\\
\\text{where} & w_0 = \\text{mean}(y) - \\langle \\text{mean\\_col}(X), \\hat{w} \\rangle
\\end{cases}
\\end{equation}
$$

mathurinm:
@PascalCarrivain thanks for bringing this up again ; there's a conflict that needs to be solved

Also @Badr-MOUFAD how could we test against a Lasso solver ? Do you remember some formula linking the regularization strentgth of the Lasso to that of the sqrt Lasso ? 
Should we, one some particular data, find by dichotomy two values of lambda that give the same solution for Lasso and SqrtLasso, and then hardcode them in a test ? It sounds good enough for me, there hasn't been a crushing pressure for this feature...

mathurinm:
Using this works: 
$$\lambda_{sqrt} = \frac{n \lambda_{lasso} }{\Vert y - X\hat \beta_{lasso} \Vert}$$
```
import numpy as np
from numpy.linalg import norm
from skglm import Lasso
from skglm.experimental import SqrtLasso

np.random.seed(0)
X = np.random.randn(10, 20)
y = np.random.randn(10)

n = len(y)

alpha_max = norm(X.T @ y, ord=np.inf) / n

alpha = alpha_max / 10

lass = Lasso(alpha=alpha, fit_intercept=False, tol=1e-8).fit(X, y)
w_lass = lass.coef_
assert norm(w_lass) > 0

scal = n / norm(y - X @ w_lass)

sqrt = SqrtLasso(alpha=alpha * scal, tol=1e-8).fit(X, y)

print(norm(w_lass - sqrt.coef_))  # this is 1e-8
```


this is not the usual equivalence (one would expect a sqrt(n) not a n, so that $\lambda_{sqrt} = \lambda_{lasso} / \hat{sigma}$), but this is because our SqrtLasso objective does not have the 1/sqrt(n) normalization in front of the datafit


@floriankozikowski when an intercept is fitted the scaling factor should just be `scal = n / norm(y - sqrt.predict(X))`

mathurinm:
Ok this works on the PR:
```
import numpy as np
from numpy.linalg import norm
from skglm import Lasso
from skglm.experimental import SqrtLasso

np.random.seed(0)
X = np.random.randn(10, 20)
y = np.random.randn(10)
y += 1

n = len(y)

alpha_max = norm(X.T @ y, ord=np.inf) / n

alpha = alpha_max / 10

lass = Lasso(alpha=alpha, fit_intercept=True, tol=1e-8).fit(X, y)
w_lass = lass.coef_
assert norm(w_lass) > 0

scal = n / norm(y - lass.predict(X))

sqrt = SqrtLasso(alpha=alpha * scal, fit_intercept=True, tol=1e-8).fit(X, y)
```

but we have a problem, sqrt.coef_ is 1 coefficient too long, because we still store the intercept as last value. This is easy to fix @PascalCarrivain

skglm/experimental/sqrt_lasso.py
```diff
@@ -132,7 +136,10 @@ def fit(self, X, y):
             Fitted estimator.
         """
         self.coef_ = self.path(X, y, alphas=[self.alpha])[1][0]
-        self.intercept_ = 0.  # TODO handle fit_intercept
+        if self.fit_intercept:
+            self.intercept_ = self.coef_[-1]
```
mathurinm, line 140: @PascalCarrivain you also need to make self.coef_ shorter by one element in that case, otherwise it still contains the itnercept as last value

mathurinm (changes requested):

.DS_Store
mathurinm, line 1: @floriankozikowski remove these 3 files

mathurinm:
This has become hard to read due to git errors, closing in favor of the restart #298

Issue #197, opened 2023-10-30
FEAT - Add GroupLasso and GroupLogisticRegression estimators 


mathurinm:

tomaszkacprzak:
Hello, following up on the [issue](https://github.com/mathurinm/celer/issues/290) from Celer, it would be great to be able to run GroupLasso with non-neg weights. The standard Lasso already uses `positive=True`. Would that be possible to implement. 

For my application I also need to run on sparse X input matrices. From what I have seen, some of the regressors can handle the `csc` matrices. Would that be possible for GroupLasso?

QB3:
Hello,
Thanks a lot for your interest!
- adding a group Lasso with positive weights should be "easy", we should be able to do it quickly (i.e. this week if we have time)
- handling sparse matrices for the group Lasso is definitely doable, but might take a bit more time to check that the implementation is correct

tomaszkacprzak:
Thank you, I am glad that it is doable. Is there anything I can do to help?

mathurinm:
@tomaszkacprzak we can start by supporting sparse matrices for the GroupBCD solver. Do you think you could send a PR doing that ? 

You'd need to implement `construct_grad_sparse` and `bcd_epoch_sparse` in `solvers/group_bcd.py`. They should do the same thing as their existing dense versions (same name without `_sparse`), but when X is a csc matrix, passed as a triplet X_data, X_indices, X_indptr

See current implementations for CD (not BCD) in `solvers/anderson_cd.py`: sparse version https://github.com/scikit-learn-contrib/skglm/blob/main/skglm/solvers/anderson_cd.py#L313 vs dense version https://github.com/scikit-learn-contrib/skglm/blob/main/skglm/solvers/anderson_cd.py#L272

Then inside the solver we just have an if when doing the epoch of coordinate descent : https://github.com/scikit-learn-contrib/skglm/blob/main/skglm/solvers/anderson_cd.py#L147

Do not hesitate to start a very WIP PR, so that we can discuss over there!

QB3:
In the meantime, I will open a separate PR for the positive weights

QB3:
Do we want a `GroupLasso` estimator [as well](https://github.com/mathurinm/celer/blob/c1b564c3c0c8fc214deb67b57f4a4dd95f66121f/celer/dropin_sklearn.py#L994) ?

mathurinm:
[...]

PR #225, opened 2024-03-04
FIX - computation of ``subdiff_distance`` in ``WeightedGroupL2`` penalty

Badr-MOUFAD:
## Context of the PR

This fixes the bug reported in issue #197 (thanks for the investigation @tomaszkacprzak)

## Contributions of the PR

- fix computation of `subdiff_distance`
- harden unittest

### Checks before merging PR

- ~[ ] added documentation for any new feature~
- [x] added unittests
- ~[ ] edited the [what's new](../doc/changes/whats_new.rst)~

### DIFF
--- a/doc/tutorials/prox_nn_group_lasso.rst
+++ b/doc/tutorials/prox_nn_group_lasso.rst
@@ -135,5 +135,27 @@ and thus, combined with Equations :eq:`prox_projection_nn_Sc` and :eq:`prox_proj
     .
 
 
+
+Subdifferential of the positive Group Lasso penalty
+===================================================
+
+For the `subdiff_diff` working set strategy, we compute the distance to the subdifferential of
+the $\Vert \cdot \Vert + \iota_{\mathbb{R}_+}$ penalty.
+
+If any component is strictly negative, the subdifferential is empty, and the distance is $+ \infty$.
+
+At a non zero point with strictly positive entries, the penalty is differentiable with only subgradient $w_g/\Vert w_g \Vert$.
+
+At 0, the subdifferential is
+.. math::
+
+    \lambda \partial \lVert \cdot \lVert_2 + \partial \delta_{\mathbb{R}_{+}^g} = \lambda \mathcal{B}_2 + \mathbb{R}^{-}_g
+
+Let $v \in \mathbb{R}^g$, and $\hat v$ its projection onto $\lambda \mathcal{B}_2 + \mathbb{R}^{-}_g$.
+It is clear that for $j$ such that $v_j \leq 0$, $v_j = \hat v_j$.
+Then, the entries in $\mathcal{S} = \{j : v_j > 0}$ are simply given by the projection of $v_\mathcal{S}$ onto $\lambda \mathcal{B}_2$.
+
+
 References
 ==========
+
--- a/doc/tutorials/prox_nn_group_lasso.rst
+++ b/doc/tutorials/prox_nn_group_lasso.rst
@@ -16,7 +16,7 @@ Let
     g:x \mapsto \norm{x}_2
     ,
 
-then
+then its Fenchel-Legendre conjugate is
 
 .. math::
     :label: fenchel
@@ -42,7 +42,7 @@ Using the Moreau decomposition, Equations :eq:`fenchel` and :eq:`prox_projection
     \text{prox}_{\lambda g}(x)
     =
     x
-    - \lambda \text{prox}_{\lambda g^\star}(x/\lambda)
+    - \lambda \text{prox}_{g^\star/\lambda }(x/\lambda)
 
 .. math::
 
@@ -112,7 +112,7 @@ As before, using the Moreau decomposition and Equation :eq:`fenchel_nn` yields
     \text{prox}_{\lambda h}(x)
     =
     x
-    - \lambda \text{prox}_{\lambda h^\star}(x/\lambda)
+    - \lambda \text{prox}_{h^\star / \lambda }(x/\lambda)
 
 .. math::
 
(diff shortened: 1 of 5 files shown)


skglm/penalties/block_separable.py
```diff
@@ -324,10 +324,11 @@ def subdiff_distance(self, w, grad_ws, ws):
...
             if self.positive and np.any(w_g < 0):
                 scores[idx] = np.inf
-            if self.positive and norm_w_g == 0:
-                # distance of -norm(grad_g) to )-infty, alpha * weights[g]]
-                scores[idx] = max(0, - norm(grad_g) - self.alpha * weights[g])
-            if (not self.positive) and norm_w_g == 0:
+            elif self.positive and norm_w_g == 0:
+                # distance of -norm(neg_grad_g) to weights[g] * [-alpha, alpha]
+                neg_grad_g = np.where(grad_g < 0., grad_g, 0.)
+                scores[idx] = max(0, norm(neg_grad_g) - self.alpha * weights[g])
```
QB3, line 330: why not using
`scores[idx] = max(0, norm(grad_g[grad_g < 0.]) - alpha * weights[g])` ?
or 
`neg_grad_g = grad_g[grad_g < 0.]) ` ?

QB3 (commented):
Thank you so much for the quick fix @Badr-MOUFAD !
For future maintenance, maybe we could add details on the group Lasso subdifferential computation in the tutorial as well https://contrib.scikit-learn.org/skglm/tutorials/prox_nn_group_lasso.html#prox-nn-group-lasso

skglm/penalties/block_separable.py
Badr-MOUFAD: yes indeed @QB3, looks clear and clean

Badr-MOUFAD: To compute the distance to the subdifferential (for group $g$)

$$
\begin{equation}
\begin{aligned}
v \in \partial \Omega & \Leftrightarrow v \in \lambda \partial \lVert \cdot \lVert_2 + \partial \delta_{\mathbb{R}_{+}^g} \\\\
 &  \Leftrightarrow v \in \lambda \mathcal{B}_2 + \mathbb{R}^{-}_g
\end{aligned}
\end{equation}
$$

hence, the distance to subdiff is the solution of this problem (Minkowski sum)

$$
\begin{equation}
\begin{aligned}
\min_{u, n} & \ \lVert u + n - v \rVert^2 \\
\mathrm{st.} & \ u \in \lambda \\mathcal{B}_2, n \in \\mathbb{R}_g^{-}
\end{aligned}
\end{equation}
$$

The implemented formula solves this problem by solving first for $n \in \\mathbb{R}_g^{-}$, then for $u \in \lambda \\mathcal{B}_2$
Intuitively, it seems right to do that for this case, but I can't prove it rigorously

QB3: Maybe we can extend [this proof](https://math.stackexchange.com/questions/2755045/how-is-the-subdifferential-of-the-l-2-norm-at-x-0-the-polar-of-the-unit-ball).
I would vote for merging the fix quickly, and then doing a PR to detail the subdifferential computation.

mathurinm: @Badr-MOUFAD can you include the derivation in https://contrib.scikit-learn.org/skglm/tutorials/prox_nn_group_lasso.html#prox-nn-group-lasso ? Since this is not "trivial"

You're looking at a projection problem; it's clear that for the negative entries, their value in the solution is the same; then you're left with the projection of the positive entries on the ball + R_-, wich is clearly the projection on the ball too (as projecting onto R- would imply changing the sign, which is suboptimal)

mathurinm: >  I would vote for merging the fix quickly, and then doing a PR to detail the subdifferential computation.

I disagree, we tend to not finish stuff that is "left for later"
The proof is not complex (we can even just give the solution)

mathurinm:
@Badr-MOUFAD @QB3 I'm done with the pass on docs, make any changes you want and merge

mathurinm (approved):

Badr-MOUFAD:
@QB3 a glance at the derivation from you is a plus

doc/tutorials/prox_nn_group_lasso.rst
```diff
@@ -135,5 +134,65 @@ and thus, combined with Equations :eq:`prox_projection_nn_Sc` and :eq:`prox_proj
...
+    \lambda \partial || \cdot ||_2 + \partial \iota_{x \geq 0} = \lambda \mathcal{B}_2 + \mathbb{R}_-^g
+    ,
+
+where :math:`\mathcal{B}_2` is the unit ball.
+
+Therefore, the distance to the subdifferential writes
+
+.. math::
+
+    D(v) = \min_{u \in \lambda \mathcal{B}_2, n \in \mathbb{R}_{-}^g} \ || u + n - v ||
```
mathurinm, line 182: @Badr-MOUFAD this hides the problem, because why is the solution for a negative v, n = 0? 

I'm subjective but I find it way clearer the way it was before, which avoids reference to an external source

Badr-MOUFAD: > @Badr-MOUFAD this hides the problem, because why is the solution for a negative v, n = 0?

if v is negative, one can set n = v and u = 0 to get a distance equal zero (zero is a lower bound on the problem)

> I'm subjective but I find it way clearer the way it was before, which avoids reference to an external source

The external source is a stack overflow thread to justify the computation of the min regardless of the order.

I believe the benefit of having formulas is to easily and efficiently look up the derivation later.
thought, I'm +1 for adding an explanation, that way we can have both worlds.

QB3: > @QB3 a glance at the derivation from you is a plus

I also think Mathurin's version was clearer.  I would also try to be more self-contained (e.g. we currently talk about the distance, without always clearly stating that this is the distance of the gradient to the subdiff), but writing something clear and self-contained might take more time.

TLDR: I am happy to merge Mathurin's version and iterate later on the group lasso doc

mathurinm (approved):

PR #235, opened 2024-03-27
MNT: change solver quantile regressor sklearn

mathurinm:
sklearn raises an error when the version of scipy is > 1.11 and the QuantileRegressor solver is `"interior-point"`

### DIFF
--- a/skglm/experimental/tests/test_quantile_regression.py
+++ b/skglm/experimental/tests/test_quantile_regression.py
@@ -28,7 +28,8 @@ def test_PDCD_WS(quantile_level):
     clf = QuantileRegressor(
         quantile=quantile_level,
         alpha=alpha/n_samples,
-        fit_intercept=False
+        fit_intercept=False,
+        solver='highs',
     ).fit(X, y)
 
     np.testing.assert_allclose(w, clf.coef_, atol=1e-5)


skglm/experimental/tests/test_quantile_regression.py
```diff
@@ -28,7 +28,8 @@ def test_PDCD_WS(quantile_level):
     clf = QuantileRegressor(
         quantile=quantile_level,
         alpha=alpha/n_samples,
-        fit_intercept=False
+        fit_intercept=False,
+        solver='highs',
```
Badr-MOUFAD, line 32: From v1.4 onward, the default solver is [`highs`](https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.QuantileRegressor.html)

But I think, it is good idea to pin it

Badr-MOUFAD (approved):

skglm/experimental/tests/test_quantile_regression.py
mathurinm: yes that's weird, I got the issue without specifying. it was quicker to fix this way, even more because it's in a test and not our codebase so I'm ok with not fully understanding...

PR #239, opened 2024-04-02
DOC add ucurve example

mathurinm:
Did a script with a student, thought I would fit well here 

Result:
image

### DIFF
--- a/examples/plot_ucurve.py
+++ b/examples/plot_ucurve.py
@@ -3,37 +3,52 @@
 Show U-curve of regularization
 ==============================
 Illustrate the sweet spot of regularization: not too much, not too little.
+We showcase that for the Lasso estimator on the ``rcv1.binary`` dataset.
 """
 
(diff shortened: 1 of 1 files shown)


QB3 (approved):
Looks great to me, thanks for the contribution!

Badr-MOUFAD (approved):
Love the example, thanks @mathurinm

examples/plot_ucurve.py
```diff
@@ -17,8 +17,8 @@
 from skglm import Lasso
 
 # %%
-# First, we load the dataset and keep 2000 features.
-# We also retrain 2000 samples in training dataset.
+# First, we load the dataset and keep ``2000`` features.
```
mathurinm: thanks @Badr-MOUFAD for the update, I'm just removing the code style here since it's not a variable and having too much of those make reading more difficult IMO

PR #245, opened 2024-04-10
Nbb

hoodaty:
## Contributions of the PR

Added an L1_plu_L2 penalty
### Checks before merging PR

- [ ] added unittests
- [ ] edited the [what's new](../doc/changes/whats_new.rst)(if applicable)

### DIFF
--- a/skglm/datafits/single_task.py
+++ b/skglm/datafits/single_task.py
@@ -27,7 +27,10 @@ class Quadratic(BaseDatafit):
     """
 
     def __init__(self):
-        pass
+        self.sample_weights = None
+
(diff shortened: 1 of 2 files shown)


mathurinm:
@hoodaty can you also edit the first message of this PR to link to the related issue?

You need to add a unit test for this, again a comparison against sklearn's Lasso with sample_weights seems the easiest way in my opinion.

QB3:
Quick design question: 

since almost no code is shared in the altered functions, would it be worth it to add a class `WeightedQuadratic` inheriting from `Quadratic`, and overwriting only the functions we need?

PR #250, opened 2024-05-27
WIP ENH add censored quadratic df

### LINKED ISSUE #249: FEAT - Add quadratic datafit with no access to the target
## Description of the feature

**Exact feature** 

Solve the following optimization problem 

$$\arg \min_{\beta} \frac{1}{2n} \|| X \beta \||^2  + \frac{1}{n} \beta^\top X^\top y +  \text{penalty} \enspace,$$

with no access to $y$, but with access to $X^\top y$.

**Additional context**

Context,
I have been discussing with @shz9 to implement a specific datafit for genomic applications (@shz9 is finishing his PhD on [statistical analysis of genomics data](https://github.com/shz9/magenpy)). From what I understood, genomics data are sensitive: one does not have access to the target $y$: one only has access to the design matrix $X$, and an estimation of $X^\top y$ (usually estimated from another dataset).

**Steps**

I guess we have to add the datafit  $$\frac{1}{2n} \|| X \beta \||^2  + \frac{1}{n} \beta^\top X^\top y \enspace ,$$ and handle the fact there is no $y$ provided.

mathurinm:
closes #249

### DIFF
--- a/skglm/datafits/single_task.py
+++ b/skglm/datafits/single_task.py
@@ -75,7 +75,8 @@ class Quadratic(BaseDatafit):
             self.Xty[j] = xty
 
     def get_global_lipschitz(self, X, y):
-        return norm(X, ord=2) ** 2 / len(y)
+        n_samples = X.shape[0]
+        return norm(X, ord=2) ** 2 / n_samples
 
     def get_global_lipschitz_sparse(self, X_data, X_indptr, X_indices, y):
         return spectral_norm(X_data, X_indptr, X_indices, len(y)) ** 2 / len(y)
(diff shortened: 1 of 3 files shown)


mathurinm:
I have an issue with the datafit update step, because the gradient of `0.5 || Xw - y + a 1_n||` with respect to `a` is `n a - (Xw - y)^t 1_n`, so to compute it one needs y.sum() 


@shz9 do people usually fit an intercept with such models ? How do they do it ?

shz9:
We usually assume $y$ is centered. So, you could go with the implementation with `y_mean`, but set it to 0 by default?

skglm/datafits/single_task.py
```diff
@@ -107,6 +108,101 @@ def full_grad_sparse(
...
+
+    Note
+    ----
+    The class is jit compiled at fit time using Numba compiler.
+    This allows for faster computations.
+    """
+
+    def __init__(self, Xty, y_mean):
+        self.Xty = Xty
+        self.y_mean = y_mean
```
Badr-MOUFAD, line 132: We better leave the constructor to define the Datafit hyper-parameters.
We can move `Xty` and `y_mean` to the `initialize` method

mathurinm: this would break the existing code: all solvers call `datafit.initialize(X, y)` internally

mathurinm: maybe a better API would be to instantiate all datafits with X, y and whatever they need (each has its own API)
then call `datafit.initialize()` that would use all stored attributes

eg
```python
datafit = Quadratic(X, y)
penalty = L1(alpha=0.1)
solver = AndersonCD()
solver.solve(datafit, penalty)
```

It would give more freedom to each datafit, to require various quantities. 

To me this makes more sense because we have datafits like Cox depneding on more than just X and y.

Wdyt @BadrMoufad ?

QB3:
I just added a way to load the design matrix in a sparse scipy format from `magenpy`.
Would show us how to load `X.T y` from `magenpy` @shz9 ?

shz9:
Sure, here's a snippet that you can easily integrate for the purposes of testing:

```python

import magenpy as mgp

# Initialize simulator object:
g_sim = mgp.PhenotypeSimulator(mgp.tgp_eur_data_path(), h2=0.5, pi=0.1)
# Simulate outcome (phenotype):
g_sim.simulate()

# Access X:
X_csr = g_sim.genotype[22].to_csr()
# Convert to csc:
X_csc = X_csr.tocsc()
# Access y:
y = g_sim.to_phenotype_table()['phenotype'].values

# Sample size:
n = X_csr.shape[0]

# To get X'y (you may need to standardize both X and y before this step):

Xty = X_csr.T.dot(y)

# For a more general case that will be useful for GWAS settings:

# (1) Perform GWAS:
g_sim.perform_gwas()

# (2) Extract marginal betas (Xty, assumes both x and y are standardized column-wise):
g_sim.sumstats_table[22].standardized_marginal_beta
```

QB3:
Thanks a lot for the code snippet!

> Would it be possible to have the interface take` X.T X` instead of `X` in some instances?

This should take more time because we will have to implement a new solver. 
We will start with making the code work for `X`, and then `X.T X.`

shz9:
OK, thanks. By the way, I found some interesting statistics literature on losses of the form that we're working on here, going back to Loh and Wainwright (2012):

- [High-dimensional regression with noisy and missing data: Provable guarantees with nonconvexity](https://projecteuclid.org/journals/annals-of-statistics/volume-40/issue-3/High-dimensional-regression-with-noisy-and-missing-data--Provable/10.1214/12-AOS1018.full)
- [CoCoLasso for high-dimensional error-in-variables regression](https://projecteuclid.org/journals/annals-of-statistics/volume-45/issue-6/CoCoLasso-for-high-dimensional-error-in-variables-regression/10.1214/16-AOS1527.full)
- [An Investigation of Methods for Handling Missing Data with Penalized Regression](https://arxiv.org/pdf/1310.2076)

These papers were focused on the regime of missing/noisy data in $X$ and they proposed losses similar to the ones we're discussing. So, the work here could have wide applicability if we manage to provide efficient and robust solvers for these losses.

This literature also provides some important caveats and considerations to keep in mind, for example, the fact that $X^\top X$ has to be positive (semi-)definite.

PR #252, opened 2024-05-30
WIP investigate Poisson regression when some y = 0

mathurinm:
Investigating #251 
@TianBao you need to use a ProxNewton solver for a Poisson datafit (this should be made clearer by #137)

For now I added 1 to all values of `y`, and I'm seeing huge values objectivewise:
```
## -- End pasted text --
Iteration 1: 1.0000000000, stopping crit: 1.78e+07
PN iteration 1: 47722467018950103902903968817223327893258723966799524558896562424211198402946080216412986890984698293256462805576153114447728829147501450550021939395283244435578230345230699746135156461194105704397260762824686990032267220059603941850087424.0000000000, stopping crit in: 2.06e+239
```

Issue #251, opened 2024-05-30
Poisson regression problems in skglm


Tianbo-Diao:
System: Windows 10
python version: 3.11.8

code:
```python
import numpy as np
from time import time
from skglm import GeneralizedLinearEstimator
from skglm import datafits
from skglm import penalties

n = 200
p = 5000
beta_value = np.array([5, 3, 0, 0, -2, 0] + [0] * (p - 6))

start_time = time()
popu_X = simulator.simu_X(n, p, example='hete')
popu_Y = simulator.simu_Y(popu_X, n, beta=beta_value, type='poisson')
print('simulate data time:', time() - start_time)

start_time = time()
result =  GeneralizedLinearEstimator(datafit= datafits.Poisson(), penalty=penalties.L1(alpha=1)).fit(X = popu_X, y = popu_Y).coef_
print('simulate data time:', time() - start_time)
print( np.where(result !=0)[0] )
print( result[np.where(result !=0)[0]] )
```

In Poisson regression, using the above code will prompt: 
```
ValueError: Target vector `y` should only take positive values when fitting a Poisson model.
```
However, personally, the value of the response variable y in Poisson regression should be able to take 0. I wonder if this prompt is related to the specific algorithm or some other setting? Is it possible to modify it to allow y to take the value of 0?

QB3:
I gave a quick look a the [implementation of the Poisson datafit](https://github.com/scikit-learn-contrib/skglm/blob/8aea6117db5f890b1b28af876ff3dfaefd3723b0/skglm/datafits/single_task.py#L411), at first glance it seems that $0$ value for $y$ should not be a problem. 

Are you aware of specific problems/different formulas when the target $y$ takes value $0$? (I ran into [this](https://stats.stackexchange.com/questions/322344/how-does-poisson-regression-handle-zeros-anyway), which links to [this](https://blog.stata.com/2011/08/22/use-poisson-rather-than-regress-tell-a-friend/)). 
If not we can switch the `np.any(y <= 0)` to `np.any(y < 0)` (https://github.com/scikit-learn-contrib/skglm/blob/8aea6117db5f890b1b28af876ff3dfaefd3723b0/skglm/datafits/single_task.py#L435)

Tianbo-Diao:
Hello, the email has been received and will be dealt with as soon as possible.

Tianbo-Diao:
[...]

PR #253, opened 2024-05-31
ENH implement gradient and allow `y_i = 0` in `Poisson` datafit

mathurinm:
This fixes some of the issues in #251 and can be merged while we conitnue to investigate the particular, data-related problem encountered in #252


Badr-MOUFAD (approved):
LGTM 🚀

PR #258, opened 2024-05-31
Sample weights mcp

sujay-pandit:
## Contributions of the PR

Added a new class 'WeightedQuadratic' in datafits/single_task.py 

### Checks before merging PR

- [ ] added unittests
- [ ] edited the [what's new](../doc/changes/whats_new.rst)(if applicable)


skglm/datafits/single_task.py
```diff
@@ -774,3 +773,114 @@ def _get_indptr(self, vals, indices):
         indptr.append(n_indices)
 
         return np.asarray(indptr, dtype=np.int64)
+
+
+class WeightedQuadratic(BaseDatafit):
```
mathurinm, line 778: I think this could be moved closer to the Quadratic class (at the top of the file)

skglm/datafits/single_task.py
```diff
@@ -774,3 +773,114 @@ def _get_indptr(self, vals, indices):
...
+    """Weighted Quadratic datafit.
+
+    The datafit reads:
+
+    .. math:: 1 / (2 xx  n_"samples") ||y - Xw||_2 ^ 2
```
mathurinm, line 783: change the expression: this is for Quadratic

skglm/datafits/single_task.py
```diff
@@ -774,3 +773,114 @@ def _get_indptr(self, vals, indices):
...
+
+    Attributes
+    ----------
+    Xty : array, shape (n_features,)
+        Pre-computed quantity used during the gradient evaluation.
+        Equal to ``X.T @ y``.
+    sample_weights : array, shape (n_train_samples,)
```
mathurinm, line 790: Just `n_samples` in our notation

skglm/datafits/single_task.py
```diff
@@ -774,3 +773,114 @@ def _get_indptr(self, vals, indices):
...
+        Weights for each sample.
+    Note
+    ----
+    The class is jit compiled at fit time using Numba compiler.
+    This allows for faster computations.
+    """
+
+    def __init__(self, sample_weights):
+        self.sample_weights = sample_weights
+        pass
```
mathurinm, line 800: remove

skglm/datafits/single_task.py
```diff
@@ -774,3 +773,114 @@ def _get_indptr(self, vals, indices):
...
+
+    def get_spec(self):
+        spec = (
+            ('Xty', float64[:]),
+            ('sample_weights', float64[:]),
+        )
+        return spec
+
+    def params_to_dict(self):
+        return {'sample_weights':self.sample_weights}
```
mathurinm, line 810: missing whitespace after `:`. You can add a linter in vscode to do this automatically for you whenever you save

skglm/datafits/single_task.py
```diff
@@ -774,3 +773,114 @@ def _get_indptr(self, vals, indices):
...
+
+    def get_lipschitz(self, X, y):
+        n_features = X.shape[1]
+        lipschitz = np.zeros(n_features, dtype=X.dtype)
+
+        for j in range(n_features):
+            lipschitz[j] = (self.sample_weights * X[:, j] ** 2).sum() / self.sample_weights.sum()
```
mathurinm, line 817: the sum of weights can be computed outside of the for loop to be faster

skglm/datafits/single_task.py
```diff
@@ -774,3 +773,114 @@ def _get_indptr(self, vals, indices):
...
+    def get_lipschitz_sparse(self, X_data, X_indptr, X_indices, y):
+        n_features = len(X_indptr) - 1
+        lipschitz = np.zeros(n_features, dtype=X_data.dtype)
+
+        for j in range(n_features):
+            nrm2 = 0.
+            for idx in range(X_indptr[j], X_indptr[j + 1]):
+                nrm2 += self.sample_weights[X_indices[idx]] * X_data[idx] ** 2
+
+            lipschitz[j] = nrm2 / self.sample_weights.sum()
```
mathurinm, line 830: same remark for sum

skglm/datafits/single_task.py
```diff
@@ -774,3 +773,114 @@ def _get_indptr(self, vals, indices):
...
+
+        return lipschitz
+
+    def initialize(self, X, y):
+        self.Xty = X.T @ (self.sample_weights * y)
```
mathurinm, line 835: I would not call it Xty if it is not equal to `X.T @ y`. I suggest `wXty` where `w` stands for weighted

mathurinm (commented):
This looks very good @sujay-pandit , I did a few comments.

We need to test this now. You can test that:
- with weights equal to 1 it gives the same thing as a Quadratic
- with integers weights, it gives the same thing as a Quadratic with duplicated lines (eg if sample_weights[0] = 3, create a new X where X[0], y[0] are repeated 3 times.

Let me know if I can clarify anything, and thanks again

QB3 (commented):
This looks great and should be perfect after taking into account @mathurinm 's comments.
Quick question: do we want to store `self.sample_weights.sum()` once for all in an attribute at initialization? 

`self.sum_sample_weights = self.sample_weights.sum()`

This looks more efficient, but also prone to errors if one changes `sample_weights`, without its sum.

mathurinm:
> This looks great and should be perfect after taking into account @mathurinm 's comments. Quick question: do we want to store `self.sample_weights.sum()` once for all in an attribute at initialization?
> 
> `self.sum_sample_weights = self.sample_weights.sum()`
> 
> This looks more efficient, but also prone to errors if one changes `sample_weights`, without its sum.

The more we develop skglm, the more I thnk we should target safety and code readability rather than speed. here in particular I think the burden is neglectable (may,be when computing a single coordinate gradient, we move for O(nnz(X[:, j]) to O(n_samples)

I'd leave it as is and optimize later if we find out it's slowing things down.

mathurinm:
I did a pass and added a test, where I test that integer sample weights are the same as having duplicating rows in X and y. 

It is failing because the two solvers take a different number of iterations to solve. They yield the same coefficients though. I wonder if this hides something more or if it's only numerical errors.
@Badr-MOUFAD can you have a quick look ?

QB3:
> It is failing because the two solvers take a different number of iterations to solve. They yield the same coefficients though. 

The test was passing for the CI and me.

> I wonder if this hides something more or if it's only numerical errors.

I added a more complex test: it can fail (with `n_samples = n_features = 50`), but for "larger" problems,  (`n_samples = 50,  n_features = 200`), one can see that all the stopping criterion values of the 10 iterations are the same (for a range of values from `1e0` to `1e-12`).
The values of the stopping criterion start to be different around `1e-13`.
From the lens of these observations, I would tend to say these are numerical issues and merge the PR.

QB3 (approved):
Overall LGTM. The errors we sometimes have look like numerical errors, we might have to add an xfail at some point.

mathurinm:
I just had to fix the test to make sure that we support 0 weights, but now everything works fine.

Thanks @sujay-pandit @QB3, this will be very useful !

sujay-pandit:
Thanks @mathurinm, @QB3 for the opportunity.

PR #282, opened 2025-04-04
conda install R on test workflow

mathurinm:


mathurinm:
Some tests are still failing but at least the install of rpy2 is OK. The other failures are due to some new sklearn requirements in test_estiamtors, will fix later.

@jolars test_penalty.test_slope is giving different results for your Slope and ours, any idea why (we should not fit an intercept in your model, but even with this locally it fails)

jolars:
Possibly a bug in sortedl1. I'm a bit behind on updating it. I'll take a look at this next week. Thanks for the heads up.4 apr. 2025 kl. 17:35 skrev mathurinm ***@***.***>:﻿
  Some tests are still failing but at least the install of rpy2 is OK. The other failures are due to some new sklearn requirements in test_estiamtors, will fix later.
@jolars test_penalty.test_slope is giving different results for your Slope and ours, any idea why (we should not fit an intercept in your model, but even with this locally it fails)—Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you were mentioned.Message ID: ***@***.***>



mathurinm left a comment (scikit-learn-contrib/skglm#282)
Some tests are still failing but at least the install of rpy2 is OK. The other failures are due to some new sklearn requirements in test_estiamtors, will fix later.
@jolars test_penalty.test_slope is giving different results for your Slope and ours, any idea why (we should not fit an intercept in your model, but even with this locally it fails)

PR #289, opened 2025-04-10
MNT add python version requirement

jolars:
## Context of the PR

The `requires-python` field in `pyproject.toml` is missing. 

## Contributions of the PR

So this PR pins the python version and adds a couple of classifiers related to it for pypi.

### Checks before merging PR

- ~[ ] added documentation for any new feature~
- ~[ ] added unit tests~
- ~[ ] edited the [what's new](../doc/changes/whats_new.rst) (if applicable)~

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,11 +22,10 @@ dependencies = [
 ]
 dynamic = ["version"]
 
-requires-python = ">=3.8"
+requires-python = ">=3.9"
 
 classifiers = [
   "Programming Language :: Python :: 3 :: Only",
-  "Programming Language :: Python :: 3.8",
   "Programming Language :: Python :: 3.9",
   "Programming Language :: Python :: 3.10",
   "Programming Language :: Python :: 3.11",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,6 +22,14 @@ dependencies = [
 ]
 dynamic = ["version"]
 
+requires-python = ">=3.11"
+
+classifiers = [
+  "Programming Language :: Python :: 3 :: Only",
+  "Programming Language :: Python :: 3.11",
+  "Programming Language :: Python :: 3.12",
+  "Programming Language :: Python :: 3.13",
+]
 
 [tool.setuptools.dynamic]
 version = {attr = "skglm.__version__"}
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -22,6 +22,17 @@ dependencies = [
 ]
 dynamic = ["version"]
 
+requires-python = ">=3.8"
```
Badr-MOUFAD, line 25: scikit-learn requires at least 3.9 (e.g [classifier](https://pypi.org/project/scikit-learn/))
I don't know to which extent skglm would work with lower version

mathurinm: 3.8 has reached its EOL 6 months ago so go for 3.9 !

mathurinm, line 25:
```suggestion
requires-python = ">=3.9"
```

pyproject.toml
```diff
@@ -22,6 +22,17 @@ dependencies = [
...
+
+classifiers = [
+  "Programming Language :: Python :: 3 :: Only",
+  "Programming Language :: Python :: 3.8",
+  "Programming Language :: Python :: 3.9",
```
mathurinm, line 30:
```suggestion
  "Programming Language :: Python :: 3.9",
```

pyproject.toml
jolars: Seems reasonable. I originally set this to 3.11 actually, but then saw this:

`ERROR: Package 'skglm' requires a different Python: 3.8.1 not in '>=3.9'`

in the circle CI test logs, so I figured you needed 3.8. But the error still remains so I guess it's something different. 

Scipy and numpy both require at least 3.11 now, numba 3.10.

mathurinm: I just bumped the version of Python used in our circleci workflow to 3.10 instead

mathurinm:
@Badr-MOUFAD pytest fails because of something that we should fix, merge if OK for you

Badr-MOUFAD (approved):
It doesn't hurt us to update this

https://github.com/scikit-learn-contrib/skglm/blob/6309a6adfb734c41722e7e17516f096ce2f6aa59/.github/workflows/flake8.yml#L23

Only needed for code format checks.

LGTM, thnx @jolars @mathurinm 💪

PR #301, opened 2025-04-29
Issue#296 -Precompile skglm to cache JIT

floriankozikowski:
## Context of the PR

Cash compilation times in timing comparison

## Contributions of the PR

Pre-warm skglm (dummy .fit) so its compile time is excluded:

 
skglm uses JIT (Numba/Cython), so the first .fit(...) pays a multi-second compile cost; pre-running a dummy .fit(max_iter=10_000) (and even one at each small max_iter) moves all that out of the timed loop. (=cache compile step)
scikit-learn is already AOT-compiled, so there’s no hidden JIT to hide—its very first .fit is a fair measurement of solver speed.

### Checks before merging PR

- [ ] review if understanding of cache compilation time is correct

### DIFF
--- a/examples/plot_compare_time.py
+++ b/examples/plot_compare_time.py
@@ -65,6 +65,12 @@ for ax, model, l1_ratio in zip(axarr, models, [1, 0.5]):
     dict_ours[model].max_iter = 10_000
     w_star = dict_ours[model].fit(X, y).coef_
     pobj_star = compute_obj(X, y, w_star, alpha, l1_ratio)
+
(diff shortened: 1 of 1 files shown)


mathurinm:
I don't see a change in the example here : https://output.circle-artifacts.com/output/job/a5711447-f3cd-4ea4-ae18-0c2088892fe3/artifacts/0/dev/auto_examples/plot_compare_time.html#sphx-glr-auto-examples-plot-compare-time-py

That's weird, can you reproduce outside the example that fitting takes a long time even when calling fit multiple times ?

Issue #302, opened 2025-04-29
DOC - skglm is slow on news20.binary


floriankozikowski:
Lasso.fit on news20.binary is ~2.5 s in skglm vs. much faster in scikit-learn

Follow-up to: scikit-learn-contrib/skglm#296

Initial Thought: JIT compilation slows down skglm, this was already adressed though, there must be another more fundamental bug in the code.

mathurinm:
@Badr-MOUFAD this was raised from this example, where despite caching compilation we are much slower than sklearn.
It feels like we are doing something wrong

Reproduce with
```
import time 
import numpy as np
from libsvmdata import fetch_libsvm


from skglm import Lasso

X, y = fetch_libsvm("news20.binary")
alpha = np.max(np.abs(X.T @ y)) / len(y) / 2
model =  Lasso(alpha=alpha, fit_intercept=False, tol=1e-12)


# cache compil
model.fit(X, y)

for _ in range(2):
    t0 = time.time()
    model.fit(X, y)
    print("Time to fit:", time.time() - t0) 
```
When running with version 0.4, it takes 0.5 seconds to fit. With 0.5dev, it takes 3 s.

Badr-MOUFAD:
Found the bug 🐛 , I will a send a PR

PR #303, opened 2025-04-29
DOC - WIP: try to fix time comparison, but could not find root cause

floriankozikowski:
## Context of the PR

Timing comparison with scikit-leran for Lasso is wrong and presents skglm in a bad light. Potential reasons are warmups, but root cause could not be identified yet. 
( Closes issue #302 )

## Contributions of the PR

WIP: Debug, interestingly the graphs appear correct if run in Jupyter Notebook instead of a python script, which underlines the assumption that something regarding warm-up / compilation time does not quite work correct. 

Screenshot 2025-04-29 at 18 47 13
Screenshot 2025-04-29 at 18 47 28

### Checks before merging PR

- [ ] fix issue (currently still draft)
- [ ] clean up code

### DIFF
--- a/examples/plot_compare_time.py
+++ b/examples/plot_compare_time.py
@@ -50,6 +50,24 @@ dict_ours["enet"] = ElasticNet(
 
 models = ["lasso", "enet"]
 
+# Global warmup to ensure all libraries are JIT-compiled before any benchmarking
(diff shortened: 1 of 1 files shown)


mathurinm:
fixed in #304

Issue #308, opened 2025-05-19
FEAT - Implement LogisticRegressionCV and ElasticNetCV?


mathurinm:
For sparse penalties, Using warm start along the path gives quite a boost compared to using GridSearchCV

This feature was asked here : https://github.com/mathurinm/celer/issues/302#issuecomment-2890509759

We can't have as many `*CV` classes as estimators, but ElasticNet and LogisticRegression seem popular enough to justify having those hardcoded (with both enet penalty).

Another solution would be to have `GeneralizedLinearEstimatorCV`, but it's not abvious at the moment how we would define the parameters to loop over (if the penalty is not always `L1plusL2`)

PR #311, opened 2025-05-22
FEAT - Basic GeneralizedLinearEstimatorCV 

floriankozikowski:
## Context of the PR

Implements a cross-validation wrapper for `GeneralizedLinearEstimator` along with a standalone example script. Provides hyperparameter grids over `alpha` and optional `l1_ratio`, custom k-fold splitting, warm starts, scoring, and final full-data refit (closes Issue  #308 ).

## Contributions of the PR

- class inherits from GeneralizedLinearEstimator, CV inspired by Celer
- full pass over l1_ratio (if available) and alpha grids
- warm_start: carry forward each fold’s solution (w) as the initial w_start for the next alpha
- score & select: compute MSE (or user-provided scorer) per fold, track the mean loss in mse_path, and pick the (alpha_, l1_ratio_) with lowest average loss
- final refit: once the best hyperparameters are found, update self.penalty in place and call super().fit(X, y) to train the returned model on the entire dataset
- plot_generalized_linear_estimator.cv provides a simple example file

### Checks before merging PR

- [x] added unit tests
- [x] edited the [what's new](../doc/changes/whats_new.rst) (if applicable)


skglm/penalties/generalized_linear_cv.py
```diff
@@ -0,0 +1,147 @@
...
+            return float(np.mean(self.datafit.loss(y_true, y_pred)))
+
+        if self.scoring is not None:
+            if callable(self.scoring):
+                return -float(self.scoring(y_true, y_pred))
+            scorer = get_scorer(self.scoring)
+            return -float(scorer._score_func(y_true, y_pred, **scorer._kwargs))
+
+        if isinstance(self.datafit, (Logistic, QuadraticSVC)):
+            return -float(np.mean(y_true == (y_pred > 0)))
```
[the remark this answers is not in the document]
mathurinm: and also, why comparing to y_pred > 0 and not y_pred?

skglm/penalties/generalized_linear_cv.py
```diff
@@ -0,0 +1,147 @@
...
+        return float(np.mean((y_true - y_pred) ** 2))
+
+    def fit(self, X, y):
+        """Fit the model using cross-validation."""
+        if not hasattr(self.penalty, "alpha"):
+            raise ValueError("'penalty' must expose an 'alpha' hyper-parameter.")
+        X, y = np.asarray(X), np.asarray(y)
```
mathurinm, line 58: why this ? this breaks sparsity if user passes sparse scipy matrix

skglm/penalties/generalized_linear_cv.py
```diff
@@ -0,0 +1,147 @@
...
+        n, p = X.shape
```
mathurinm, line 59: we use n_samples, n_features as notation

skglm/penalties/generalized_linear_cv.py
```diff
@@ -0,0 +1,147 @@
...
+        rng = np.random.RandomState(self.random_state)
+
+        if self.alphas is not None:
+            alphas = np.sort(self.alphas)[::-1]
+        else:
+            alpha_max = np.max(np.abs(X.T @ y)) / n
+            alphas = np.logspace(
```
mathurinm, line 66: use np.geomspace to avoid get the result directly

skglm/penalties/generalized_linear_cv.py
```diff
@@ -0,0 +1,147 @@
...
+                np.log10(alpha_max),
+                np.log10(alpha_max * self.eps),
+                self.n_alphas
+            )[::-1]
+        has_l1_ratio = hasattr(self.penalty, "l1_ratio")
+        l1_ratios = [1.] if not has_l1_ratio else np.atleast_1d(
+            self.l1_ratio if self.l1_ratio is not None else [1.])
+
+        folds = list(_kfold_split(n, self.cv, rng))
```
mathurinm, line 75: you probably can avoid building the full list, that the goal of an enumerator. not crucial here

skglm/penalties/generalized_linear_cv.py
```diff
@@ -0,0 +1,147 @@
...
+        n_folds = len(folds)
+
+        mse_path = np.empty((len(l1_ratios), len(alphas), n_folds))
```
mathurinm, line 78: if l1_ratio is None we want a 1d array directly, not 2D

skglm/penalties/generalized_linear_cv.py
```diff
@@ -0,0 +1,147 @@
...
+        best_loss = np.inf
+
+        def _solve_fold(k, train, test, alpha, l1, w_start):
+            pen_kwargs = {k: v for k, v in self.penalty.__dict__.items()
+                          if k not in ("alpha", "l1_ratio")}
+            if has_l1_ratio:
+                pen_kwargs['l1_ratio'] = l1
+            pen = type(self.penalty)(alpha=alpha, **pen_kwargs)
```
mathurinm, line 86: you could just do `self.penalty.alpha = alpha`, no ?

skglm/penalties/generalized_linear_cv.py
```diff
@@ -0,0 +1,147 @@
...
+
+            kw = dict(X=X[train], y=y[train], datafit=self.datafit, penalty=pen)
+            if 'w' in self.solver.solve.__code__.co_varnames:
```
mathurinm, line 89: what does this do ? this seems overly complex

skglm/penalties/generalized_linear_cv.py
```diff
@@ -0,0 +1,147 @@
...
+                kw['w'] = w_start
+            w = self.solver.solve(**kw)
+            w = w[0] if isinstance(w, tuple) else w
```
mathurinm, line 92: which solvers return a tuple ?

skglm/penalties/generalized_linear_cv.py
```diff
@@ -0,0 +1,147 @@
...
+
+            coef, intercept = (w[:p], w[p]) if w.size == p + 1 else (w, 0.0)
+
+            y_pred = X[test] @ coef + intercept
```
mathurinm, line 96: for logistic regression the prediction is not this, it's the sign of this vector

you can delegate everything to a GeneralizedEstimator (creating it by passing datafit, penalty, solver, then changing the alpha attribute and using warm_start=True)

skglm/penalties/generalized_linear_cv.py
```diff
@@ -0,0 +1,147 @@
...
+        # Refit on full dataset
+        self.penalty.alpha = self.alpha_
+        if hasattr(self.penalty, "l1_ratio"):
+            self.penalty.l1_ratio = self.l1_ratio_
+        super().fit(X, y)
+        self.alphas_ = alphas
+        self.mse_path_ = mse_path
+        return self
+
+    def predict(self, X):
```
mathurinm, line 127: delegate to an instance of GeneralizedLinearEstimator instead

skglm/penalties/generalized_linear_cv.py
```diff
@@ -0,0 +1,147 @@
...
+                    self.l1_ratio_ = float(l1_ratio) if l1_ratio is not None else None
+
+        # Refit on full dataset
+        self.penalty.alpha = self.alpha_
+        if hasattr(self.penalty, "l1_ratio"):
+            self.penalty.l1_ratio = self.l1_ratio_
+        super().fit(X, y)
+        self.alphas_ = alphas
+        self.mse_path_ = mse_path
+        return self
```
mathurinm, line 125: Look at how LassoCV works: after fitting, you can directly call it to predict, because it stores the best Lasso. you need to do the same, storing the best model coefficient, the best  `best_alpha_ ` and recreating a GeneralizedLienarModel at the end, that fits on the whole dataset X:

look at how it's done here : 
https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/linear_model/_coordinate_descent.py#L1882

mathurinm (dismissed):

skglm/penalties/generalized_linear_cv.py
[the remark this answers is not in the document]
floriankozikowski: float I honestly just used for convenience, but I could also call .item() if you prefer that. 
For y_pred, you are right I will change it, its redundant as self.predict(X) already returns class labels. 

Otherwise I could just use sklearn.metrics.accuracy_score and mean_squared_error if thats better

skglm/penalties/generalized_linear_cv.py
floriankozikowski: true, I remove np.asarray(X), but I keep y since targets are always dense, or am I missing something?

mathurinm: all preprocessing of design matrix and target is done in GeneralizedLinearEstimator.fit. You need to leverage tthis and not do anything here

skglm/penalties/generalized_linear_cv.py
floriankozikowski: understood, but you mean 2D and 3D array, right? As we still have multiple alphas and multiple folds or no?

skglm/penalties/generalized_linear_cv.py
floriankozikowski: I thought since we run the folds in parallel, changing self.penalty.alpha directly could cause issues if multiple jobs modify it at the same time, so I created a new penalty for each fold

skglm/cv.py
```diff
@@ -0,0 +1,139 @@
+import numpy as np
+from joblib import Parallel, delayed
+from skglm.datafits import Logistic, QuadraticSVC
+from skglm.estimators import GeneralizedLinearEstimator
+
+
+def _kfold_split(n_samples, k, rng):
```
mathurinm, line 7: can you rely on sklearn KFold() instead ? It uses stratified kfold by default and should yield shorter code (less error prone)

skglm/cv.py
```diff
@@ -0,0 +1,132 @@
...
+                pen_kwargs['l1_ratio'] = l1
+            pen = type(self.penalty)(alpha=alpha, **pen_kwargs)
+
+            est = GeneralizedLinearEstimator(
+                datafit=self.datafit, penalty=pen, solver=self.solver
+            )
+            est.solver.warm_start = True
+            est.fit(X[train], y[train])
+            y_pred = est.predict(X[test])
+            return est.coef_, est.intercept_, self._score(y[test], y_pred)
```
mathurinm, line 68: Would it make more sense to delegate to est.score here?

mathurinm:
@floko can you provide some timings on datasets of reasonable sizes ?


We should handle warm start better, by storing `K` estimators (one for each fold) and using warm start  on them (so that on fold `k`, when fitting for one value of `alpha`, we use warm start with the previous value of `alpha` on the same fold.

Right now warm start is not used since we recreate an estimator at each alpha.

skglm/tests/test_cv.py
```diff
@@ -0,0 +1,37 @@
...
+from skglm.solvers import AndersonCD
+from skglm.cv import GeneralizedLinearEstimatorCV
+
+
+@pytest.mark.parametrize("seed", [0, 42])
+def test_elasticnet_cv_matches_sklearn(seed):
+    """Test GeneralizedLinearEstimatorCV matches sklearn GridSearchCV for ElasticNet."""
+    X, y = make_regression(n_samples=100, n_features=20, noise=0.1, random_state=seed)
+
+    alphas = np.array([0.001, 0.01, 0.1, 1.0])
```
mathurinm, line 17: picking absolute values for alphas is dangerous ; scale them by `\Vert X.T @ y \Vert_\infty / n` which is the smallest alpha for which the solution is 0

floriankozikowski:
@Badr-MOUFAD good catch, I removed it from the PR description (initially I had it like this, but then changed it at some point). Plus, I made the unit test shorter now. Lets me know if thats fine.
@mathurinm initially you added a part where you also test for alpha = 1e-4. For now I removed it as it was in the debugging sections of the script. But if you want it back let me know. When I tried it out the pytest still worked, but was much slower and I get convergence warnings.

skglm/tests/test_cv.py
```diff
@@ -0,0 +1,42 @@
...
+        cv=cv, scoring='neg_mean_squared_error', n_jobs=1
+    ).fit(X, y)
+
+    skglm_model = GeneralizedLinearEstimatorCV(
+        Quadratic(), L1_plus_L2(0.1, 0.5), AndersonCD(max_iter=10000, tol=1e-8),
+        alphas=alphas, l1_ratio=l1_ratios, cv=5, random_state=seed, n_jobs=1
+    ).fit(X, y)
+
+    assert sklearn_model.best_params_['alpha'] == skglm_model.alpha_
+    assert sklearn_model.best_params_['l1_ratio'] == skglm_model.l1_ratio_
```
Badr-MOUFAD, line 38: we better use [np assert equal](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_equal.html), it provides better feedback

skglm/cv.py
```diff
@@ -0,0 +1,181 @@
...
+                    split_iter = kf.split(np.arange(n_samples))
+                fold_results = Parallel(self.n_jobs)(
+                    delayed(_solve_fold)(k, tr, te, alpha, l1_ratio, warm_start[k])
+                    for k, (tr, te) in enumerate(split_iter)
+                )
+
+                for k, (coef_fold, intercept_fold, loss_fold) in \
+                        enumerate(fold_results):
+                    warm_start[k] = (coef_fold, intercept_fold)
+                    scores_path[idx_ratio, idx_alpha, k] = loss_fold
```
Badr-MOUFAD, line 142: to fit in one line

```suggestion
                for k, (coef_k, intercept_k, loss_k) in enumerate(fold_results):
                    warm_start[k] = (coef_k, intercept_k)
                    scores_path[idx_ratio, idx_alpha, k] = loss_k
```

skglm/cv.py
```diff
@@ -0,0 +1,181 @@
...
+        self.n_jobs = n_jobs
+        self.random_state = random_state
+        self.eps = eps
+        self.n_alphas = n_alphas
+
+    def _score(self, y_true, y_pred):
+        """Compute the performance score (higher is better)."""
+        if isinstance(self.datafit, (Logistic, QuadraticSVC)):
+            return float(np.mean(y_true == y_pred))
+        return -float(np.mean((y_true - y_pred) ** 2))
```
Badr-MOUFAD, line 78: I'm +1 with using sklearn `accuracy_score` and `mean_squared_error`
this forces us to stay consistent with the sklearn API, as we are drop in replacement of it

Badr-MOUFAD (approved):
LGMT @floriankozikowski, thanks for this amazing job 💪 

just minor remarks before merging

PR #316, opened 2025-06-10
FIX - add `alpha` to SLOPE penalty

### LINKED ISSUE #315: BUG - attribute `alpha` is not available for `penalty=SLOPE()` for string representation
## Describe the bug

I am getting test errors from skglm (https://github.com/benchopt/benchmark_slope/actions/runs/15496605501/job/43634699241#step:7:471):

```
self       = <[AttributeError("'SLOPE' object has no attribute 'alpha'") raised in repr()] GeneralizedLinearEstimator object at 0x13454a440>
```

The problem, I think, is that on the following line you expect `alpha` to exist: 

https://github.com/scikit-learn-contrib/skglm/blob/d57e3614cf9246b319192101f51959b7377c4514/skglm/estimators.py#L208

But for `SLOPE` it is `alphas`:

https://github.com/scikit-learn-contrib/skglm/blob/d57e3614cf9246b319192101f51959b7377c4514/skglm/penalties/non_separable.py#L27

## Steps to reproduce

```python
from skglm import GeneralizedLinearEstimator, SLOPE

from skglm.penalties import SLOPE

alphas = [0.1, 0.5, 1.0, 2.0, 5.0]

model = GeneralizedLinearEstimator(penalty = SLOPE(alphas))

repr(model)
```

**Additional context**

jolars:
## Context of the PR

See #315 

## Contributions of the PR

Add `alpha` to make `repr()` work on objects created with the SLOPE penalty. Closes #315.

### Checks before merging PR

- [x] added unit tests
- [x] ~edited the [what's new](../doc/changes/whats_new.rst) (if applicable)~

### DIFF
--- a/skglm/penalties/non_separable.py
+++ b/skglm/penalties/non_separable.py
@@ -14,6 +14,8 @@ class SLOPE(BasePenalty):
         Contain regularization levels for every feature.
         When ``alphas`` contain a single unique value, ``SLOPE``
         is equivalent to the ``L1``penalty.
+    alpha : float, default=1.0
+        Scaling factor for the penalty. `alphas` is multiplied by this value.
 
     References
     ----------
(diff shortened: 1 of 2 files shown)


jolars:
Actually I don't really have an opinion on how to handle this. My only interest in this is really to fix the bug and prevent the benchopt tests from erroring out :smile: . So I'm happy with any solution, really. That being said, I think it's worthwhile to have `alpha` be a parameter since it is almost almost this parameter along with whatever other variables (such as q) you might use to parameterize the penalty sequence. I do, however, think that the use of both `alpha` and `alphas` to mean essentially different things is a little confusing, but this way at least you don't break anything.

jolars:
I forgot to answer your question. Yes, at least the packages that I have made always have both an argument for the scaling of the penalty sequence and the penalty sequence itself. Yes it's redundant, but also practical.

mathurinm (approved):

PR #323, opened 2025-07-23
TST: Add sklearn <-> skglm match tests for Poisson and Gamma predictions

floriankozikowski:
Follows up on #321 : Add unit tests verifying sklearn prediction compatibility

This PR addresses the request from @mathurinm in #321 to add unit tests ensuring that skglm's Poisson and Gamma estimators produce the same predictions as sklearn on simple data.

These tests validate that the prediction fix from #321 (applying `exp()` transform for log-link datafits) correctly matches sklearn's behavior.

### DIFF
--- a/skglm/estimators.py
+++ b/skglm/estimators.py
@@ -19,7 +19,7 @@ from sklearn.multiclass import OneVsRestClassifier, check_classification_targets
 
 from skglm.solvers import AndersonCD, MultiTaskBCD, GroupBCD, ProxNewton, LBFGS
 from skglm.datafits import (
-    Cox, Quadratic, Logistic, Poisson, PoissonGroup, QuadraticSVC,
+    Cox, Quadratic, Logistic, Poisson, PoissonGroup, Gamma, QuadraticSVC,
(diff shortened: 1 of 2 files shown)


floriankozikowski:
@Badr-MOUFAD @mathurinm Ready for Review :)

floriankozikowski:
@Badr-MOUFAD 
Good idea with making `inverse_link` the identity by default. I just added this. 

Regarding Question 2:
Now, logistic inherits the default identity `inverse_link` from BaseDatafit (see Q1).
The classification logic remains in `predict()` rather than inverse_link because I think `inverse_link` should handle pure mathematical transformations (like log-odds to probabilities). Classification involves class mapping with `self.classes_[indices]`,  which I would say is an estimator-level responsibility. 
We could implement a proper sigmoid inverse link for Logistic to convert log-odds to probabilities, but since the classification branch never calls it the default identity is sufficient. 
Let me know what you think and I can change it!

mathurinm:
Agree with @floriankozikowski, the prediction logic for classification is different

mathurinm:
Thanks @floriankozikowski !

Issue #320, opened 2025-07-18
ProxNewton and Poisson Datafit Yields Incorrect Predictions


jonpedros:
I've encountered an issue where skglm.GeneralizedLinearEstimator using a Poisson() datafit and a ProxNewton() solver produces incorrect results on a simple synthetic dataset when fit_intercept=True (the default).

The solver reports successful convergence, but the predict() method returns systematically wrong values, leading to a large negative pseudo-R².

```python
import numpy as np
import pandas as pd
import matplotlib.pyplot as plt
import seaborn as sns
import formulaic

from skglm import GeneralizedLinearEstimator
from skglm.datafits import Poisson
from skglm.penalties import L1_plus_L2
from skglm.solvers import ProxNewton
from sklearn.metrics import mean_poisson_deviance

def generate_generic_poisson_data(n_samples=20000, seed=42):
    """Generates a simple, well-behaved dataset for a Poisson GLM."""
    print("\n1. Generating synthetic data...")
    np.random.seed(seed)
    
    true_intercept = -1.5
    true_coefs = {
        'x_continuous': 1.8,
        'C(x_categorical)[T.B]': -1.2,
        'x_noise': 0.0
    }

    data = pd.DataFrame({
        'x_continuous': np.random.rand(n_samples),
        'x_categorical': np.random.choice(['A', 'B'], n_samples),
        'x_noise': np.random.randn(n_samples)
    })

    formula = "1 + x_continuous + C(x_categorical) + x_noise"
    X_design_for_eta = formulaic.model_matrix(formula, data=data)
    
    feature_names = [col for col in X_design_for_eta.columns if col != 'Intercept']
    eta = (true_intercept + 
           X_design_for_eta[feature_names] @ pd.Series(true_coefs)[feature_names])
    mu = np.exp(eta)
    
    data['y'] = np.random.poisson(mu)
    
    return data, true_intercept, true_coefs


# --- Main script ---
# 1. Generate data
df, true_intercept, true_coefs = generate_generic_poisson_data()

# 2. Create design matrix and response vector
print("\n2. Creating design matrix...")
formula = "y ~ x_continuous + C(x_categorical) + x_noise"
y_true_s, X_design = formulaic.model_matrix(formula, data=df)

y_true = y_true_s.to_numpy().ravel()
X = X_design.drop(columns='Intercept').to_numpy()
X_feature_names = X_design.drop(columns='Intercept').columns

# 3. Configure and fit the model
[...]

PR #326, opened 2025-07-29
FEAT - add fit_intercept support for LBFGS

### LINKED ISSUE #325: FEAT - Add fit_intercept support to LBFGS solver
## Description of the feature

The LBFGS solver currently doesn't support `fit_intercept=True`. A use case of this was described in Issue #320 

## Considered alternatives

Manual data centering works but complicates the optimization and user experience. In many use cases sklearn is currently is stil the faster alternative. 

**Additional context**

```
np.random.seed(42)
X = np.random.randn(1000, 50)
y = np.random.poisson(np.exp(X.sum(axis=1) * 0.01 - 1))

# This should work but currently fails as its not supported
estimator = GeneralizedLinearEstimator(
    datafit=Poisson(),
    penalty=L2(alpha=0.01),
    solver=LBFGS(fit_intercept=True)  # <- Currently not supported
)
estimator.fit(X, y)

# Compare with sklearn (works fine)
sklearn_model = PoissonRegressor(alpha=0.01, fit_intercept=True, solver='lbfgs')
sklearn_model.fit(X, y)
```

Note: Pattern exists in `anderson_cd.py` (lines 163-166) and other solvers to see how to implement fit_intercept method.

floriankozikowski:
## Context of the PR

Closes #325 

## Contributions of the PR

- add fit_intercept support for LBFGS 
- add default warm_start=False, so LBFGS works with debug script in #320

### Checks before merging PR

- [ ] added unit tests
- [ ] edited the [what's new](../doc/changes/whats_new.rst) (if applicable)

### DIFF
--- a/skglm/solvers/lbfgs.py
+++ b/skglm/solvers/lbfgs.py
@@ -24,6 +24,9 @@ class LBFGS(BaseSolver):
     tol : float, default 1e-4
         Tolerance for convergence.
 
+    fit_intercept : bool, default False
+        Whether or not to fit an intercept.
+
     verbose : bool, default False
         Amount of verbosity. 0/False is silent.
     """
@@ -31,9 +34,11 @@ class LBFGS(BaseSolver):
     _datafit_required_attr = ("gradient",)
     _penalty_required_attr = ("gradient",)
 
-    def __init__(self, max_iter=50, tol=1e-4, verbose=False):
+    def __init__(self, max_iter=50, tol=1e-4, fit_intercept=False, verbose=False):
         self.max_iter = max_iter
         self.tol = tol
+        self.fit_intercept = fit_intercept
+        self.warm_start = False
         self.verbose = verbose
 
     def _solve(self, X, y, datafit, penalty, w_init=None, Xw_init=None):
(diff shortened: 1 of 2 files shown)


skglm/solvers/lbfgs.py
```diff
@@ -46,25 +51,46 @@ def _solve(self, X, y, datafit, penalty, w_init=None, Xw_init=None):
...
-            datafit_value = datafit.value(y, w, Xw)
-            penalty_value = penalty.value(w)
+            if self.fit_intercept:
+                Xw = X @ w[:-1] + w[-1]
+                datafit_value = datafit.value(y, w[:-1], Xw)
+                penalty_value = penalty.value(w[:-1])
+            else:
+                Xw = X @ w
+                datafit_value = datafit.value(y, w, Xw)
+                penalty_value = penalty.value(w)
```
Badr-MOUFAD, line 61: What do you think about this refactoring @floriankozikowski ?
```suggestion
            w_features = w[:n_features]
            Xw = X @ w_features
            if self.fit_intercept:
                Xw += w[-1]
                
            datafit_value = datafit.value(y, w_features , Xw)
            penalty_value = penalty.value(w_features)
```

skglm/solvers/lbfgs.py
```diff
@@ -46,25 +51,46 @@ def _solve(self, X, y, datafit, penalty, w_init=None, Xw_init=None):
...
-            Xw = X @ w
-            datafit_grad = datafit.gradient(X, y, Xw)
-            penalty_grad = penalty.gradient(w)
-
-            return datafit_grad + penalty_grad
+            if self.fit_intercept:
+                Xw = X @ w[:-1] + w[-1]
+                datafit_grad = datafit.gradient(X, y, Xw)
+                penalty_grad = penalty.gradient(w[:-1])
+                intercept_grad = datafit.intercept_update_step(y, Xw)
```
mathurinm, line 70: there may be an issue here, because intercept_update_step does not compute the gradient, but a scaled version of it (it's mulliplied by the stepsize).
The safest way to do it would be to call `raw_grad().sum()`, which is equivalent to `np.ones(n_features) @ raw_grad()`, e.g. the gradient with respect to a feature full of ones.

Sounds good @Badr-MOUFAD ?

floriankozikowski:
@Badr-MOUFAD I tried the refactor (also considering your comment @mathurinm )
Let me know what you think. Didn't find an option to make this refactor shorter

floriankozikowski:
Btw, looking at the initial issue #320 that initiated this PR, the intercept fitting improved the sklearn speed difference, but it is still slower. 

--- Fitting Time Comparison ---
skglm (LBFGS):   8.0444 seconds
sklearn (L-BFGS):     2.4578 seconds
 >> sklearn was 3.27x faster.

I guess in this PR, we only focus on the intercept, but maybe we should open a new issue investigating other causes for this. 

If you approve the refactor, I can delete the debug script (issue320) and we can merge.

skglm/solvers/lbfgs.py
Badr-MOUFAD: Good catch @mathurinm,

I have a little concern because it seems that in some part of the code (`Logostic` datafit) `intercept_update_step` account for the stepsize, whereas in other parts (`Quadratic`, `Huber`, `Poisson`, ...) `intercept_update_step` evaluate the gradient

Badr-MOUFAD:
Thanks @floriankozikowski for the timing comparison.
Weird that sklearn is x3 faster, but I agree to tackle this issue in a separate PR

Can you pls add the intercept to the unittest and update the whats'new page

floriankozikowski:
@Badr-MOUFAD thanks for the feedback! It should be complete now and I'd say we can merge. I won't have access to my laptop the next two weeks, so if anything comes up, I will look at it mid-August again.

PR #339, opened 2026-06-05
Add WeightedL1_plus_L2 penalty

### LINKED ISSUE #338: FEAT - WeightedElasticNetPenalty: per feature penalty strength factor as in glmnet
Dear all, thank you for this great package. I am doing multi omics modeling using skglm for different steps. One feature i have been using in glment is the argument "penalty.factor", a factor by which the penalty is multiplied for each feature, as to my understanding. 

I understand this argument would be a sequence/array of "factors", (,n_features), one for each gene. By default, this is 1 for each gene.

If relevant and possible, my naive Internet research has shown me the possibility of different penalty factors for L1 and L2 for elastic net penalty factor, but this is something I understand even less and would leave up to experts.

I know there is WeightedLasso with with the celer solver. However, my use cases often require elastic net. 
Would it be possible to add this, particularly as a standalone penalty for use with GeneralizedLinearEstimator?

A nice add-on would be WeightedRidgePenalty. To my understanding this can be faster than Ridge Regression using the Elastic Net Algorithm, but of course I can 

To understand the use cases I am thinking about:
1. Batch Effects. In genomics we have batch effects and other nuisance variables. Statistics procedures I have seen use penalty.factor in glmnet to remove penalization from these factors by setting the factor to zero. 
2. I have two predictor matrices for the same samples. When concatenated, I need to penalize features from each differently. But group penalty is not what I want (removal of an entire group?), instead I want one penalty.factor for each feature in a matrix.
3. Finally, I want to incorporate per feature prior knowledge by adjusting penalty strengths.

Context: glmnet documentation for the penalty.factor argument:
**penalty.factor**
Separate penalty factors can be applied to each coefficient. This is a number that multiplies lambda to allow differential shrinkage. Can be 0 for some variables, which implies no shrinkage, and that variable is always included in the model. Default is 1 for all variables (and implicitly infinity for variables listed in exclude). Note: the penalty factors are internally rescaled to sum to nvars.

It would be great to have this in skglm, so I can use it for pretty much all around modeling.

Thank you and best regards,
Max

mathurinm:
closes #338 

@Badr-MOUFAD should be straightforward to review: compared to L1_plus_L2, all `alpha * l1_ratio` become `alpha * l1_ratio * weight_l1[j]` and similarly `alpha * (1 - l1_ratio)` gets multiplied by `weights_l2[j]`

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -61,6 +61,7 @@ Penalties
    MCPenalty
    PositiveConstraint
    WeightedL1
+   WeightedL1_plus_L2
    WeightedGroupL2
    WeightedMCPenalty
    SCAD
--- a/doc/changes/0.6.rst
+++ b/doc/changes/0.6.rst
@@ -3,5 +3,7 @@
 Version 0.6 (in progress)
 -------------------------
 
+- Added support for wieghted ElasticNet penalty via :class:`skglm.penalties.WeightedL1_plus_L2`.
+
 - :class:`skglm.solvers.LBFGS` now supports fitting an intercept with the `fit_intercept` parameter.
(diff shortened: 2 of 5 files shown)


Badr-MOUFAD (approved):
looks good! thanks @mathurinm 🚀

PR #341, opened 2026-06-30
ENH: Add WeightedPoisson datafit with sample weights support

Vrinda12-tech:
## Summary

This PR adds a weighted variant of the Poisson datafit, enabling sample weights support for Poisson regression in skglm.

## Implementation

- New class `WeightedPoisson` in `skglm/datafits/single_task.py`
- Follows the `WeightedQuadratic` pattern
- Numba JIT compiled for performance
- Sparse matrix support (CSC format)
- Numerical stability: `np.clip(Xw, -500, 500)` prevents `exp` overflow
- Safety floors in Lipschitz constants to prevent division by zero

## Mathematical Derivation

The weighted Poisson loss is:

L = (1/Σw) * Σ( w_i * (exp((Xw)_i) - y_i * (Xw)_i) )

The weighted gradient:

∇L = (w * (exp(Xw) - y)) / Σw

The weighted Hessian:

H = (w * exp(Xw)) / Σw

## Tests Added

- ✅ Equality with unweighted Poisson (equal weights)
- ✅ Correct scaling with arbitrary weights
- ✅ Zero weight handling
- ✅ Negative weight validation
- ✅ Sparse matrix support
- ✅ Lipschitz constant computation
- ✅ Integration with GeneralizedLinearEstimator

All 7 tests are passing locally.

## References

- `WeightedQuadratic` implementation pattern
- `Poisson` datafit for unweighted version

### DIFF
--- a/skglm/datafits/__init__.py
+++ b/skglm/datafits/__init__.py
@@ -1,6 +1,6 @@
 from .base import BaseDatafit, BaseMultitaskDatafit
 from .single_task import (Quadratic, QuadraticSVC, Logistic, Huber, Poisson, Gamma,
-                          Cox, WeightedQuadratic, QuadraticHessian,)
+                          Cox, WeightedQuadratic, QuadraticHessian,WeightedPoisson)
 from .multi_task import QuadraticMultiTask
 from .group import QuadraticGroup, LogisticGroup, PoissonGroup
 
@@ -10,5 +10,5 @@ __all__ = [
     Quadratic, QuadraticSVC, Logistic, Huber, Poisson, Gamma, Cox,
     QuadraticMultiTask,
     QuadraticGroup, LogisticGroup, PoissonGroup, WeightedQuadratic,
-    QuadraticHessian
+    QuadraticHessian,WeightedPoisson
 ]
--- a/skglm/tests/test_datafits.py
+++ b/skglm/tests/test_datafits.py
@@ -6,7 +6,7 @@ from sklearn.linear_model import HuberRegressor
 from numpy.testing import assert_allclose, assert_array_less
 
 from skglm.datafits import (Huber, Logistic, Poisson, Gamma, Cox, WeightedQuadratic,
-                            Quadratic, QuadraticHessian)
+                            Quadratic, QuadraticHessian,WeightedPoisson)
 from skglm.penalties import L1, WeightedL1
 from skglm.solvers import AndersonCD, ProxNewton
 from skglm import GeneralizedLinearEstimator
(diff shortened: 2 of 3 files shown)


skglm/datafits/__init__.py
```diff
@@ -1,6 +1,6 @@
 from .base import BaseDatafit, BaseMultitaskDatafit
 from .single_task import (Quadratic, QuadraticSVC, Logistic, Huber, Poisson, Gamma,
-                          Cox, WeightedQuadratic, QuadraticHessian,)
+                          Cox, WeightedQuadratic, QuadraticHessian,WeightedPoisson)
```
mathurinm, line 3: you need spaces after the commas

skglm/tests/test_datafits.py
```diff
@@ -279,7 +279,209 @@ def test_HessianQuadratic():
     np.testing.assert_allclose(lasso.coef_, qpl1.coef_)
     # check that it's not just because we got alpha too high and thus 0 coef
     np.testing.assert_array_less(0.1, np.max(np.abs(qpl1.coef_)))
+# =============================================================================
```
mathurinm, line 286: this looks like a duplicate of what's 3 lines below

skglm/tests/test_datafits.py
```diff
@@ -279,7 +279,209 @@ def test_HessianQuadratic():
...
+    wp = WeightedPoisson(weights)
+    wp.initialize(X, y)
+
+    raw_grad = wp.raw_grad(y, Xw)
+    assert raw_grad[0] == 0
+
+
+def test_weighted_poisson_validation():
+    """Test input validation for WeightedPoisson."""
+    import numpy as np
```
mathurinm: you can put the imports at the top of the file

mathurinm (commented):
thanks a lot for the initiative !
There are lots of tests, do you think you could shorten them a bit for easier maintenance ? like factorize some aprts

also, can you have a single test where you have weights equal to 0, 1 or 2, and test that you get the same result as fitting with duplicated and removed samples (weights=2 or 0 respectively). I think this would remove the need for testing the lipschitz constant for example

mathurinm:
@Badr-MOUFAD at some point I wonder if it wouldn't be easier if every datafit had weights by default... but not that much time to think about it

Vrinda12-tech:
@mathurinm Thank you for the detailed review. I have addressed all the points:

1. **Reduced the number of tests**: I replaced `test_weighted_poisson_equality`, `test_weighted_poisson_scaling`, `test_weighted_poisson_zero_weights`, and `test_weighted_poisson_lipschitz` with a single **duplication test** (`test_weighted_poisson_duplication`).

2. **The duplication test**: It sets weights to 0 (remove), 2 (duplicate), and 1 (keep), then compares the weighted model against an unweighted model fit on the manually duplicated/removed dataset. This implicitly validates the loss, gradient, Hessian, and Lipschitz constants.

3. **Alpha scaling**: I initially tried to scale `alpha` for the unweighted model. I realized that because both losses are averaged (`1/n` for unweighted, `1/sum(w)` for weighted), `alpha` should be **identical** for both models. The test now uses the same `alpha` and passes.


All 4 tests (`duplication`, `validation`, `sparse`, and `integration`) are passing locally.

Let me know if you have any other suggestions!

